PR #1, opened 2021-01-07
newsroom-core

ride90:
~~⚠️ this pr is in progress and commits messages might be dummy (they'll be squashed soon)~~

**newsroom-core**
depends on https://github.com/superdesk/newsroom-app/pull/1


features/steps/steps.py
```diff
@@ -1,56 +1 @@
...
-    assert href, data
-    set_placeholder(context, 'NEXT_PAGE', href)
-
-
-@then('we get "{text}" in text response')
-def we_get_text_in_response(context, text):
-    with context.app.test_request_context(context.app.config['URL_PREFIX']):
-        assert(isinstance(get_body(context.response), str))
-        assert(text in get_body(context.response))
+from newsroom.tests.steps import * # noqa
```
petrjasek, line 1: just wondering why having tests in newsroom ?

petrjasek (approved):

features/steps/steps.py
ride90: To have an ability to reuse a core's [contest](https://github.com/superdesk/newsroom-app/blob/9dfa7f740f8265429196aaaef4c886b6bf770b10/server/tests/conftest.py#L1), [fixtures](https://github.com/superdesk/newsroom-app/blob/9dfa7f740f8265429196aaaef4c886b6bf770b10/server/tests/fixtures.py#L1) and [steps](https://github.com/superdesk/newsroom-app/blob/9dfa7f740f8265429196aaaef4c886b6bf770b10/server/features/steps/steps.py) from the newsroom-app:

requirements.txt
```diff
@@ -1,7 +1,14 @@
...
+flask-limiter>=0.9.5.1,<0.9.6
+Flask-Caching>=1.9.0
+flask_pymongo>=0.5.2,<1.0
+honcho>=1.0.1
+gunicorn>=19.7.1
+icalendar>=4.0.3,<4.1
+PyRTF3>=0.47.5
+xhtml2pdf>=0.2.4
+superdesk-core==2.0.10
+superdesk-planning@https://github.com/superdesk/superdesk-planning/archive/v2.0.1.zip
```
MarkLark86, line 14: Is there a reason to use github here instead of pypi? i.e. `superdesk-planning==2.0.2` should work right?

ride90: Hi Mark,
No reason, I think at the time I was writing this file planning 2.0 was not on pypy yet.
I will change it.

PR #7, opened 2021-10-07
merge core changes from stt

petrjasek:


data/ui_config.json
```diff
@@ -0,0 +1,77 @@
+[
```
MarkLark86, line 1: Should we include this config in core?

newsroom/commands/initialize_data.py
```diff
@@ -22,7 +24,10 @@ def run(self, entity_name=None, force=False, init_index_only=False):
         # create indexes in mongo
         app.init_indexes()
         # put mapping to elastic
-        app.data.init_elastic(app)
+        try:
+            app.data.init_elastic(app)
+        except elasticsearch.exceptions.TransportError as err:
+            logger.error("Error when initializing elastic %s", err)
```
MarkLark86, line 30: Any idea about this error:
```bash
Error when initializing elastic TransportError(500, 'mapper_exception', "Can't update attribute for type [_doc.extra.enabled] in index mapping")
```

newsroom/template_filters.py
```diff
@@ -38,7 +38,7 @@ def datetime_long(datetime):
 
 
 def date_header(datetime):
-    return format_datetime(parse_date(datetime if datetime else utcnow()), 'EEEE, MMMM d, yyyy')
+    return format_datetime(parse_date(datetime if datetime else utcnow()), 'EEEE, dd.MM.yyyy')
```
MarkLark86, line 41: This format should really be in a config somewhere

package-lock.json
```diff
@@ -328,14 +328,113 @@
       }
```
MarkLark86, line 328: Do we need a package-lock in development branches

webpack.config.js
```diff
@@ -3,32 +3,31 @@
...
+        cards_js: path.resolve(__dirname, './assets/cards/index.js'),
+        user_profile_js: path.resolve(__dirname, './assets/user-profile/index.js'),
+        newsroom_css: path.resolve(__dirname, './assets/style.js'),
+        wire_js: path.resolve(__dirname, './assets/wire/index.js'),
+        home_js: path.resolve(__dirname, './assets/home/index.js'),
+        agenda_js: path.resolve(__dirname, './assets/agenda/index.js'),
+        notifications_js: path.resolve(__dirname, './assets/notifications/index.js'),
+        company_reports_js: path.resolve(__dirname, './assets/company-reports/index.js'),
+        print_reports_js: path.resolve(__dirname, './assets/company-reports/components/index.js'),
+        am_news_js: path.resolve(__dirname, './assets/am-news/index.js'),
```
MarkLark86, line 24: We should probably remove AAP specific sections at some point

newsroom/commands/initialize_data.py
petrjasek: yep that's due to reusing previous index, needs reindexing, but first I need updated instance

data/ui_config.json
petrjasek: seems like it needs some default value, will update it

newsroom/template_filters.py
MarkLark86: What about this one? A simple config with default in `newsroom/web/default_settings.py` should suffice

PR #12, opened 2021-10-28
[NHUB-16] Configurable metadata fields for wire compact list

MarkLark86:


petrjasek:
what's the issue with tests folder?

MarkLark86:
The Superdesk's `tests` folder is being installed, and the Newsroom one isn't. Simply doing a `pip install superdesk-core` provides the `tests` folder, but when trying to install Newsroom on top, the Superdesk takes precedence.

petrjasek:
I see, that looks like an issue with superdesk core rather, tests should be ignored and not part of packages, will check that

PR #16, opened 2021-11-04
[STTNHUB-139] Ability to prepend embargoed items in wire search results

MarkLark86:


newsroom/web/default_settings.py
```diff
@@ -375,3 +375,5 @@
 DATE_FORMAT_HEADER = "EEEE, dd.MM.yyyy"
 DATETIME_FORMAT_SHORT = "short"
 DATETIME_FORMAT_LONG = "dd/MM/yyyy HH:mm"
+
+PREPEND_EMBARGOED_TO_WIRE_SEARCH = strtobool(env('PREPEND_EMBARGOED_TO_WIRE_SEARCH', 'false'))
```
petrjasek, line 379: wouldn't probably set this via environment, so setting it to just `False` by default would be fine imo

newsroom/wire/views.py
```diff
@@ -173,6 +174,10 @@ def bookmarks():
 
 @blueprint.route('/wire/search')
 def search():
+    if app.config['PREPEND_EMBARGOED_TO_WIRE_SEARCH']:
+        args = request.args.to_dict()
+        args['prepend_embargoed'] = True
```
petrjasek, line 179: maybe we could use here `args.setdefault('prepend_embargoed', True)` so we could control it from UI if needed

petrjasek (approved):

PR #21, opened 2021-11-19
[CPNHUB-11] Ability to share a "My Topic" with the user's company

MarkLark86:


newsroom/push.py
```diff
@@ -725,14 +725,19 @@ def notify_agenda_topic_matches(item, users_dict):
...
-                section=topic.get('topic_type') or 'wire'
-            )
+        if topic['_id'] not in topic_matches:
+            continue
+
+        for user_id in topic.get('subscribers') or []:
+            user = users.get(str(user_id))
+
+            if user and user.get('receive_email'):
+                send_new_item_notification_email(
```
petrjasek, line 735: would be good to group that somehow, probably per item so we would send 1 email per 1 item listing the matching topics instead of 1 email for each topic

petrjasek (approved):

PR #28, opened 2021-12-01
Nhub 38

darconny:


assets/wire/components/WireListItem.jsx
```diff
@@ -188,17 +188,19 @@ class WireListItem extends React.Component {
...
-                                    item={item}
-                                    picture={picture}
-                                    videos={videos}
-                                    divider={false}
-                                />
-                            )}
-                            <Embargo item={item} />
-                            <UrgencyLabel item={item} listConfig={listConfig} />
-                            {item.headline}
+                            <div className="wire-articles__item-heading">
```
petrjasek, line 191: just not sure about the naming, we have `item-headline` and `item-heading`, what's the difference?

petrjasek (approved):

assets/wire/components/WireListItem.jsx
darconny: We need one more wrapper inside, and it's not a modifier. Maybe something like `wire-articles__item-headline-inner` ?

petrjasek: ok

PR #36, opened 2021-12-15
merge cp-lji branch to develop

petrjasek:


petrjasek:
@MarkLark86 I will go over it but if you could also scan it for any nonsense that would be great

assets/utils.js
```diff
@@ -17,6 +17,16 @@ import {
     TO_BE_CONFIRMED_TEXT
 } from './agenda/utils';
 
+/*
+ * Import and load all locales that will be used in moment.js
+ * This should match the LANGUAGES defined in settings.py
+ *
+ * 'en' comes by default
+ */
+import 'moment/locale/fr-ca';
```
MarkLark86, line 26: This would need to be improved at some point, to use configured languages

newsroom/templates/download_item.txt
```diff
@@ -1,11 +1,7 @@
 {{ item.slugline }}
 {{ item.headline }}
+{{ item.source }}
```
MarkLark86, line 3: This changes the default format of text downloads. Might need this modification in the customer's theme folder instead

MarkLark86:
@petrjasek I went over the changes. Not sure if I caught everything, but it mostly looked fine

PR #38, opened 2021-12-17
[NHUB-53] fix-ui: Use Assets endpoint for Image Embeds

MarkLark86:


petrjasek (approved):
I would consider doing this on the server when fetching the item and maybe even storing the updated version, but for a short term this is fine too

PR #44, opened 2021-12-22
Accessibility work

darconny:


assets/home/components/search-bar.jsx
```diff
@@ -11,7 +11,7 @@ export class SearchBar extends React.PureComponent {
                             <i className="icon--search icon--gray-light" />
                         </span>
                         <div className="search__form input-group searchForm--active">
-                            <form className="form-inline" action="/wire">
+                            <form className="form-inline" action="/wire" role="search" aria-label="search">
```
petrjasek, line 14: there should be probably gettext used for the label

assets/user-profile/components/UserProfileApp.jsx
```diff
@@ -116,7 +116,7 @@ class UserProfileApp extends React.Component {
                                     {links.find((link) => link.active).label}
                                 </h5>
                                 <div className="content-bar__right">
-                                    <button className="content-bar__menu" aria-label="Close" onClick={this.props.hideModal}>
+                                    <button className="content-bar__menu" aria-label="Close" role="button" onClick={this.props.hideModal}>
```
petrjasek, line 119: there should be gettext for aria-label

newsroom/templates/wire_item.html
```diff
@@ -8,9 +8,9 @@
     <section class="content-header">
         <div class="content-bar navbar justify-content-between">
 
-            <span class="content-bar__menu">
+            <button class="content-bar__menu" aria-label="Close">
```
petrjasek, line 11: there should be gettext for aria-label

assets/home/components/search-bar.jsx
darconny: ok let me fix it, tnx

petrjasek (approved):

PR #50, opened 2022-01-11
CPNHUB-7 ADD: Added client management to settings.

dev-tsoni:


assets/clients/components/Clients.jsx
```diff
@@ -0,0 +1,148 @@
...
+        this.save = this.save.bind(this);
+        this.deleteClient = this.deleteClient.bind(this);
+    }
+
+    isFormValid() {
+        let valid = true;
+        let errors = {};
+
+        if (!this.props.clientToEdit.name) {
+            errors.name = ['Please provide client name'];
```
petrjasek, line 35: gettext missing

petrjasek:
pls push the backend here as well so we can test it all together

petrjasek:
would probably call that module `oauth_clients` as just `clients` is not really obvious

newsroom/clients/clients.py
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    datasource = {
+        'source': 'clients',
+        'default_sort': [('name', 1)]
+    }
+    item_methods = ['GET', 'PATCH', 'DELETE']
+    resource_methods = ['GET', 'POST']
+    mongo_prefix = MONGO_PREFIX
+    internal_resource = True
```
petrjasek, line 27: why is that internal resource? we need the api, and there would be no need for those views

assets/clients/components/EditClient.jsx
```diff
@@ -39,6 +39,13 @@ class EditClient extends React.Component {
                                     value={this.props.client.name}
                                     onChange={this.props.onChange}
                                 />
+                                {this.props.client.secret_key &&
```
petrjasek, line 42: I would add a message there:
`Make sure to copy your new client secret now. You won’t be able to see it again.`

petrjasek: and we need to make the client id visible somewhere too

assets/clients/reducers.js
```diff
@@ -0,0 +1,109 @@
...
+    case NEW_CLIENT: {
+        const newClient =  {
+            name: '',
+            secret_key: '',
+        };
+
+        return {...state, clientToEdit: newClient, errors: null};
+    }
+
+    case CANCEL_EDIT: {
```
petrjasek, line 70: we should make sure the client secret won't be there once the editor is closed

newsroom/clients/views.py
```diff
@@ -0,0 +1,87 @@
...
+
+@blueprint.route('/clients/new', methods=['POST'])
+@account_manager_only
+def create():
+    """
+    Creates the client with given client
+    """
+    client = get_json_or_400()
+
+    new_company, password = get_client_updates(client)
```
petrjasek, line 42: not sure why call the method `get_client_updates`, it's not used for updating

petrjasek (commented):
I would still rename those modules, clients doesn't really say much, api_clients or oauth_clients would be better imo

assets/clients/reducers.js
dev-tsoni: Yes, when we close the editor, the `CANCEL_EDIT` is called, making the `clientToEdit` null, and thus removing `secret_key`

assets/clients/components/EditClient.jsx
dev-tsoni: Okay, let me add them.

newsroom/auth_server/models.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self, data):
+        self._id = data["_id"]
+        self.pwd_hash = data["password"]
+
+    @property
+    def client_id(self):
+        return str(self._id)
+
+    def check_token_endpoint_auth_method(self, method):
+        return method == "client_secret_basic"
```
petrjasek, line 30: can we make this work also for `client_secret_post` ?
https://docs.authlib.org/en/latest/flask/2/grants.html#client-credentials-grant

newsroom/web/default_settings.py
```diff
@@ -385,3 +387,5 @@
 
 #: allow embargoed items on dashboard
 DASHBOARD_EMBARGOED = True
+AUTH_SERVER_EXPIRATION_DELAY = env("AUTH_SERVER_EXPIRATION_TIME", 60 * 60 * 24)  # 1 day by default
```
petrjasek, line 390: would set this rather to 4h or something

petrjasek (approved):
would be good to add some tests for this

MarkLark86:
> would probably call that module `oauth_clients` as just `clients` is not really obvious

I agree, `clients` doesn't make much sense. This goes for both the front-end and back-end file names themselves, as well as any string such as `Client Management` (could be something like `OAuth Clients` instead so admins know what this settings page is for).

newsroom/auth_server/auth.py
```diff
@@ -0,0 +1,32 @@
...
+    def check_auth(self, token, allowed_roles, resource, method):
+        """
+        This function is called to check if a token is valid. Must be
+        overridden with custom logic.
+        :param token: token.
+        :param allowed_roles: allowed user roles.
+        :param resource: resource being requested.
+        :param method: HTTP method being executed (POST, GET, etc.)
+        """
+        if not app.config.get("AUTH_SERVER_SHARED_SECRET"):
```
petrjasek, line 22: I would probably log a warning here that it's not configured and thus clients can't use it

newsroom/oauth_clients/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from flask_babel import lazy_gettext
+from .clients import ClientResource, ClientService
+
+blueprint = Blueprint('clients', __name__)
+
+from . import views   # noqa
+
+
+def init_app(app):
+    superdesk.register_resource('clients', ClientResource, ClientService, _app=app)
```
petrjasek, line 12: let's rename the resource to `oauth_clients`

petrjasek: and the settings app as well

tests/core/test_oauth_clients.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_oauth_clients(client):
+    test_login_succeeds_for_admin(client)
+    # Register a new client
+    response = client.post('/clients/new', data=json.dumps({
+        'name': 'client1',
+    }), content_type='application/json')
+
+    assert response.status_code == 201
```
petrjasek, line 13: would be good to assert the client secret is there and then use it to generate auth token

MarkLark86 (approved):

tests/core/test_oauth_clients.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+    assert response.status_code == 201
+
+    oauth_client = get_resource_service("oauth_clients").find_one(
+        req=None, name="client1"
+    )
+
+    # Check for the client secret
+    password = response.json.get("password", None)
+    if not password:
```
petrjasek, line 23: it's better to do `assert password is not None` so you get some hint in case there is an error

tests/core/test_oauth_clients.py
```diff
@@ -0,0 +1,58 @@
...
+
+    payload = {"grant_type": "client_credentials"}
+    token_auth_response = client.post(
+        "api/auth_server/token",
+        headers={"Authorization": "Basic %s" % encoded_u},
+        data=payload,
+    )
+
+    # Now use oauth token to access API
+    api_data_response = client.get(
```
petrjasek, line 39: I wonder what are you testing here, we don't use the jwt token auth for the api, or do we?

newsroom/auth_server/auth.py
```diff
@@ -13,7 +13,7 @@ class JWTAuth(TokenAuth):
     Implements JWT auth logic.
     """
 
-    def check_auth(self, token, allowed_roles, resource, method):
```
petrjasek: I think you have to keep those for the authentication to work later when hooked to eve

tests/core/test_oauth_clients.py
```diff
@@ -0,0 +1,53 @@
...
+    userpass = username + ":" + password
+    encoded_u = base64.b64encode(userpass.encode()).decode()
+
+    payload = {"grant_type": "client_credentials"}
+    token_auth_response = client.post(
+        "api/auth_server/token",
+        headers={"Authorization": "Basic %s" % encoded_u},
+        data=payload,
+    )
+    token = token_auth_response.json["access_token"]
```
petrjasek, line 34: you can check the response code on token_auth_response, that should give some hint what is wrong on the server.
and you could also add the client_id and client_secret to the payload, maybe it's an issue with the basic auth

dev-tsoni: While running locally, I am able to get `access_token` from `token_auth_response`.
And I am using `client_id` and `client_secret` to generate `encoded_u` which is used with `Authorization` in `headers`.
And It works fine locally. Just little doubt full about it running on server.

petrjasek: I see, so what do you suggest?

petrjasek (approved):

PR #64, opened 2022-01-21
NHUB-74 ADD: Better error message for reset password.

dev-tsoni:


newsroom/templates/password_reset_link_expiry.html
```diff
@@ -0,0 +1,10 @@
+{% extends "layout.html" %}
+
+{% block content %}
+<section class="content">
+    <div class="jumbotron">
+        <h1>{{ _("That's an error.") }}</h1>
+        <p class="lead">{{ _('The link in the email has already expired. Please request to reset your password again.', url=request.path) }}</p>
```
petrjasek, line 7: we could add a link to the page where this can be done

petrjasek: that url param doesn't seem very useful

petrjasek (approved):

PR #68, opened 2022-01-25
[NHUB-75] improve: Support rich text fields in agenda

MarkLark86:


assets/agenda/components/AgendaInternalNote.jsx
```diff
@@ -35,8 +37,16 @@ export default function AgendaInternalNote({internalNote, onlyIcon, noMargin, mt
         return (
             <div className={classNames('wire-column__preview_article-note', {'m-0': noMargin}, {'mt-2': mt2})}>
                 <i className="icon--info icon--red icon--info--smaller" title={labelText}/>
-                <span className='ml-1'>{internalNote.split('\n').map((item, key) =>
-                    <span key={key}>{item}<br/></span>)}</span>
+                <span className='ml-1'>
+                    {internalNote[0] !== '<' ?
+                        internalNote.split('\n').map((item, key) => (
+                            <span key={key}>{item}<br/></span>
```
petrjasek, line 43: what about using `<p>` instead of `<span>` + `<br>`

assets/utils.js
```diff
@@ -753,3 +753,19 @@ export function isActionEnabled(configKey) {
     const config = getConfig(configKey, {});
     return (action) => config[action.id] == null || config[action.id];
 }
+
+function convertHtmlToPlainText(html) {
```
petrjasek, line 757: I think this conversion has an issue for markup like:
```
<p>one</p><p>two</p>
```
will generate `onetwo`. not sure if that could be an issue in this case, but if we might be dealing with ingested content probably yes. there is this helper `getTextFromHtml` which handles it.

petrjasek (approved):

PR #69, opened 2022-01-27
CPNHUB-22 ADD: Added register button on login page.

dev-tsoni:


petrjasek:
this should be done in newsroom-app-cp and they don't use newsroom registration, it should be links as specified in the ticket

PR #84, opened 2022-02-11
[NHUB-98] Add config option to allow login from expired company

MarkLark86:


newsroom/utils.py
```diff
@@ -251,6 +251,9 @@ def is_company_enabled(user, company=None):
 
 
 def is_company_expired(company):
+    if app.config.get('ALLOW_EXPIRED_COMPANY_LOGINS'):
```
petrjasek, line 254: I wonder if we should log this for admins to check or something

MarkLark86: We have the `CompanyExpiryAlerts` celery task that sends an email to the `company_expiry_alert_recipients` setting.
Will this suffice for an admin to react to expiring companies.

What if we were to create a new report, that would indicate the number of logins for a Company, and in the table have columns such as "Last 24hrs", "Last 48hrs", "Last 7 Days", "Last 30 days" or something along those lines. This could be used not only for expiring companies but also active ones too.

petrjasek: that celery task should be enough then, we don't need to alert systems via additional logging. that login report would be good, or maybe just store some last login info on the company

petrjasek (approved):

PR #87, opened 2022-02-14
NHUB-37 ADD: Disable the deletion of user admin

dev-tsoni:


assets/users/utils.js
```diff
@@ -27,7 +27,7 @@ export function getUserTypes(user) {
 }
 
 export function isUserAdmin(user) {
-    return get(user, 'user_type') === 'administrator';
+    return get(user, 'user_type') === 'Administrator';
```
MarkLark86, line 30: This is case sensitive, and is stored with the lowercase version `administrator`

newsroom/companies/views.py
```diff
@@ -127,7 +127,6 @@ def delete(_id):
     """
     Deletes the company and users of the company with given company id
     """
-    get_resource_service('users').delete_action(lookup={'company': ObjectId(_id)})
```
MarkLark86, line 130: Not sure what you're trying to achieve here. This stops the deletion of **ALL** users when deleting a company, which means we will have users that are not associated with a company that no longer exists.

MarkLark86:
Based on the ticket, we want to provide validation errors when:
* attempting to delete a user who is an admin
* attempting to delete a company who has an associated user who is an admin

So I think this validation should go into the resources themselves,
* https://github.com/superdesk/newsroom-core/blob/develop/newsroom/companies/companies.py#L82
* https://github.com/superdesk/newsroom-core/blob/develop/newsroom/users/users.py#L112

Using the `on_delete` method, you can raise a validation error

assets/users/utils.js
dev-tsoni: I checked the value of admin user type, it was `Administrator`. This helps in differentiating between admin and non admin user. And thus the PR would show delete option only for non admin user.

MarkLark86:
@dev-tsoni There are now no changes in this PR.

assets/users/components/EditUser.jsx
```diff
@@ -129,7 +129,7 @@ function EditUser({user, onChange, errors, companies, onSave, onResetPassword, o
                         value={gettext('Save')}
                         onClick={onSave} />
 
-                    {user._id && isUserAdmin(currentUser) && <input
+                    {user._id && isUserAdmin(currentUser) && (user._id!==currentUser._id) && <input
```
petrjasek, line 132: we usually use spaces around operators `id !== currentUser`

petrjasek (changes requested):

newsroom/companies/views.py
```diff
@@ -127,7 +127,10 @@ def delete(_id):
...
+    try:
+        get_resource_service('users').delete_action(lookup={'company': ObjectId(_id)})
+    except Exception:
+        return jsonify({'company': gettext('Can not delete current user.')}), 403
```
MarkLark86, line 133: It would be better to provide this error message in the `UsersService.on_delete` method, which would allow this `try...except` to catch all types of errors and pass them onto the user.

Because of this change, it would also make sense to use a more generic key for the error instead of `company`, maybe `error`.

newsroom/users/users.py
```diff
@@ -149,3 +149,8 @@ def password_match(self, password, hashed_password):
 
     def on_deleted(self, doc):
         app.cache.delete(str(doc.get('_id')))
+
+    def on_delete(self, doc):
+        if doc.get('_id') == get_user_id():
+            raise Exception
+        app.cache.delete(str(doc.get('_id')))
```
MarkLark86, line 156: This should not be needed, as we already have this in the `on_deleted` function, which is called **after** the User is deleted

MarkLark86 (changes requested):

newsroom/companies/views.py
dev-tsoni: I have added the error message here so as to provide same error message to the UI.

MarkLark86 (approved):

PR #89, opened 2022-02-15
[NHUB-96] Improve notifications to handle more message types

MarkLark86:
Improves notifications by:
* Only sending `count` with websocket notifications, and not the `notification` or `content item`
* Only loading notifications and their associated items if notification bar is open
* Adding the ability to register new notification message types
* Add new notifications for shared Wire, Agenda and Topics, as well as when new content arrives that matches a Topic

### DIFF
--- a/assets/notifications/components/notificationItems.jsx
+++ b/assets/notifications/components/notificationItems.jsx
@@ -1,0 +1,148 @@
+import * as React from 'react';
+import {get} from 'lodash';
+
+import {gettext, shortDate} from 'utils';
+import {BasicNotificationItem} from './BasicNotificationItem';
+
+export const registeredNotifications = [];
+
+export function registerNotification(condition, component) {
+    registeredNotifications.unshift({condition, component});
+}
+
+export function renderNotificationComponent(notification, item) {
+    let notificationEntry = registeredNotifications.find(
+        (entry) => entry.condition(notification, item)
+    );
+
+    if (notificationEntry == null || notificationEntry.component == null) {
+        console.error('Unable to find Notification Item component', notification, item);
+
+        return null;
+    }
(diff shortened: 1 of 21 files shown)


assets/notifications/components/notificationItems.jsx
```diff
@@ -0,0 +1,148 @@
...
+}
+
+function getNotificationUrl(notification, item) {
+    if (get(notification, 'data.url')) {
+        return notification.data.url;
+    }
+
+    return notification.resource === 'agenda' ?
+        `/wire?item=${item._id}` :
+        `/agenda?item=${item._id}`;
```
petrjasek, line 50: looks to me like this should be the other way around

assets/notifications/components/notificationItems.jsx
```diff
@@ -0,0 +1,148 @@
...
+
+// New Wire item that matches action history (such as item Downloaded)
+registerNotification(
+    (notification, item) => (
+        (get(notification, 'action') === 'history_match' && get(notification, 'resource') === 'text') ||
+        // Fallback Wire notification (legacy functionality)
+        get(notification, 'resource') === 'text' || item.type === 'text'
+    ),
+    (notification, item) => (
+        <BasicNotificationItem
```
petrjasek, line 65: actually why not return the data rather than component? like:
```ts
(notification, item) => ({
    header: ...,
    body: ...,
    ...
})
```

MarkLark86: Returning a rendered component allows the registrant to control the look, feel and functionality of that notification entry, while providing a common `BasicNotificationItem` to use.

petrjasek (approved):

PR #91, opened 2022-02-17
NHUB-91 ADD: Copy button in OAuth client

dev-tsoni:


assets/oauth_clients/components/EditClient.jsx
```diff
@@ -51,6 +51,7 @@ class EditClient extends React.Component {
                                         name='client secret'
                                         label={gettext('Client Secret')}
                                         value={this.props.client.secret_key}
+                                        iconName='icon--copy'
```
petrjasek, line 54: imo we should rather have a property like `copyAction={true}`, it's not really about icons

petrjasek (approved):

assets/components/TextInput.jsx
```diff
@@ -17,13 +17,24 @@ function TextInput({
...
     return (
         <InputWrapper error={error} name={name}>
             {label && (
                 <label htmlFor={name}>{label}</label>
             )}
+            {copyAction &&
+                <button  
+                    className='icon-button' 
+                    onClick={(e) => {e.preventDefault();navigator.clipboard.writeText(value);}}
+                    title='Copy'
```
MarkLark86, line 32: Missing `gettext`

PR #96, opened 2022-02-23
[CPNHUB-10] fix: Use ObjectId for Navigation Company IDs

MarkLark86:
The `Management API` was specifically searching for `Company Products` using `ObjectId`, as well as updating `Navigation Companies` using `ObjectId`.

The `Web API` on the other hand was using `str` instead, so any `Company Product` permissions updated from the `Web UI` would not be returned from the `Management API`.

Instead of placing a hack fix into the `Management API`, I updated the Navigation schema to use `ObjectId`.

### DIFF
--- a/data_updates/00006_20220223-155546_products.py
+++ b/data_updates/00006_20220223-155546_products.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8; -*-
+# This file is part of Superdesk.
+# For the full copyright and license information, please see the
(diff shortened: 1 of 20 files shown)


petrjasek (approved):
I wonder if specific json_decoder could handle this automatically, but not sure if we're consistent in this so this might cause more problems at the end

MarkLark86:
> 

I believe it would be. Maybe just tackling these scenario's one at a time. For example, the Product's Navigations are stored as strings as well

PR #100, opened 2022-03-03
fix mongo index handling and make company names unique

petrjasek:
we were not setting mongo indexes in tests or the application
so there were users with same emails and that worked fine,
now this is fixes so had to fix the tests as well to avoid using
name emails or company names

NHUB-86


features/web_api/wire_search.feature
```diff
@@ -1,3 +1,4 @@
+@wip
```
MarkLark86, line 1: `@wip` should be removed

tests/test_api_tokens.py
```diff
@@ -1,50 +0,0 @@
-from flask import json
```
MarkLark86, line 1: Why are these tests removed?

petrjasek: the same file is also in core folder, so this is just duplicate - https://github.com/superdesk/newsroom-core/blob/develop/tests/core/test_api_tokens.py

MarkLark86 (approved):

PR #103, opened 2022-03-09
CPNHUB-94 ADD: "Created externally" label for non user topics.

dev-tsoni:


assets/search/components/TopicForm.jsx
```diff
@@ -39,14 +39,17 @@ const TopicForm = ({original, topic, save, onChange, globalTopicsEnabled, onSubs
                     readOnly={readOnly}
                 />
             )}
-            {!globalTopicsEnabled ? null : (
+            {!(globalTopicsEnabled && original.user) ? null : (
```
MarkLark86, line 42: This will remove the `Share with my Company` checkbox when creating a new `Topic`.

MarkLark86 (changes requested):

assets/search/components/TopicForm.jsx
dev-tsoni: When we create `topic` from UI, It is visible in `My Topics`. Then I have to click on `Share with my Company` option available to move it to `company topics`.
But when I create a topic from `API` without `User`, it get's a label `shared externally`.

MarkLark86: The user should be able to create the Topic as a Company Topic when they initially create.
Therefor this checkbox should be available when creating a topic in the UI, not just updating the topic

dev-tsoni: Ohh, okay. that means it is totally unnecessary.

MarkLark86: No it is necessary to remove the checkbox, if the topic is an existing topic

MarkLark86 (approved):

PR #106, opened 2022-03-15
CPNHUB-94 ADD: Added text "by CP via API".

dev-tsoni:


petrjasek:
I would rather not do this, that string would only work for CP, will comment on the jira ticket

petrjasek (changes requested):
pls use gettext

assets/components/AuditInformation.jsx
```diff
@@ -48,7 +48,7 @@ class AuditInformation extends React.Component {
                 <div className='wire-column__preview__date'>
                     {gettext('Updated {{updator}}at {{updated}}',
                         {
-                            updator: updator ? gettext(`by ${updator.first_name} ${updator.last_name} `) : 'by System ',
+                            updator: updator ? gettext(`by ${updator.first_name} ${updator.last_name} `) : gettext('by System '),
```
petrjasek, line 51: imo the space shouldn't be part of the gettext call, that might get removed on transifex

petrjasek (approved):

PR #110, opened 2022-03-21
NHUB-103 ADD: Last login for oauth-clients

dev-tsoni:


newsroom/auth_server/oauth2.py
```diff
@@ -32,6 +33,14 @@
...
 def issue_token():
+    current_time = utcnow()
+
+    client_id = request.authorization.get("username")
+    if client_id:
+        client = get_cached_resource_by_id("oauth_clients", client_id)
+        superdesk.get_resource_service("oauth_clients").system_update(
+            ObjectId(client_id), {"last_active": current_time}, client
+        )
     return authorization.create_token_response()
```
petrjasek, line 44: not sure if this can fail in general, but imo would be better to do it like:
```
try:
    return authorization.create_token_response()
else:
    log the time
```

assets/oauth_clients/components/ClientListItem.jsx
```diff
@@ -11,6 +11,9 @@ function ClientListItem({client, onClick}) {
         >
             <td className="name">{client.name}</td>
             <td>{shortDate(client._created)}</td>
+            {(client.last_active) && (
+                <td>{shortDate(client.last_active)}</td>
```
petrjasek, line 15: I think the `<td>` should always be there not to break the layout

newsroom/auth_server/oauth2.py
```diff
@@ -32,7 +33,19 @@
 
 @blueprint.route(TOKEN_ENDPOINT, methods=["POST"])
 def issue_token():
-    return authorization.create_token_response()
+    current_time = utcnow()
+    try:
+        token_response = authorization.create_token_response()
+    except Exception:
+        pass
```
petrjasek, line 40: not sure why ignoring the exception?

petrjasek (approved):

PR #112, opened 2022-03-24
NHUB-103 FIX: token generation working for both type of client authentication.

dev-tsoni:
…tication.


petrjasek:
pls add a test for authentication with `client_id` and `client_secret` in the payload

petrjasek (approved):

PR #117, opened 2022-03-30
NHUB-105 ADD: "News only" and "Top/Featured Stories" configurable.

dev-tsoni:


newsroom/agenda/views.py
```diff
@@ -91,6 +91,13 @@ def search():
     return send_response('agenda', response)
 
 
+@blueprint.route('/agenda/count')
```
petrjasek, line 94: it should be probably `agenda_featured/count` and not just `agenda` that's a bit confusing

newsroom/agenda/views.py
```diff
@@ -91,6 +91,13 @@ def search():
...
+@login_required
+def get_agenda_featured_item_count():
+    items = get_resource_service('agenda_featured').get_all()
```
petrjasek, line 97: fetching all seems a bit expensive to realize if there are items, imo we can just try to find one and either it's there or not. also making an api call for it seems unnecessary, it could be part of the view data so you would have that info there from the start and we wouldn't need extra action/reducer for this

dev-tsoni: I am not able to find any relevant data that could tell me if featured item exists inside `initData`. It might contain `featured` boolean variable though.
Also, there is one more thing, here I am returning the count of all the available items, irrespective of date. Do we need this toggle to be shown as per the date in the filter? Like if I select today's date, so if there are any featured items available for today, the toggle will be shown, else not?

newsroom/settings.py
```diff
@@ -77,8 +77,7 @@ def get_client_config():
     for key, setting in (get_setting() or {}).items():
         if key not in ['_updated', 'version_creator']:
             value = setting.get('value', setting.get('default'))
-            if value:
```
dev-tsoni, line 80: Also @petrjasek Is it safe here to remove this check. It was necessary for assigning null value for `newsonly` filter. It didn't caused any issues locally though.

MarkLark86:
@dev-tsoni We already have a config for the displaying of the 'News Only' toggle.
* https://github.com/superdesk/newsroom-core/blob/develop/newsroom/web/default_settings.py#L302
* https://github.com/superdesk/newsroom-core/blob/develop/assets/utils.js
* https://github.com/superdesk/newsroom-core/blob/develop/assets/wire/components/WireApp.jsx#L187

It would be good to add the config for the Agenda/FeaturedStories toggle in the same way as the above 3 files.

dev-tsoni:
> @dev-tsoni We already have a config for the displaying of the 'News Only' toggle.
> 
> * https://github.com/superdesk/newsroom-core/blob/develop/newsroom/web/default_settings.py#L302
> * https://github.com/superdesk/newsroom-core/blob/develop/assets/utils.js
> * https://github.com/superdesk/newsroom-core/blob/develop/assets/wire/components/WireApp.jsx#L187
> 
> It would be good to add the config for the Agenda/FeaturedStories toggle in the same way as the above 3 files.
> 
> A config for this might be good. If a customer started to use Featured Stories once, this toggle will be there forever.

Alright, I'll try to use and update them as well.

newsroom/agenda/views.py
petrjasek: I mean you would need to add it there instead of doing that query.
so in this `get_view_data` function you could add:
```
    'has_agenda_featured_items': get_resource_service('agenda_featured').find_one(req=None) is not None,
```

petrjasek (approved):

PR #118, opened 2022-03-30
[NHUB-117] Initial migration to Bootstrap 5

darconny:
-- DO NOT MERGE --
I need help for fixing the errors and then I can start with UI refactoring regarding to new BS classes and RTL support. I suppose we keep this work separate from develop branch until we finish most of the work.
@petrjasek and @MarkLark86 pls advise


package.json
```diff
@@ -35,15 +36,14 @@
     "moment": "2.23.0",
     "moment-timezone": "0.5.23",
     "node-sass": "^4.5.3",
-    "popper.js": "1.14.4",
     "react": "16.6.0",
-    "react-bootstrap-typeahead": "3.2.4",
-    "react-datepicker": "1.8.0",
+    "react-bootstrap-typeahead": "5.2.2",
```
MarkLark86, line 40: Upgrading these libraries will require testing for any breaking changes

assets/styles/labels.scss
```diff
@@ -72,6 +72,7 @@ $labels-palette: (
 // Generate label modifiers for each theme colour defined
 @each $color, $value in $theme-colors {
     .label--#{$color} {
-        @include alert-variant(theme-color-level($color, $alert-bg-level), theme-color-level($color, $alert-border-level), theme-color-level($color, $alert-color-level));
+        // TODO: Migrate this to a Bootstrap v5 variant
+        //@include alert-variant(theme-color-level($color, $alert-bg-level), theme-color-level($color, $alert-border-level), theme-color-level($color, $alert-color-level));
```
MarkLark86, line 76: @darconny This was causing build errors, second argument is not a colour. Commenting out so it can at least build

webpack.config.js
```diff
@@ -42,7 +42,10 @@ const config = {
         rules: [
             {
                 test: /\.jsx?$/,
-                include: path.resolve(__dirname, 'assets'),
+                include: [
+                    path.resolve(__dirname, 'assets'),
+                    path.resolve(__dirname, 'node_modules/bootstrap')
```
petrjasek, line 47: why we need this?

MarkLark86: The module files for Bootstrap include a spread operator. Without compilation, these fail

MarkLark86:
The source files build, but now there are runtime errors (jquery functions such as `.tooltip` don't exist)

MarkLark86:
We should probably remove the usage of Jquery. Bootstrap 5 removed it from theirs, and we can replace our usage with pure JS APIs to do the same. This would resolve the runtime errors with things like `.tooltip`

webpack.config.js
petrjasek: I see

MarkLark86: Hopefully with removing jQuery, and importing classes like Tooltip directly may fix this. Will look into it tomorrow

assets/styles/labels.scss
darconny: yes that is fine, to make the app work and we do a refactoring later on

MarkLark86:
@darconny I have fixed most of the runtime errors. Newshub now works except for Agenda and a couple of the Reports.
At least you can now run the UI locally and continue your work.
On Monday I will continue on fixing the remaining errors.

webpack.config.js
MarkLark86: @petrjasek Specifically importing the Bootstrap modules directly did not work, it is still importing the dist files that include the spread operator.

This project to upgrade to Bootstrap v5 also gives us the opportunity to upgrade other packages.
* Babel v7 / preset-env (and target the last x browser versions)
* Webpack v5

This might fix the issue with Bootstrap v5, as newer Browsers support the spread operator (and Bootstrap v5 no longer supports older Browsers).
What are your thoughts?

darconny:
@MarkLark86 That's great thanks a lot.

PR #120, opened 2022-04-04
CPNHUB-94 FIX: subscribers required true.

dev-tsoni:


newsroom/topics/topics.py
```diff
@@ -19,7 +19,8 @@ class TopicsResource(newsroom.Resource):
         'is_global': {'type': 'boolean', 'default': False},
         'subscribers': {
             'type': 'list',
-            'schema': newsroom.Resource.rel('users', required=True)
+            'schema': newsroom.Resource.rel('users'),
+            'required': True
```
petrjasek, line 23: why are subscribers required?

assets/components/AuditInformation.jsx
```diff
@@ -28,7 +28,7 @@ class AuditInformation extends React.Component {
         const users = get(this.props, 'users.length') ? this.props.users : this.props.editUsers;
 
         if (field === '_created' && item._created) {
-            const creator = getItemFromArray(item.original_creator,
+            const creator = getItemFromArray(item.users,
```
petrjasek, line 31: why this change?

newsroom/topics/topics.py
dev-tsoni: There was an issue while updating the topics. Where subscribers was required. Fixed it and updated the PR.

assets/components/AuditInformation.jsx
dev-tsoni: This was done earlier when `original_creator` was not updated when we created topics via API. Since now it is updated as per the user, this is not required.

petrjasek (approved):

PR #127, opened 2022-04-26
remove cs translation

petrjasek:


MarkLark86 (approved):

MarkLark86:
@petrjasek I'm guessing these translations are either incomplete or no longer maintained, is that the reason why these are being removed?

petrjasek:
exactly

PR #129, opened 2022-04-27
CPNHUB-109 ADD: Gettext for translation to be effective.

dev-tsoni:


assets/agenda/components/AgendaCoverageExistsFilter.jsx
```diff
@@ -26,13 +26,13 @@ function AgendaCoverageExistsFilter ({toggleFilter, activeFilter}) {
             <button
                 key='coverage-planned'
                 className='dropdown-item'
-                onClick={() => toggleFilter(filter.field, 'planned')}
+                onClick={() => toggleFilter(filter.field, gettext('planned'))}
```
MarkLark86, line 29: In this context, `toggleFilter` expects the same string that the server uses to construct the query https://github.com/superdesk/newsroom-core/blob/develop/newsroom/agenda/agenda.py#L466

Therefor using `gettext` here will cause a bug.

MarkLark86 (changes requested):

assets/agenda/components/AgendaCoverageExistsFilter.jsx
dev-tsoni: The text was not being translated into French, adding `gettext` is doing the job. I am not sure about its aftereffects.

MarkLark86: The server is expecting the value `planned` when constructing the Elasticsearch query.
Using `gettext` here will change this value to something else, if English has not been selected

assets/components/DropdownFilterButton.jsx
```diff
@@ -4,7 +4,14 @@ import {gettext} from 'utils';
 import classNames from 'classnames';
 
 const getActiveFilterLabel = (filter, activeFilter, isActive) => {
-    return isActive ? activeFilter[filter.field][0] : gettext(filter.label);
+    let filterLabel = '';
+    if (isActive) {
+        filterLabel = filter.transform  ? gettext(filter.transform(activeFilter[filter.field][0])) : gettext(activeFilter[filter.field][0]);
```
petrjasek, line 9: imo just returning from here and the else block instead of using a variable for that would make it a bit simpler

petrjasek (approved):

assets/components/DropdownFilterButton.jsx
```diff
@@ -4,7 +4,12 @@ import {gettext} from 'utils';
 import classNames from 'classnames';
 
 const getActiveFilterLabel = (filter, activeFilter, isActive) => {
-    return isActive ? activeFilter[filter.field][0] : gettext(filter.label);
+    if (isActive) {
+        return filter.transform ? gettext(filter.transform(activeFilter[filter.field][0])) : gettext(activeFilter[filter.field][0]);
```
petrjasek, line 8: not sure we need to call that gettext together with `filter.transform`, that should be handled there then no?

PR #134, opened 2022-05-02
CPNHUB-94 FIX: Added validation check for topic type required.

dev-tsoni:


newsroom/topics/topics.py
```diff
@@ -22,7 +22,7 @@ class TopicsResource(newsroom.Resource):
             'schema': newsroom.Resource.rel('users', required=True),
         },
         'timezone_offset': {'type': 'integer', 'nullable': True},
-        'topic_type': {'type': 'string', 'nullable': True},
+        'topic_type': {'type': 'string', 'required': True},
```
MarkLark86, line 25: I think setting an `allowed` list here would help validation, so the API client doesn't create an unsupported topic of type.
Allowed values should be `wire` or `agenda`

MarkLark86 (approved):

PR #139, opened 2022-05-04
fix: Migration script fails if user was deleted

MarkLark86:


data_updates/00003_20211118-181158_topics.py
```diff
@@ -22,8 +22,15 @@ def forwards(self, mongodb_collection, mongodb_database):
         }
 
         for topic in mongodb_collection.find({}):
+            topic_id = topic.get(config.ID_FIELD)
+            user_id = topic.get('user')
+
+            if not users.get(str(user_id)):
```
petrjasek, line 28: that lookup won't work if the `user['_id']` is `ObjectId`

PR #142, opened 2022-05-06
[NHUB-127] Make FilterPanel open/tab defaults configurable

MarkLark86:
Also remember open state in browser local storage

### DIFF
--- a/newsroom/web/default_settings.py
+++ b/newsroom/web/default_settings.py
@@ -305,6 +305,16 @@ CLIENT_CONFIG = {
     'item_actions': {},
     'display_abstract': DISPLAY_ABSTRACT,
     'display_credits': False,
+    'filter_panel_defaults': {
+        'tab': {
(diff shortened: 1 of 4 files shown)


newsroom/web/default_settings.py
```diff
@@ -305,6 +305,16 @@
     'item_actions': {},
     'display_abstract': DISPLAY_ABSTRACT,
     'display_credits': False,
+    'filter_panel_defaults': {
```
petrjasek, line 308: I wonder if we should have it here or in `ui_config` db, what would be the rule of thumb here?

petrjasek (approved):

newsroom/web/default_settings.py
MarkLark86: I was looking at that today, we should improve the way we're doing config for `client` side.
As we have `client_config` in `settings.py`, and `ui_config` in `Mongo`.

I think for these defaults it doesn't really matter where, as after the user logs in it's all stored in browser local storage.

PR #144, opened 2022-05-06
make elastic count above 10k

petrjasek:
NHUB-128

### DIFF
--- a/newsroom/web/default_settings.py
+++ b/newsroom/web/default_settings.py
@@ -366,6 +366,9 @@ ELASTICSEARCH_SETTINGS.setdefault("settings", {})["query_string"] = {
     'analyze_wildcard': False
 }
 
+# count above 10k
+ELASTICSEARCH_TRACK_TOTAL_HITS = True
+
 #: server working directory
 #: should be set in settings.py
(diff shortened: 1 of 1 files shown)


newsroom/web/default_settings.py
```diff
@@ -366,6 +366,9 @@
     'analyze_wildcard': False
 }
 
+# count above 10k
+ELASTICSEARCH_TRACK_TOTAL_HITS = True
```
MarkLark86, line 370: I wonder how this would affect customers with large number of documents & large queries.
What are your thoughts on leaving default to 10k, and changing this to `True` per customer?
In the UI, if the count is `10,000` then maybe we show `10,000+` to indicate there are more (if the customer(s) are okay with that experience)

petrjasek: I did some testing locally and didn't see any effect so would keep it on by default with the option to turn it off if needed.
We have it on in superdesk and it was never an issue afaik.

MarkLark86 (approved):

newsroom/web/default_settings.py
MarkLark86: Sounds good

PR #149, opened 2022-05-12
execute scripts in body html

petrjasek:
so newsroom can render embeds using those.
only executing scripts with src not inline js.

NHUB-119

### DIFF
--- a/assets/ui/components/ArticleBodyHtml.jsx
+++ b/assets/ui/components/ArticleBodyHtml.jsx
@@ -17,10 +17,13 @@ class ArticleBodyHtml extends React.PureComponent {
 
         // use memoize so this function is only called when `body_html` changes
         this.getBodyHTML = memoize(this._getBodyHTML.bind(this));
+
+        this.bodyRef = React.createRef();
     }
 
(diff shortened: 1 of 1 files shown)


assets/ui/components/ArticleBodyHtml.jsx
```diff
@@ -66,6 +70,43 @@ class ArticleBodyHtml extends React.PureComponent {
...
+
+                script.src = url;
+                script.async = true;
+
+                script.onload = () => {
+                    document.body.removeChild(script);
+                };
+
+                script.onerrror = (error) => {
+                    throw new URIError('The script ' + error.target.src + 'didn\'t load.');
```
MarkLark86, line 102: It might be good to include `error` in the exception raised, to make debugging easier

petrjasek: you will see the error in the network tab

MarkLark86: Hah, good point

MarkLark86 (approved):

PR #158, opened 2022-05-23
SDCP-621 ADD: Hide company topic feature.

dev-tsoni:


MarkLark86:
@dev-tsoni This feature already exists since v2.0, https://github.com/superdesk/newsroom-core/pull/46

You will see in some of the files you changed, a variable called `globalTopicsEnabled`:
https://github.com/superdesk/newsroom-core/blob/develop/assets/search/components/TopicsTab.jsx#L43

The config is configurable per section, i.e. `Wire`, `Agenda` and/or `Monitoring`

In CP-LJI instance you can change this config in these two places to disable it for them:
* https://github.com/superdesk/newsroom-app-cp-lji/blob/master/server/data/ui_config.json#L62
* https://github.com/superdesk/newsroom-app-cp-lji/blob/master/server/data/ui_config.json#L124

PR #164, opened 2022-05-27
[NHUB-132] Add profile setting for inApp notifications

MarkLark86:


petrjasek (approved):
maybe we could actually use the browser notifications if we have special toggle for this, but as an improvement later

PR #167, opened 2022-06-02
NHUB-136 ADD: Renamed navigation to global topic in settings.

dev-tsoni:


assets/navigations/components/EditNavigation.jsx
```diff
@@ -19,7 +19,7 @@ class EditNavigation extends React.Component {
         this.state = {activeTab: 'navigation-details'};
 
         this.tabs = [
-            {label: gettext('Navigation'), name: 'navigation-details'},
+            {label: gettext('Global Topic'), name: 'navigation-details'},
```
petrjasek, line 22: imo this one should be `Global Topics`

dev-tsoni: image
Here, since we are editing single global topic, I kept it singular.

petrjasek (approved):

PR #170, opened 2022-06-09
NHUB-150 ADD: Source filter added to my topics.

dev-tsoni:


assets/search/components/TopicParameters.jsx
```diff
@@ -60,11 +61,12 @@ const TopicParameters = ({topic, navigations, locators, filterGroupLabels}) => {
             {renderParam(gettext('Search'), get(topic, 'query') ? [topic.query] : [])}
             {renderParam(gettext('Date Created'), dateLabels)}
             {renderParam(gettext('Topics'), navs)}
-            {renderParam(get(filterGroupLabels, 'service', gettext('Category')), filters.service)}
-            {renderParam(get(filterGroupLabels, 'subject', gettext('Subject')), filters.subject)}
-            {renderParam(get(filterGroupLabels, 'genre', gettext('Content Type')), filters.genre)}
-            {renderParam(get(filterGroupLabels, 'urgency', gettext('Urgency')), filters.urgency)}
-            {renderParam(get(filterGroupLabels, 'source', gettext('Source')), filters.source)}
+            {renderParam(get(filterGroupLabels, 'service', gettext('Category')),  wireAggskeys.includes('service') ? filters.service : [])}
```
MarkLark86, line 64: We want to remove this hard coding of filter groups, as these change from customer to customer.
What you need to use here instead is the selector `assets/search/selectors:filterGroups` https://github.com/superdesk/newsroom-core/blob/develop/assets/search/selectors.js#L27

This will give you the list of filter groups configured for the customer specific instance, as well as the page (Wire & Agenda), which this PR does not address

MarkLark86 (approved):

PR #171, opened 2022-06-09
fix sendmail error

petrjasek:
we were using custom policy which doesn't do line breaks.

https://docs.python.org/3.8/library/email.policy.html#email.policy.HTTP

CPNHUB-132

### DIFF
--- a/newsroom/email.py
+++ b/newsroom/email.py
@@ -3,10 +3,9 @@ from typing import List, Optional, Dict, Any
 from typing_extensions import TypedDict
 
 from superdesk import get_resource_service
-from superdesk.emails import SuperdeskMessage  # it handles some encoding issues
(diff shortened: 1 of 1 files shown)


petrjasek:
I might also fix superdesk-core but want to test this one first

MarkLark86:
@petrjasek Does this affect SDESK-720 (https://github.com/superdesk/superdesk-core/pull/794), as the solution to this ticket was to use `policy=email.policy.HTTP`

petrjasek:
yep it's reverting that change, but that was 5 years ago so we have a different python version now, but still we need to test that usecase

petrjasek:
@MarkLark86 switched to utf8 smtp policy

MarkLark86 (approved):

PR #174, opened 2022-06-17
[CPNHUB-124] fix: Show Event description in email

MarkLark86:


newsroom/templates/email_agenda.txt
```diff
@@ -8,6 +8,8 @@
 {% endif %}{% if links | length is gt 1 %}Web: {% for link in links %}{{ link }}
 {% endfor %}
 {% endif %}
+{% if item.definition_long or item.definition_short %}item.definition_long or item.definition_short
```
petrjasek, line 11: there should be `{{ item.definition_long or item.definition_short }}` imo

petrjasek (approved):

PR #176, opened 2022-06-21
[STTNHUB-162]: Store Planning item separately on push

MarkLark86:


newsroom/push.py
```diff
@@ -251,60 +257,69 @@ def get_event_dates(event):
...
+    if not orig_agendas.count():
+        # event id exists in planning item but event is not in the system
+        logger.warning('Event {} for planning {} couldn\'t be found'.format(planning['event_item'], planning))
+        return None
+
+    orig_agenda = orig_agendas[0]
+    agenda = deepcopy(orig_agendas[0])
+
+    if planning.get("state") in [WORKFLOW_STATE.CANCELLED, WORKFLOW_STATE.KILLED] or \
+            planning.get("pubstatus") == "cancelled":
```
petrjasek, line 302: not sure what we use in planning, but in sd this is actually `canceled` as in [ninjs standard](http://www.iptc.org/std/ninjs/ninjs-schema_1.1.json)

petrjasek (approved):

newsroom/push.py
MarkLark86: No worries, I'll double check the Event JSON formatter

MarkLark86: I don't know the reason why, but in Planning we use `cancelled`:
* https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/common.py#L71
* https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/common.py#L78
* https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning.py#L1552

petrjasek: I see, we should harmonize that eventually but that's for another PR

PR #179, opened 2022-06-24
[CPNHUB-125] fix(email): Missing coverage date, language and empty map image

MarkLark86:
Also improve distinction between Event and Planning descriptions/sections

### DIFF
--- a/newsroom/agenda/utils.py
+++ b/newsroom/agenda/utils.py
@@ -125,8 +125,10 @@ def get_coverage_status_text(coverage):
         return 'in progress at {}.'.format(get_coverage_scheduled_date(coverage))
 
     if coverage.get('workflow_status') == ASSIGNMENT_WORKFLOW_STATE.COMPLETED:
-        return '{}{}.'.format('updated' if len(coverage.get('deliveries') or []) > 1 else 'available',
-                              ' at ' + get_coverage_publish_time(coverage))
+        return '{}{}.'.format(
+            'updated' if len(coverage.get('deliveries') or []) > 1 else 'available',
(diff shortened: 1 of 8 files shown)


newsroom/agenda/utils.py
```diff
@@ -125,8 +125,10 @@ def get_coverage_status_text(coverage):
         return 'in progress at {}.'.format(get_coverage_scheduled_date(coverage))
 
     if coverage.get('workflow_status') == ASSIGNMENT_WORKFLOW_STATE.COMPLETED:
-        return '{}{}.'.format('updated' if len(coverage.get('deliveries') or []) > 1 else 'available',
-                              ' at ' + get_coverage_publish_time(coverage))
+        return '{}{}.'.format(
+            'updated' if len(coverage.get('deliveries') or []) > 1 else 'available',
```
petrjasek, line 129: we should use gettext on those strings

MarkLark86: As this is used in templates, `gettext` won't work here unless we programatically change the gettext language for this email request.

It is for this reason I moved this into a template for FR-CA:
https://github.com/superdesk/newsroom-app-cp/blob/develop/server/theme/agenda_status_text.txt

petrjasek: I see ok

petrjasek (approved):

newsroom/agenda/utils.py
MarkLark86: But I really dislike constructing strings like these in python for jinja templates. We should look at this at some point

PR #181, opened 2022-06-28
[STTNHUB-163] Provide Planning Only filter in Agenda

MarkLark86:


assets/agenda/components/AgendaEventsOnlyFilter.jsx
```diff
@@ -3,39 +3,57 @@ import PropTypes from 'prop-types';
 import {gettext} from 'utils';
 import AgendaFilterButton from './AgendaFilterButton';
 
-function AgendaEventsOnlyFilter ({toggleFilter, eventsOnlyView}) {
-    const eventOnlyLabel = gettext('Show Events Only');
+function AgendaEventsOnlyFilter ({toggleFilter, itemTypeFilter}) {
```
petrjasek, line 6: we might also rename the component

petrjasek (approved):

assets/agenda/components/AgendaEventsOnlyFilter.jsx
MarkLark86: Good point. I'll do that tomorrow

PR #190, opened 2022-07-21
add saml auth support

petrjasek:
CPNHUB-137


newsroom/auth/utils.py
```diff
@@ -0,0 +1,57 @@
+import flask
+import werkzeug
+import superdesk
+
+from typing import Optional, TypedDict
+from flask_babel import _
+from superdesk.utc import utcnow
+
+
+class UserData(TypedDict, total=False):
```
MarkLark86, line 10: This might be better located in `newsroom/users/users.py` file, as this is where the `UsersResource` is defined

MarkLark86 (approved):

PR #193, opened 2022-08-02
[STTNHUB-171] Ability to configure nested aggregation search on a field

MarkLark86:
Enabled using `AGENDA_GROUPS` config, should be able to extend for use with `Wire` later on

### DIFF
--- a/newsroom/search_config.py
+++ b/newsroom/search_config.py
@@ -1,0 +1,117 @@
+from typing import List, Dict, Any, KeysView, Type
+import logging
+from copy import deepcopy
+
+from superdesk.resource import not_analyzed
+from newsroom.factory.app import BaseNewsroomApp
+from newsroom import Resource
+
+
+logger = logging.getLogger(__name__)
+nested_agg_groups = {}
+
+
+def is_search_field_nested(resource_type: str, field: str):
+    """Returns ``True`` if the ``resource_type`` is configured for nested search group"""
+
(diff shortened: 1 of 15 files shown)


newsroom/search_config.py
```diff
@@ -0,0 +1,117 @@
...
+        original = {}
+
+    new_schema: Dict[str, Any] = {
+        "type": "nested",
+        # Flatten the field in the parent object too
+        # This is so non-nested queries/aggregations still work
+        # (such as in existing Topics)
+        "include_in_parent": True,
+        "properties": (original.get("mapping") or {}).get("properties") or {
+            "qcode": not_analyzed,
```
petrjasek, line 83: I think we're using rather `"code"` there

newsroom/search_config.py
```diff
@@ -0,0 +1,117 @@
...
+nested_agg_groups = {}
+
+
+def is_search_field_nested(resource_type: str, field: str):
+    """Returns ``True`` if the ``resource_type`` is configured for nested search group"""
+
+    return field in (nested_agg_groups.get(resource_type) or {})
+
+
+def init_nested_aggregation(
```
petrjasek, line 20: just wondering why doing it dynamically, imo the groups config is something we might set/change later but would be good to have the mapping on an instance from the start so we don't have to reindex when changing the config

petrjasek (approved):

PR #194, opened 2022-08-11
avoid long lines in email body

petrjasek:
according to RFC 5322 any line in an email can't be longer than 998
characters so try to pretty print the html which reformats it
instead of printing everything on a single line, also enforce
line limit on text via splitting longer lines.

CPNHUB-132


MarkLark86 (approved):

marwoodandrew (approved):
Better than my attempt! Should maybe done in Superdesk as well!

petrjasek:
right @marwoodandrew will check how it works and put it there eventually

PR #199, opened 2022-09-06
[NHUB-165] Support nested agg search on wire items

MarkLark86:


newsroom/commands/fix_topic_nested_filters.py
```diff
@@ -0,0 +1,142 @@
...
+from .manager import manager
+
+
+@manager.command
+def fix_topic_nested_filters():
+    """Fix My/Company Topics after adding ``Nested Agg`` to Wire/Agenda group configs
+
+    Previously ``subject`` was used to provide a filter for all custom subjects (such as CPs ``distribution``).
+    After removing ``subject`` from ``WIRE_GROUPS`` config, Topics will need to be updated to reflect this change.
+    This command will move the ``subject`` filter values into their newly configured ``Nested Agg` group.
```
petrjasek, line 20: just to check if I get this right - the issue is that those saved searches wouldn't match the values in aggregations once we reshuffle those groups? so it wouldn't be visible what was selected there?

petrjasek (approved):

PR #202, opened 2022-09-13
[NHUB-140]: New command to purge expired Agenda items

MarkLark86:


newsroom/commands/remove_expired_agenda.py
```diff
@@ -0,0 +1,179 @@
...
+            }),
+        }
+
+        items = list(agenda_service.internal_get(req=req, lookup=None))
+
+        if not len(items):
+            break
+
+        for item in items:
+            ids_processed.add(item[config.ID_FIELD])
```
petrjasek, line 137: not sure what's the use for this

petrjasek (approved):

newsroom/commands/remove_expired_agenda.py
MarkLark86: Ah, forgot that one. I re-wrote that es query multiple times.
I should use that and add it into the query so the next request does not include the items already processed. Issue here though is the query may become too big if I keep adding in these IDs.
Will fix it tomorrow.

MarkLark86: @petrjasek I removed this variable entirely. Considering we're talking about Events & their associated Coverages, I don't foresee a case where the Coverage would be 60+ days after the Event's scheduled end date (based on example `AGENDA_EXPIRY_DAYS` config).

PR #208, opened 2022-09-16
[STTNHUB-177] fix(agenda_search): Include Planning items when running query filters

MarkLark86:


newsroom/agenda/agenda.py
```diff
@@ -684,6 +704,25 @@ def get(self, req, lookup):
...
+
+        if any([key in planning_filters for key in (json.loads(req.args.get("filter") or "{}")).keys()]):
+            return set()
+
+        orig_args = req.args
+        req.args = dict(req.args)
+        req.args["itemType"] = "events"
+        item_ids = set([
+            item["_id"]
+            for item in super().get(req, lookup)
```
petrjasek, line 721: imo we should do some projection here if we only need ids

petrjasek (approved):

petrjasek (approved):
just not sure if we should be sending that extra data in `_links`, maybe `_hits` would be better to make it more obvious it's related to the search

PR #216, opened 2022-10-11
fix failing email notifications due to line length

petrjasek:
iter wasn't going inside element so make it recursive

CPNHUB-132


newsroom/email.py
```diff
@@ -64,11 +64,11 @@ def handle_long_lines_html(html):
         return html
     parsed = etree.fromstring(html, parser=etree.HTMLParser())
     etree.indent(parsed, space=" ")  # like pretty print but upfront
-    for elem in parsed.iter():
-        if elem.text is not None and len(elem.text) > MAX_LINE_LENGTH:
-            elem.text = handle_long_lines_text(elem.text) + "\n"
-        if elem.tail is not None and len(elem.tail) > MAX_LINE_LENGTH:
-            elem.tail = handle_long_lines_text(elem.tail) + "\n"
+    for elem in parsed.iter("*"):
+        if elem.text is not None and len(elem.text) > 100:
```
MarkLark86, line 68: Where is this `100` coming from? Is it just a safeguard, using a low number, compared to `MAX_LINE_LENGTH`?

MarkLark86 (approved):

newsroom/email.py
petrjasek: yep exactly, as those tags can be nested it needs some buffer to handle that

PR #220, opened 2022-10-25
refactor auth code

petrjasek:
- remove duplicated code in view/utils
- move utils from view to utils
- avoid using cache when authenticating, but add `auth_ttl` to session and consider it valid until it expires (then check again via db)
- add helpers for session handling

### DIFF
--- a/newsroom/decorator.py
+++ b/newsroom/decorator.py
@@ -1,15 +1,18 @@
 from functools import wraps
-from flask import request, redirect, url_for, session, abort
-from newsroom.utils import is_valid_login
+from flask import request, redirect, url_for, abort
+from newsroom.auth.utils import (
+    clear_user_session,
+    is_current_user_account_mgr,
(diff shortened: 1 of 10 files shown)


newsroom/decorator.py
```diff
@@ -33,10 +32,7 @@ def decorated_function(*args, **kwargs):
 def account_manager_only(f):
     @wraps(f)
     def decorated_function(*args, **kwargs):
-        if session.get("user_type") not in [
-            "administrator",
-            "account_management",
-        ] or not is_valid_login(session.get("user")):
+        if not is_current_user_admin() and not is_current_user_account_mgr() or not is_valid_session():
```
MarkLark86, line 35: It might be good to group these first two conditions in parentheses, so it makes it easier to read the order of operations.

MarkLark86 (approved):

PR #223, opened 2022-10-27
[NHUB-193] Improve location filter by supporting regions and places

MarkLark86:


newsroom/agenda/views.py
```diff
@@ -287,3 +287,145 @@ def related_wire_items(wire_id):
...
+
+    if not query.startswith("*") and not query.endswith("*"):
+        query = f"*{query}*"
+
+    def gen_agg_filter(field: str):
+        return {
+            "bool": {
+                "must": [
+                    {
+                        "query_string": {
```
petrjasek, line 305: might be better to use `match` query here or `match_bool_prefix`

newsroom/agenda/views.py
```diff
@@ -287,3 +287,145 @@ def related_wire_items(wire_id):
...
+        regions.append(
+            {
+                "name": country_bucket["key"],
+                "type": "country",
+            }
+        )
+
+    return (
+        flask.jsonify(
+            {
```
petrjasek, line 425: if you return just dict it's the same as using jsonify

petrjasek (approved):

newsroom/agenda/views.py
MarkLark86: Using `match` or `match_bool_prefix` requires to normalise the terms, as to provide case insensitive search. These searches also do not allow to partial match (only `startswith` match), where as `query_string` provides this and much more (although most functionality won't be used by the general public users).

To improve performance, I have removed the `query_string` filters if there is no search text provided.

PR #224, opened 2022-11-08
handle all day events and events with no end date

petrjasek:
fixes:
- on push avoid key error with missing tz
- date based queries handle differently for all_day events
- fix UI not to display time when it shouldn't

NHUB-197

### DIFF
--- a/assets/utils.js
+++ b/assets/utils.js
@@ -338,6 +338,11 @@ export function getScheduleType(item) {
     const start = moment(item.dates.start);
     const end = moment(item.dates.end);
     const duration = end.diff(start, 'minutes');
+
+    if (item.dates.all_day) {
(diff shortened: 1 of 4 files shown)


assets/utils.js
```diff
@@ -401,54 +408,81 @@ export function formatAgendaDate(item, group, localTimeZone = true) {
...
+    const endDate = formatDate(end);
+    const endTime = formatTime(end);
+    const timezone = localTimeZone ? '' : getFormattedTimezone(start);
+
+    switch (true) {
+    case isTBCItem && startDate !== endDate:
+        return gettext('{{startDate}} to {{endDate}} (Time to be confirmed)', {
+            startDate,
+            endDate,
+            timeToBeConfirmed: TO_BE_CONFIRMED_TEXT,
```
MarkLark86, line 427: `timeToBeConfirmed` not used in translation string

MarkLark86 (approved):

PR #225, opened 2022-11-09
[NHUB-195] New privilege to restrict coverage details

MarkLark86:
Depends on https://github.com/superdesk/superdesk-core/pull/2394

### DIFF
--- a/newsroom/agenda/utils.py
+++ b/newsroom/agenda/utils.py
@@ -1,4 +1,4 @@
-from typing import Dict, NamedTuple, Any, Literal
+from typing import Dict, NamedTuple, Any, Literal, Tuple, Set
 from datetime import timedelta
 
 from flask import current_app as app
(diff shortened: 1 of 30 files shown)


newsroom/agenda/utils.py
```diff
@@ -185,3 +187,97 @@ def get_item_type(item: Dict[str, Any]) -> Literal["event", "planning"]:
...
+        )
+        push_notification(
+            name,
+            filters={"include": {"company": list(restricted_companies)}},
+            _items=remove_restricted_coverage_info(items),
+            **kwargs,
+        )
+
+
+def push_agenda_item_notification(name, item, **kwargs):
```
petrjasek, line 265: not sure if we need both methods, in theory this could be handled by calling:
```
push_agenda_items_notification("name", [item])
```
which would avoid some duplication

petrjasek (approved):

petrjasek (approved):

PR #230, opened 2022-11-15
[NHUB-193] Improve location dropdown style

MarkLark86:


petrjasek:
looks good to me, but will rather leave the review to @darconny

darconny (approved):
Let's merge it so we can test it with real data.

PR #231, opened 2022-11-15
it was converting the timestamps to local timezone

petrjasek:
which could make it go up/down one day, so now avoid that using moment utc mode which will display it as it is stored.

CPNHUB-161


MarkLark86 (approved):
There is also this bug CPNHUB-162: https://github.com/superdesk/newsroom-core/blob/develop/assets/utils.js#L476
`starTime` instead of `startTime`

petrjasek:
@MarkLark86 fixing it here as well

PR #247, opened 2022-12-12
fix filtering based on fields with nested config

petrjasek:
use WIRE_GROUPS config to figure out if field needs a nested filter.

STTNHUB-195

### DIFF
--- a/newsroom/search.py
+++ b/newsroom/search.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Union, Dict, Any
+from typing import List, Optional, Union, Dict, Any
 from copy import deepcopy
 
 from flask import current_app as app, json, abort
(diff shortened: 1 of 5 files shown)


newsroom/search.py
```diff
@@ -259,7 +259,43 @@ def get_internal_request(self, search):
...
+                                ],
+                            },
+                        },
+                    },
+                }
+            else:
+                query = {"terms": {self.get_aggregation_field(key): val}}
+            bool_query["must"].append(query)
+
+    def get_nested_config(self, field):
```
MarkLark86, line 290: There is similar functionality here: https://github.com/superdesk/newsroom-core/blob/release/2.3/newsroom/search_config.py#L27

Might be good to have this functionality in the one place, common to all resource types

newsroom/search.py
```diff
@@ -259,7 +259,43 @@ def get_internal_request(self, search):
         return internal_req
 
     def set_bool_query_from_filters(self, bool_query: Dict[str, Any], filters: Dict[str, Any]):
-        bool_query["must"] += [{"terms": {self.get_aggregation_field(key): val}} for key, val in filters.items() if val]
+        for key, val in filters.items():
+            if not val:
+                continue
+            nested_config = self.get_nested_config(key)
+            if nested_config:
```
MarkLark86, line 266: This will also need to be added for Agenda searches, https://github.com/superdesk/newsroom-core/blob/release/2.3/newsroom/agenda/agenda.py#L580

MarkLark86 (approved):

PR #258, opened 2022-12-22
add user products field

petrjasek:
which allows products to be assigned to a user.
if so it will use those instead of company products for search filtering.

CPCN-35

### DIFF
--- a/newsroom/search.py
+++ b/newsroom/search.py
@@ -18,6 +18,7 @@ from newsroom.products.products import (
     get_products_by_navigation,
     get_products_by_company,
     get_product_by_id,
+    get_products_by_user,
 )
 from newsroom.auth import get_user
 from newsroom.companies import get_user_company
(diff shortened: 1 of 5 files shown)


newsroom/auth/utils.py
```diff
@@ -120,17 +126,49 @@ def is_valid_session():
     )
 
 
+def get_user_company(user) -> Optional[Dict]:
```
MarkLark86, line 129: This is already defined in `newsroom.companies.utils`
https://github.com/superdesk/newsroom-core/blob/develop/newsroom/companies/utils.py#L4

newsroom/search.py
```diff
@@ -449,11 +450,21 @@ def prefill_search_products(self, search):
...
-                    product_type=search.section,
-                )
+                user_products = []
+                if search.user and search.user.get("products"):
+                    user_products = get_products_by_user(
+                        search.user,
+                        search.section,
+                    )
+
+                if user_products:
```
MarkLark86, line 460: If a `User` has `Products` defined, and this is part of the functional spec, we have to keep in mind this when exposing it to the UI (and place it in the Newshub docs somewhere).

Otherwise I fear it may become confusing when a `Company` has a `Product` but `End User` cannot see the items because the `User` has a `Product` override in place

MarkLark86 (approved):

newsroom/search.py
petrjasek: yep right we will have to extend the current user management, in 2.4 this will be visible via company management which will be the only place where this could be actually set

newsroom/auth/utils.py
petrjasek: I've added this due to a circular import it seems, will be doing some more refactoring in a following PR so will handle it there

PR #268, opened 2023-01-16
refactor company - products relation

petrjasek:
store product refs on companies so we can put there additional metadata like seats.

CPCN-75

### DIFF
--- a/data_updates/00009_20230116-145407_products.py
+++ b/data_updates/00009_20230116-145407_products.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8; -*-
+# This file is part of Superdesk.
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+#
+# Author  : petr
+# Creation: 2023-01-16 14:54
+
+from superdesk.commands.data_updates import DataUpdate as _DataUpdate
+
+
+class DataUpdate(_DataUpdate):
+
+    resource = 'products'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        db = mongodb_database
+        products = list(db.products.find({}))
+        for company in db.companies.find({}):
+            company_products = [
+                {"_id": p["_id"], "section": p.get("product_type") or "wire"}
+                for p in products
+                if p.get("companies") and str(company["_id"]) in set(map(str, p["companies"]))
+            ]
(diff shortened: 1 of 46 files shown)


MarkLark86:
@petrjasek Looks good

data_updates/00009_20230116-145407_products.py
```diff
@@ -0,0 +1,29 @@
...
+        for company in db.companies.find({}):
+            company_products = [
+                {"_id": p["_id"], "section": p.get("product_type") or "wire"}
+                for p in products
+                if p.get("companies") and str(company["_id"]) in set(map(str, p["companies"]))
+            ]
+            db.companies.update_one({"_id": company["_id"]}, {"$set": {"products": company_products}})
+
+    def backwards(self, mongodb_collection, mongodb_database):
+        raise NotImplementedError()
```
MarkLark86, line 29: This should be `pass`, so **if** we need to roll back updates this will pass not fail

newsroom/companies/companies.py
```diff
@@ -46,7 +46,19 @@ class CompaniesResource(newsroom.Resource):
         "allowed_ip_list": {"type": "list", "mapping": {"type": "string"}},
         "original_creator": newsroom.Resource.rel("users"),
         "version_creator": newsroom.Resource.rel("users"),
+        "products": {
+            "type": "list",
+            "schema": {
+                "type": "dict",
+                "schema": {
+                    "_id": newsroom.Resource.rel("products"),
+                    "seats": {"type": "number"},
```
MarkLark86, line 55: Should this have a default of `0`, so this attribute is optional but always has a value?

data_updates/00009_20230116-145407_products.py
```diff
@@ -0,0 +1,29 @@
...
+from superdesk.commands.data_updates import DataUpdate as _DataUpdate
+
+
+class DataUpdate(_DataUpdate):
+
+    resource = "products"
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        db = mongodb_database
+        products = list(db.products.find({}))
```
MarkLark86, line 19: Should we also remove `companies` from the `products` resource?

MarkLark86 (approved):

MarkLark86 (approved):

data_updates/00009_20230116-145407_products.py
petrjasek: I kept it there instead of doing a rollback

PR #269, opened 2023-01-17
improvement: Upgrade Bootstrap to v5 and support RTL

MarkLark86:


webpack.config.js
```diff
@@ -42,7 +42,11 @@ const config = {
         rules: [
             {
                 test: /\.jsx?$/,
-                include: path.resolve(__dirname, 'assets'),
+                include: [
+                    path.resolve(__dirname, 'assets'),
+                    path.resolve(__dirname, 'node_modules/bootstrap'),
+                    path.resolve(process.cwd(), 'node_modules/bootstrap'),
```
petrjasek, line 48: I wonder why we would need this?

petrjasek (approved):

webpack.config.js
MarkLark86: Because Bootstrap only supports es6 and above. Even their compiled min versions for production includes spread operator etc.

So we need to transpile it to es2015 like we're already doing.

An improvement might be to increase our version of ecma that we support in browsers (using babel preset env package)

petrjasek (approved):

PR #276, opened 2023-01-31
split fetching items and aggregations in wire

petrjasek:
NHUB-248


package.json
```diff
@@ -36,9 +36,11 @@
     "moment-timezone": "0.5.23",
     "node-sass": "^4.5.3",
     "popper.js": "1.14.4",
-    "react": "16.6.0",
+    "primeicons": "^6.0.1",
+    "primereact": "^8.7.0",
+    "react": "^17.0.2",
```
MarkLark86, line 41: I wonder if it's a good idea to upgrade to a new major version of React for a bugfix & improvement release of 2.3.
Was this because of bringing in `primereact`?

petrjasek: yep exactly, I didn't see any breaking changes in the release notes nor any issues with the UI so I did it, but might need some extra testing. if we find any issue I would revert that part and use standard loader

MarkLark86: Maybe we should rename the deprecated React lifecycle methods: https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles.

petrjasek: should be ok still https://reactjs.org/blog/2020/08/10/react-v17-rc.html#other-breaking-changes , we might need to do more changes when going for v18

MarkLark86 (approved):

PR #281, opened 2023-02-02
Fix : Newly created user can’t set their password [NHUB-238]

devketanpro:


petrjasek:
this would let non admin user do anything to other users, I suppose that's not the idea

devketanpro:
> this would let non admin user do anything to other users, I suppose that's not the idea

@petrjasek I apologise. I didn't understand your point; could you perhaps explain it to me in more detail, or what I'm missing based on your approach?

newsroom/users/users.py
```diff
@@ -187,6 +187,8 @@ def check_permissions(self, doc, updates=None):
         """Check if current user has permissions to edit user."""
         if not request or request.method == "GET":  # in behave there is test request context
             return
+        if not is_current_user_admin() and request.method == "POST":
```
petrjasek, line 190: if the user is not an admin it should check if the user being edited is the user who is editing, so prevent standard user from changing other users data

devketanpro: > it should check if the user being edited is the user who is editing

As user is not logged in so I guess I can check and verify token which is being sent in mail and using the token I can get the user and verify the details as well. 

please provide your suggestions on it. :)

thanks

petrjasek: I see what you mean now, it should check then for specific view and not user type
via [request.url_rule](https://flask.palletsprojects.com/en/2.2.x/api/#flask.Request.url_rule) so it would be something like:
```py
if request and request.url_rule and request.url_rule.rule == '/reset_password/<token>':
    return
```

devketanpro: Thanks !!

petrjasek (approved):

PR #284, opened 2023-02-03
[CPCN-86] Add datetime param to Product Seat Request email

MarkLark86:


newsroom/company_admin/views.py
```diff
@@ -75,11 +76,14 @@ def send_product_seat_request_email():
         to=recipients,
         template="additional_product_seat_request_email",
         template_kwargs=dict(
+            app_name=app.config["SITE_NAME"],
```
petrjasek, line 79: you can use `config.SITE_NAME` that's available by default in flask

petrjasek (approved):

newsroom/company_admin/views.py
MarkLark86: Currently in this repo we're using `app.config["SITE_NAME"]` whenever we want the name, where as `config.SITE_NAME` is not used anywhere. I have used this way to get the name to be consistent with other uses.

petrjasek: ok np

PR #288, opened 2023-02-09
delete refs to product when product is deleted

petrjasek:
NHUB-255

### DIFF
--- a/newsroom/products/products.py
+++ b/newsroom/products/products.py
@@ -40,7 +40,13 @@ class ProductsResource(newsroom.Resource):
 
 
 class ProductsService(newsroom.Service):
-    pass
+    def on_deleted(self, doc):
+        lookup = {"products._id": doc["_id"]}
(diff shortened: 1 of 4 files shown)


newsroom/products/products.py
```diff
@@ -40,7 +40,13 @@ class ProductsResource(newsroom.Resource):
 
 
 class ProductsService(newsroom.Service):
-    pass
+    def on_deleted(self, doc):
+        lookup = {"products._id": doc["_id"]}
+        for resource in ("users", "companies"):
+            items = superdesk.get_resource_service(resource).get(req=None, lookup=lookup)
```
MarkLark86, line 46: Will this be affected by pagination? I don't see an issue with `companies` resource, but maybe the `users` resource could have more than the default page size.

petrjasek: there should be no limit if not set on the request

MarkLark86 (approved):

PR #298, opened 2023-02-22
[NHUB-259] fix(bs5): Dropdown menus not hiding if filter is active

MarkLark86:
The solution here is a hack, as I was unable to get the dropdown button & menu to behave properly when applying the `active` class using React.

It seems that whenever the `className` of the `DropdownFilterButton` is changed via React, Bootstrap loses control of the menu (show/hide etc). I tested this by adding a `className` property and passing in the `active` class from a higher component, which caused the menu to not behave properly when the class changed

### DIFF
--- a/assets/components/DropdownFilterButton.jsx
+++ b/assets/components/DropdownFilterButton.jsx
@@ -2,41 +2,64 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-function DropdownFilterButton({id, isActive, autoToggle, onClick, icon, label, textOnly, iconColour}) {
-    return (
-        <button
-            id={id}
-            type="button"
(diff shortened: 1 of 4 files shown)


assets/components/DropdownFilterButton.jsx
```diff
@@ -2,41 +2,64 @@ import React from 'react';
...
+        return (
+            <button
+                id={id}
+                key={id}
+                type="button"
+                ref={(elem) => this.dom.button = elem}
+                className={classNames(
+                    'btn btn-sm d-flex align-items-center px-2 ms-2',
+                    {
+                        'btn-text-only': textOnly,
```
MarkLark86, line 36: As another test, I even tried `'foobar': isActive` here, and the menu still does not hide on click.

PR #303, opened 2023-02-23
fix filter dropdown

petrjasek:
this closes it on re-render which happens when something is selected, had to inline the button otherwise it wouldn't get the ref which it needs to initialize the menu properly (the menu must be next element).

NHUB-295

### DIFF
--- a/assets/agenda/components/AgendaDropdown.jsx
+++ b/assets/agenda/components/AgendaDropdown.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {gettext} from 'utils';
+import {Dropdown} from '../../components/Dropdown';
+
+export function AgendaDropdown({filter, activeFilter, toggleFilter, children}) {
+
+    const isActive = activeFilter[filter.field];
+
+    const getActiveFilterLabel = (filter, activeFilter, isActive) => {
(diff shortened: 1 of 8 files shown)


MarkLark86:
@petrjasek This is also happening on the `AgendaFilterButton` and `MultiSelectDropdown` components.

Might be good to implement a generic dropdown container, which provides the button and maybe children as the menu itself, that way the fix is implemented in the one place, and all dropdowns use this new container/component

petrjasek:
it still didn't work somehow for calendar dropdown, at the end I gave up on bootstrap events handling and implemented the clicking manually

assets/agenda/components/AgendaDropdown.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {gettext} from 'utils';
+import {Dropdown} from '../../components/Dropdown';
+
+export function AgendaDropdown({filter, activeFilter, toggleFilter, children}) {
+
+    const isActive = activeFilter[filter.field];
+
+    const getActiveFilterLabel = (filter, activeFilter, isActive) => {
```
MarkLark86, line 10: There probably isn't a need for this to be a function, as we're calling it straight away and using the value.
This could be replaced with a variable assignment instead.

MarkLark86 (approved):

PR #304, opened 2023-02-28
 Fix : Gallery publishing issues [NHUB-30]

devketanpro:


petrjasek (changes requested):
there are other usages of `getPicture` where it's expecting to get single picture and we shouldn't change it to return 1 or a list, but you can introduce a new helper which would get all article images and use it in the preview/detail instead

assets/wire/components/ItemDetails.jsx
```diff
@@ -48,7 +48,7 @@ function ItemDetails({
     listConfig,
     filterGroupLabels,
 }) {
-    const picture = getPicture(item);
+    const picture = getPictureList(item);
```
petrjasek, line 51: this should be rather called pictures now, and not sure if that `isCustomRendition` would work with a list?

PR #305, opened 2023-03-01
[NHUB-263] fix: Allow public users to edit their profile

MarkLark86:


newsroom/users/users.py
```diff
@@ -191,11 +198,23 @@ def check_permissions(self, doc, updates=None):
...
         if is_current_user_company_admin():
             manager = get_user()
             if doc.get("company") and doc["company"] == manager.get("company"):
-                if not updates or all([key in COMPANY_ADMIN_ALLOWED_UPDATES for key in updates.keys()]):
+                if not updated_fields or all([key in COMPANY_ADMIN_ALLOWED_UPDATES for key in updated_fields]):
+                    return
+        if request and request.url_rule and request.url_rule.rule:
+            if request.url_rule.rule in ["/reset_password/<token>", "/token/<token_type>"]:
+                return
+            elif request.url_rule.rule == "/users/<_id>":
```
petrjasek, line 217: just wondering - would this also affect the mgmt api?

petrjasek (approved):

newsroom/users/users.py
MarkLark86: @petrjasek This shouldn't be an issue, as we're overriding the function in `cp.mgmt_api.user.CPUsersService.check_permission`
```python
class CPUsersService(UsersService):
    ....
    def check_permissions(self, doc, updates=None):
        """Avoid testing if user has permissions."""
        pass
```

PR #309, opened 2023-03-07
Fix:Search in company admin settings is not working [CPCN-95]

devketanpro:


MarkLark86:
@devketanpro I'm not sure how this fixes searching of the Companies.
https://github.com/superdesk/newsroom-core/blob/develop/assets/company-admin/components/CompanyAdminApp.jsx#L98

Indicates that if the `sectionId` is not users then don't update any query.
For these Company products though, might need to implement the searching of these (so the list of Company Products are searchable/filterable)

assets/company-admin/components/CompanyAdminApp.jsx
```diff
@@ -95,10 +95,10 @@ class CompanyAdminAppComponent extends React.Component {
                     <div className="content">
                         <ListBar
                             onNewItem={this.props.sectionId !== 'users' ? undefined : this.props.newUser}
-                            setQuery={this.props.sectionId === 'users' ? this.props.setQuery : undefined}
-                            fetch={this.props.sectionId === 'users' ? this.props.fetchUsers : this.setProductFilterQuery}
+                            setQuery={this.props.setQuery}
+                            fetch={this.props.sectionId === 'users' ?  this.props.fetchUsers : this.setProductFilterQuery }
                             buttonText={gettext('New User')}
-                            enableQueryAction={this.props.sectionId === 'users'}
+                            enableQueryAction={this.props.sectionId === 'users' || this.props.sectionId === 'my_company'}
```
petrjasek, line 101: not sure, but shouldn't be just `true` now? or is there any case when it wouldn't be enabled

assets/company-admin/components/CompanyDetails.jsx
```diff
@@ -30,14 +29,21 @@ function CompanyDetailsComponent({company, showSeatRequestModal, setSection, com
...
-                                    onNameClicked={() => setSection('users', seat._id)}
-                                />
-                            ))}
+
+                            {
+                                (query === null ?
+                                    Object.values(seats[company._id]).filter((seat) => seat.section === section._id) :
+                                    Object.values(seats[company._id]).filter((seat) => seat.section === section._id && seat.name.toString().includes(query))
+                                )
+                                    .map((seat) => (
```
petrjasek, line 38: I'm a bit confused with the wording here, shouldn't this be rather `product` instead of `seat`?

assets/company-admin/components/CompanyAdminApp.jsx
devketanpro: IMO there is not any case where it would not be enabled

assets/company-admin/components/CompanyDetails.jsx
devketanpro: sounds meaningful and gives context

assets/company-admin/components/CompanyAdminApp.jsx
devketanpro: As a result, we could simply set it to true.

assets/company-admin/components/CompanyDetails.jsx
```diff
@@ -30,14 +29,21 @@ function CompanyDetailsComponent({company, showSeatRequestModal, setSection, com
...
-                                    seat={seat}
-                                    showSeatRequestModal={showSeatRequestModal}
-                                    onNameClicked={() => setSection('users', seat._id)}
-                                />
-                            ))}
+
+                            {
+                                (query === null ?
+                                    Object.values(products[company._id]).filter((product) => product.section === section._id) : 
+                                    Object.values(products[company._id]).filter((product) => product.section === section._id && product.name.toString().includes(query))
```
MarkLark86, line 36: It would be good to convert the query and product.name to lowercase when comparing these two

assets/company-admin/components/CompanyAdminApp.jsx
petrjasek: yep

MarkLark86 (approved):

PR #310, opened 2023-03-07
Fix : Users list does not display in alphabetical order by First Name [CPCN-99]

devketanpro:


assets/users/components/filters/UserListSortFilter.jsx
```diff
@@ -38,6 +41,10 @@ export class UserListSortFilter extends React.PureComponent {
         this.getFilterLabel = this.getFilterLabel.bind(this);
     }
 
+    componentDidMount() {
+        this.props.fetchUsers();
```
petrjasek, line 45: just wondering why we need it now and didn't need this before?

devketanpro: we must fetch an updated list when we click on the users component, which I believe is a reasonable technique.

petrjasek: the users should be already there, maybe it just needs a change here:
https://github.com/superdesk/newsroom-core/blob/develop/newsroom/users/users.py#L93

petrjasek (approved):

PR #316, opened 2023-03-15
Fix : No user #count on the User tab [CPCN-102]

devketanpro:


assets/company-admin/components/CompanyAdminApp.jsx
```diff
@@ -134,6 +134,9 @@ class CompanyAdminAppComponent extends React.Component {
                                     {gettext('Users')}
                                 </button>
                             </div>
+
+                            {this.props.sectionId === 'users' ? <label className='label label--big label--rounded'>USERS : {
```
MarkLark86, line 138: Use `gettext` to flag this label for translation (also include `totalusers` in the translation)

MarkLark86 (approved):

PR #317, opened 2023-03-16
FIX : No admin icon on company admin accounts on user tab [CPCN-101]

devketanpro:


MarkLark86:
@devketanpro This should be closed, as the functionality should be working already (see the comments in the associated ticket)

PR #326, opened 2023-03-27
update fi translations

MarkLark86:


newsroom/translations/fi/LC_MESSAGES/messages.po
```diff
@@ -1461,108 +1475,39 @@ msgstr "Yhdistetty ilmoituspalvelimeen!"
 msgid "Disconnected from Notification Server!"
 msgstr "Yhteys ilmoituspalvelimeen katkennut!"
 
-#: assets/agenda/actions.js:116
+#: assets/agenda/actions.js:112
 msgid "Your inquiry has been sent successfully"
 msgstr "Hakusi on lähetetty"
 
-#: assets/agenda/actions.js:189
```
petrjasek, line 1468: I think we should keep those following items in, they seem to be still there in 2.3, probably some refactoring

MarkLark86: These ones were only used with the copying of metadata to the clipboard,
which has been moved to the server side via templates. I double checked these already

petrjasek: I see, all good then

petrjasek (approved):

PR #330, opened 2023-03-28
Fix : When an item is corrected the refresh button does not appear [NHUB-277]

devketanpro:


assets/reducers.js
```diff
@@ -341,7 +341,7 @@ export function defaultReducer(state={}, action) {
             ...state,
             newItems: uniq([
                 ...state.newItems,
-                ...(action.data._items.filter((item) => !item.nextversion && !state.itemsById[item._id]).map((item) => item._id))
+                ...(action.data._items.filter((item) => (!item.nextversion && !state.itemsById[item._id]) || item.ednote).map((item) => item._id))
```
MarkLark86, line 344: It might be better to compare the item's `state` to the one already in the store, rather than the existence of `ednote`. i.e.

```typescript
const newItems = action.data._items
    .filter((item) => (!item.nextversion && get(state.itemsById, `${item._id}.state`) !== item.state))
    .map((item) => item._id)
```

devketanpro: I attempted this approach, but we don't get 'item['state']' because it's missing from the item.

devketanpro: so for the update item case in this condition `!item.nextversion && get(state.itemsById, `${item._id}.state`) !== item.state}`
all are undefined

MarkLark86: Sorry my bad, I didn't realise we don't have `state` in Newshub Wire items. Try comparing `versioncreated` instead then

MarkLark86:
@devketanpro If you pull the latest from `develop` branch, the instance should build

MarkLark86 (approved):

PR #331, opened 2023-03-29
[CPCN-110] fix: Create separate endpoint to update User Profile

MarkLark86:
@petrjasek I created a new endpoint specifically for updating of User Profiles, as we've had numerous issues around creating/updating users between the profile page and settings page (section & product privileges being lost, profile data not being saved etc).

Having a dedicated endpoint that only allows `USER_PROFILE_UPDATES` reduces the risks of breaking the ability to save a User Profile

### DIFF
--- a/newsroom/users/views.py
+++ b/newsroom/users/views.py
@@ -235,7 +235,7 @@ def get_updates_from_form(form: UserForm):
 @login_required
 def edit_user_profile(_id):
     if not is_current_user(_id):
-        flask.abort(401)
+        flask.abort(403)
 
     user_id = ObjectId(_id)
     user = find_one("users", _id=user_id)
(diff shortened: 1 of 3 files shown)


newsroom/users/views.py
```diff
@@ -231,6 +231,26 @@ def get_updates_from_form(form: UserForm):
     return updates
 
 
+@blueprint.route("/users/<_id>/profile", methods=["POST"])
+@login_required
+def edit_user_profile(_id):
+    if not is_current_user(_id):
+        flask.abort(401)
```
petrjasek, line 238: should be rather 403 imo

petrjasek (approved):

newsroom/users/views.py
MarkLark86: Whoops, I'll fix that one tomorrow

PR #341, opened 2023-04-17
[NHUB-281] Static app for UI design purposes

MarkLark86:


MarkLark86:
@darconny I have added the files that you provided in the ticket. I removed the style and assets files, so that you can work directly with the style files from the actual Newshub app.

To start the app all you have to do is:
```bash
npm install
npm start design-app
```

The above is outlined in the updated README.md file

petrjasek (approved):

PR #348, opened 2023-04-20
Fix : display the correct user type in the user type field in the Company Management view [CPCN-136]

devketanpro:


assets/users/utils.js
```diff
@@ -2,10 +2,10 @@ import {gettext} from '../utils';
...
 
 export const userTypes = [
-    {value: 'administrator', text: gettext('Administrator'), show_acc_mgr: false},
-    {value: 'internal', text: gettext('Internal'), show_acc_mgr: false},
+    {value: 'administrator', text: gettext('Administrator'), show_acc_mgr: true},
+    {value: 'internal', text: gettext('Internal'), show_acc_mgr: true},
     {value: 'public', text: gettext('Public'), show_acc_mgr: true},
-    {value: 'account_management', text: gettext('Account Management'), show_acc_mgr: false},
+    {value: 'account_management', text: gettext('Account Management'), show_acc_mgr: true},
     {value: 'company_admin', text: gettext('Company Admin'), show_acc_mgr: true},
```
petrjasek, line 9: I wonder if we need that `show_acc_mgr` value now, seems like it's always on

petrjasek (approved):

assets/users/utils.js
devketanpro: actually previously we off that value according to the ticket CPCN-130.

petrjasek (approved):

PR #351, opened 2023-04-27
fix user sections stored as list during migration

petrjasek:
CPCN-137

### DIFF
--- a/data_updates/00010_20230427-151329_users.py
+++ b/data_updates/00010_20230427-151329_users.py
@@ -22,25 +22,30 @@ class DataUpdate(_DataUpdate):
 
         for user in mongodb_collection.find({}):
             company = companies.get(user.get("company"))
-            if not company:
-                continue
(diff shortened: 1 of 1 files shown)


data_updates/00010_20230427-151329_users.py
```diff
@@ -22,25 +22,30 @@ def forwards(self, mongodb_collection, _mongodb_database):
...
+
+            if isinstance(user.get("sections"), list):
+                updates["sections"] = {name: True for name in user["sections"]}
+
+            if company:
+                company_section_names = get_company_section_names(company)
+                company_product_ids = get_company_product_ids(company)
+                user_sections = updates.get("sections") or user.get("sections") or {}
+
+                updates = {
```
MarkLark86, line 35: This will override the `sections` attribute changes from above, will need to use `update` instead.

MarkLark86 (approved):

PR #353, opened 2023-05-02
[CPCN-138] fix(api): Include False values when generating flask.g.settings

MarkLark86:
A bug with boolean settings, such as `allow_companies_to_manage_products`, where a valid `False` value was not being stored in `flask.g.settings`, resulting in the default value being used instead.

### DIFF
--- a/newsroom/settings.py
+++ b/newsroom/settings.py
@@ -59,7 +59,7 @@ def get_setting(setting_key=None, include_audit=False):
         settings = copy.deepcopy(flask.current_app._general_settings)
         if values:
             for key, val in values.get("values", {}).items():
-                if val is not None and settings.get(key):
+                if (val or val is False) and settings.get(key):
                     settings[key]["value"] = val
             if include_audit:
                 settings["_updated"] = values.get("_updated")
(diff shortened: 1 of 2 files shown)


devketanpro (approved):

newsroom/settings.py
```diff
@@ -59,7 +59,7 @@ def get_setting(setting_key=None, include_audit=False):
         settings = copy.deepcopy(flask.current_app._general_settings)
         if values:
             for key, val in values.get("values", {}).items():
-                if val and settings.get(key):
+                if (val or val is False) and settings.get(key):
```
petrjasek, line 62: wondering if this shouldn't be rather:
```
if val is not None and settings.get(key) is not None:
```
otherwise we won't use the value when there is `False` in settings

MarkLark86: I don't believe so, as we have a test for this when the value is an empty string, then the default value is used.

So I'd say that the default should be used if the value is either None or an empty string, otherwise use the supplied value

MarkLark86: I have updated the PR so it uses `not (val is None or val == "")`, so that we're explicitly stating these two values will result in a default value being applied.

petrjasek: I see, my bad I thought we would have the value in that `settings` dict but it's actually the config

PR #354, opened 2023-05-02
Fix : Search results for label on top of the list is empty [NHUB-214]

devketanpro:


assets/search/components/ContentSearchResults.jsx
```diff
@@ -125,7 +125,7 @@ class ContentSearchResultsComponent extends React.Component {
                             {!this.props.showTotalLabel ? null : (
                                 <React.Fragment>
                                     <span className="search-results-info__text flex-column">
-                                        <span>{gettext('search results for:')}</span>
+                                        <span>{gettext('search results for:')}{this.props.searchParams.query}</span>
```
petrjasek, line 128: might be good to have a space there

devketanpro: my bad I forgot to add.

assets/search/components/ContentSearchResults.jsx
```diff
@@ -125,7 +125,7 @@ class ContentSearchResultsComponent extends React.Component {
                             {!this.props.showTotalLabel ? null : (
                                 <React.Fragment>
                                     <span className="search-results-info__text flex-column">
-                                        <span>{gettext('search results for:')}</span>
+                                        <span>{gettext('search results for: ')}{this.props.searchParams.query}</span>
```
petrjasek, line 128: might be better to have it like:
```gettext('search results for: {{query}}', {query: this.propssearchParams.query})```
or to keep that space outside gettext so it won't be missed when translating

File: assets/search/components/ContentSearchResults.jsx [hunk 3]
```diff
@@ -125,7 +125,7 @@ class ContentSearchResultsComponent extends React.Component {
                             {!this.props.showTotalLabel ? null : (
                                 <React.Fragment>
                                     <span className="search-results-info__text flex-column">
-                                        <span>{gettext('search results for: ')}{this.props.searchParams.query}</span>
+                                        <span>{gettext('search results for: ') + this.props.searchParams.query}</span>
```
petrjasek, line 128: this won't have the space when it's removed during translation, would need something like:
```
gettext('search results for:') + ' ' + this.props.searchParams.query
```
to be sure the space is there

PR #356, opened 2023-05-03
Fix : Agenda section shows blank if user language is French [CPNHUB-231]

devketanpro:


assets/components/CalendarButton.jsx
```diff
@@ -28,8 +26,12 @@ class CalendarButton extends React.Component {
     render() {
         const isStartDateToday = moment.isMoment(this.state.startDate) && !this.state.startDate.isSame(moment(), 'day');
         const datePicker = (<DatePicker
-            customInput={<CalendarButtonWrapper active={isStartDateToday}/>}
-            dateFormat={AGENDA_DATE_PICKER_FORMAT_SHORT}
+            customInput={
+                <CalendarButtonWrapper
+                    active={isStartDateToday}
+                    date={moment(this.state.startDate).format('YYYY-MM-DD')}
```
petrjasek, line 32: I don't see date prop in `CalendarButtonWrapper` so not sure what this does actually?

devketanpro: my bad its typo. here we have to pass `value` prop

petrjasek:
what about that default config?

devketanpro:
with these changes 

for fr :- 
image
and for en :-
image

PR #372, opened 2023-05-17
[CPCN-67] Feature: Add advanced search to Wire

MarkLark86:


features/web_api/wire_advanced_search.feature
```diff
@@ -0,0 +1,116 @@
...
+        ["sports-results-today-1"]
+        """
+        When we get "/wire/search?advanced_search={"exclude":["Weather"],"fields":["headline"]}"
+        Then we get the following order
+        """
+        ["sports-results-today-1"]
+        """
+
+    @auth @admin @skip
+    Scenario: Search multiple fields
```
MarkLark86, line 108: Using ES validate & explain, I got the following:

Slugline & Headline:
```
+(
    +blended(terms:[slugline:sydney, headline:sydney])
    +blended(terms:[slugline:current, headline:current])
    +blended(terms:[slugline:report, headline:report])
)
```

Headline & Body HTML
```
+(
    (+body_html:sydney +body_html:current +body_html:report) |
    (+headline:sydney +headline:current +headline:report)
)
```

As you can see, the `headline & body_html` does not work as intended

petrjasek: ok so we need to tweak the mapping right? it would be useful to control it in newsroom-core repo anyhow, so we might register an items resource for this explicitly here rather than doing the changes only in sd-core

newsroom/search.py
```diff
@@ -639,6 +654,43 @@ def apply_request_filter(self, search):
...
+        if not search.args.get("advanced_search"):
+            return
+
+        search.advanced = json.loads(search.args["advanced_search"])
+        fields = search.advanced.get("fields") or self.default_advanced_search_fields
+        if not fields:
+            return
+
+        def gen_match_query(keywords: str, operator: str, multi_match_type):
+            if len(fields) == 1:
```
petrjasek, line 667: just wondering if that multi match wouldn't do the same thing, is this part needed?

newsroom/search.py
```diff
@@ -66,6 +66,13 @@ def get_filter_query(
     return {"terms": {aggregation_field: val}}
 
 
+class AdvancedSearchParams(TypedDict):
+    all: str
+    any: str
+    exclude: str
+    fields: str
```
petrjasek, line 73: imo this might be still rather `List[str]`

petrjasek (approved):

newsroom/search.py
MarkLark86: You're right, under the hood it probably creates the same query. I'll upload a fix shortly

newsroom/search.py
MarkLark86: Whoops

PR #378, opened 2023-05-25
Fix: Test product in agenda section of products shows a blank page with error [CPCN-158]

devketanpro:


assets/products/components/EditProduct.jsx
```diff
@@ -53,14 +53,14 @@ class EditProduct extends React.Component {
         if (product.product_type === 'agenda') {
             q = JSON.stringify({
                 query: product.query,
-                planning_item_query: product.planning_item_query
+                planning_item_query: product.planning_item_query,
+                sd_product_id: product.sd_product_id,
```
MarkLark86, line 57: I don't believe we have code on the backend to cater for this `q={"sd_product_id":"<product_id>"}` param. See: https://github.com/superdesk/newsroom-core/blob/develop/newsroom/agenda/agenda.py#L933

What might work instead is to append the ID to the `query` param, just like we are with the Wire test query.

i.e.
```typescript
q = JSON.stringify({
    query: getProductQuery(product),
    planning_item_query: product.planning_item_query,
});
```

petrjasek (approved):

PR #379, opened 2023-05-25
improve saml config options

petrjasek:
we can set on company email domain which will be tested when authenticating via saml and if user email matches it it will assign the company to user.

there is also new config for custom client saml configs which allows us to configure user specific login page which picks the config to be used for SAML auth. then it uses the company domain to match the config value.

STTNHUB-222

### DIFF
--- a/assets/companies/components/EditCompanyDetails.jsx
+++ b/assets/companies/components/EditCompanyDetails.jsx
@@ -81,6 +81,14 @@ export function EditCompanyDetails({company, companyTypes, users, errors, onChan
                     error={errors ? errors.contact_email : null}
                 />
 
+                <TextInput
+                    name='auth_domain'
+                    label={gettext('SSO domain')}
+                    value={company.auth_domain || ''}
+                    onChange={onChange}
+                    error={errors ? errors.auth_domain : null}
+                />
+
                 <SelectInput
                     name='country'
                     label={gettext('Country')}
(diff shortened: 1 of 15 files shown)


assets/companies/components/EditCompanyDetails.jsx
```diff
@@ -81,6 +81,14 @@ export function EditCompanyDetails({company, companyTypes, users, errors, onChan
                     error={errors ? errors.contact_email : null}
                 />
 
+                <TextInput
```
MarkLark86, line 84: Should we hide this input if SSO is not enabled, as there would be no need to show an input that has no affect.

newsroom/auth/views.py
```diff
@@ -255,3 +255,11 @@ def set_locale():
     if locale and locale in app.config["LANGUAGES"]:
         flask.session["locale"] = locale
     return flask.redirect(flask.url_for("auth.login"))
+
+
+@blueprint.route("/login/<client>", methods=["GET"])
```
MarkLark86, line 260: Would this be better off in `newsroom.auth.saml` file instead, so we're not adding an endpoint that has no use if SSO is not enabled?

newsroom/templates/login_locale_dropdown.html
```diff
@@ -0,0 +1,16 @@
+{% if get_client_locales()|length is gt 1 %}
+<div class="card-body pt-1">
+    <form class="form" role="form" method="post" action="{{ url_for('auth.set_locale') }}">
+        <div class="form-group">
+            <label for="language">{{ gettext('Language') }}</label>
```
MarkLark86, line 5: `id` attribute matching this `for` is missing on the `select` tag

MarkLark86 (approved):
Looks fine, only a few small suggestions

PR #380, opened 2023-05-26
Treat situation if there is no product assigned to a user [NHUB-285]

devketanpro:


assets/agenda/components/AgendaApp.jsx
```diff
@@ -86,7 +86,10 @@ class AgendaApp extends BaseApp {
 
     render() {
         if (this.state.initialLoad) {
-            return this.renderLoader();
+            return <div className="wire-articles__item-wrap col-12">
```
MarkLark86, line 89: The `renderLoader` is there to indicate to the user that the page/app is still loading. Therefor to display this message before loading has completed may confuse end users.

Instead you would need to check the user products, and if they're not admin, then show this alert

assets/agenda/components/AgendaApp.jsx
```diff
@@ -85,8 +85,15 @@ class AgendaApp extends BaseApp {
     }
 
     render() {
-        if (this.state.initialLoad) {
-            return this.renderLoader();
+        if (this.state.initialLoad) {    
+            if(this.props.userType !== 'administrator') {
```
petrjasek, line 89: I think it should also check if user has any products, otherwise this will be there for every non admin user no?
and this should be in theory only after that initialLoad block:
```ts
if (this.state.initialLoad) {
    return this.renderLoader();
}
if (this user is not admin && this user has no products) {
    return <div>{'You don't have any products...'})</div>;
}
...
```

devketanpro: that's not worked, I guess because we got `this.state.initialLoad = true`, so it returns only `renderLoader()`

petrjasek: you mean that it keeps that state `initialLoad` because of missing products? I think we should handle that then, there is probably an error returned from the server so we can handle that in the action and change the state accordingly

assets/agenda/actions.js
```diff
@@ -675,3 +683,22 @@ export function stopWatchingCoverage(coverage, item) {
...
     };
 }
+
+export function fetchUser(id) {
+    return function (dispatch) {
+        return server.get(`/users/${id}`)
+            .then((data) => {
+                dispatch(getUser(data));
+            })
+            .catch((error) => errorHandler(error, dispatch, setError));
```
petrjasek, line 693: what I had in mind was handling that 403 server response here and set some flag on state which would be then used in render to display the message, not adding an extra request for the profile

assets/agenda/actions.js
```diff
@@ -675,3 +686,16 @@ export function stopWatchingCoverage(coverage, item) {
             }, (error) => { errorHandler(error, dispatch);});
     };
 }
+
+export function errorHandler(error, dispatch, setError) {
+    console.error('error', error);
+    if (setError) {
+        if (error.response && error.response.status === 403) {
+            dispatch(setErrorMessage('No Products'));
```
petrjasek, line 694: imo it would make sense to set here the actual message using gettext and then only display `status.errorMessage` in the UI so we can handle different response codes if needed, storing there `No Products` which is not checked later on makes it rather confusing.

petrjasek (approved):

PR #381, opened 2023-05-26
Fix : remove the products info and numbers from the Users tab based on key allow_companies_to_manage_products [NHUB-289]

devketanpro:
https://sofab.atlassian.net/browse/NHUB-289?focusedCommentId=122408

### DIFF
--- a/assets/company-admin/components/CompanyUsers.jsx
+++ b/assets/company-admin/components/CompanyUsers.jsx
@@ -2,14 +2,14 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 
-import {gettext} from 'utils';
+import {gettext, getConfig} from 'utils';
 import {selectUser as _selectUser} from 'users/actions';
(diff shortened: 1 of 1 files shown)


assets/company-admin/components/CompanyUsers.jsx
```diff
@@ -2,14 +2,14 @@ import React from 'react';
...
-import {gettext} from 'utils';
+import {gettext, getConfig} from 'utils';
 import {selectUser as _selectUser} from 'users/actions';
 import {currentCompanySelector, companySectionListSelector, userIdSelector, userIdMapSelector} from '../selectors';
 
 import {CompanyUserListItem} from './CompanyUserListItem';
 
 function CompanyUsersComponent({users, usersById, selectUser, activeUserId, currentCompany, companySections}) {
-    const sections = companySections[currentCompany._id];
+    const sections = getConfig('allow_companies_to_manage_products') ? companySections[currentCompany._id] : [];
```
MarkLark86, line 12: Rather than executing `getConfig` twice, it might be good to store the result and use that instead

petrjasek (approved):

PR #384, opened 2023-05-30
handle company products with 0 seats as unlimited

petrjasek:
so every user in a company has access to those

NHUB-290

### DIFF
--- a/tests/core/test_products.py
+++ b/tests/core/test_products.py
@@ -16,6 +16,8 @@ def product(app):
         "name": "Sport",
         "description": "Top level sport product",
         "is_enabled": True,
+        "query": "sports",
+        "product_type": "wire",
     }
     app.data.insert("products", [_product])
     return _product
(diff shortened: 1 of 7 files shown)


tests/core/test_products.py
```diff
@@ -207,6 +209,35 @@ def test_delete_assigned_product(client, app, product, companies, user):
...
+    app.data.insert("products", [product2])
+    app.data.insert(
+        "items",
+        [
+            {"headline": "finance item", "type": "text"},
+            {"headline": "sports item", "type": "text"},
+        ],
+    )
+
+    print("COMP", public_company)
```
MarkLark86, line 229: PR includes `print` statement

MarkLark86 (approved):

tests/core/test_products.py
petrjasek: :man_facepalming:

PR #387, opened 2023-06-01
Search terms highlighted in Wire [CPCN-17]

devketanpro:


petrjasek:
this should actually go to develop branch, won't be part of hotfix release

assets/wire/components/WireListItem.jsx
```diff
@@ -261,7 +265,9 @@ class WireListItem extends React.Component {
 
                         {isExtended && (
                             <div className="wire-articles__item__text">
-                                <p>{shortText(item, 40, listConfig)}</p>
+                                {item.es_highlight ? <div
+                                    dangerouslySetInnerHTML={({__html:item.es_highlight.body_html && item.es_highlight.body_html[0]})}
```
petrjasek, line 269: does this contain only the paragraphs matching the text or it's the full body?

assets/ui/components/ArticleHeadline.jsx
```diff
@@ -3,7 +3,9 @@ import PropTypes from 'prop-types';
 
 export default function ArticleHeadline({item}) {
     return (
-        item.headline && <h2 className='wire-column__preview__headline'>{item.headline}</h2> || null
+        item.es_highlight && item.es_highlight.headline ? <div
+            dangerouslySetInnerHTML={({__html: item.es_highlight.headline[0]})}
+        /> : item.headline && <h2 className='wire-column__preview__headline'>{item.headline}</h2> || null
```
petrjasek, line 8: imo we should always end up with:
```
<h2 className='wire-column__preview__headline'>
```
so if there are highlights it should go inside that h2 so we keep the styling, same for slugline

newsroom/search.py
```diff
@@ -507,7 +507,11 @@ def prefill_search_highlights(self, search, req):
                     "lenient": True,
                 },
             )
-            elastic_highlight_query["fields"] = {"body_html": elastic_highlight_query["fields"]["body_html"]}
+            elastic_highlight_query["fields"] = {
+                "body_html": elastic_highlight_query["fields"]["body_html"],
```
petrjasek, line 511: lets avoid this field specific query so it might also work with advanced search, so the fields should be like:
```py
    "body_html": {},
    "headline": {},
    "slugline": {},
}
```
it might also need the change in the condition `if app.data.elastic.should_highlight(req) and query_string`
because we want this to work also with advanced search must contain and should contain fields

petrjasek (commented):
pls add some tests also including the advanced search queries

devketanpro:
> this should actually go to develop branch, won't be part of hotfix release

okay I create a PR for the develop branch.

assets/wire/components/WireListItem.jsx
devketanpro: we got paragraphs matching the text. so we shows first match here.

newsroom/search.py
devketanpro: > it might also need the change in the condition if app.data.elastic.should_highlight(req) and query_string
no, I checked we don;t need to change the condition , as it is worked with advanced search filter.

PR #388, opened 2023-06-01
Search terms highlighted in Wire [CPCN-17]

devketanpro:


newsroom/search.py
```diff
@@ -513,7 +513,12 @@ def prefill_search_items(self, search):
     def prefill_search_highlights(self, search, req):
         query_string = search.args.get("q")
         query_string_settings = app.config["ELASTICSEARCH_SETTINGS"]["settings"]["query_string"]
-        if app.data.elastic.should_highlight(req) and query_string:
+        advanced_search_query = search.args.get("advanced_search")
+        if (
+            (app.data.elastic.should_highlight(req) and query_string)
+            or advanced_search_query
+            and len(json.loads(advanced_search_query).get("fields", [])) != 0
+        ):
```
petrjasek, line 521: this should be rather:
```py
if app.data.elastic.should_highlight(req) and (
    query_string or
    (advanced_search and advanced_search.get("all")) or
    (advanved_search and advanced_search.get("any"))
):
```
the fields config is not so important there, it's rather the queries which are important

assets/ui/components/ArticleHeadline.jsx
```diff
@@ -3,7 +3,13 @@ import PropTypes from 'prop-types';
 
 export default function ArticleHeadline({item}) {
     return (
-        item.headline && <h2 className='wire-column__preview__headline'>{item.headline}</h2> || null
+        <h2 className='wire-column__preview__headline'>
```
petrjasek, line 6: in general I wouldn't change the behaviour when we return null if the field is empty, so should be rather:
```
return item.headline ? (
<h2 ...>...</h2>
) : null;
```

newsroom/search.py
devketanpro: this condition fails when we applied only advanced search,  we did not get `search.args.get("q")`

assets/ui/components/ArticleHeadline.jsx
devketanpro: Done

newsroom/search.py
petrjasek: which part fails? there is an `or` between `query_string` and advanced search checks so should be enough to have any of those

devketanpro: we got app.data.elastic.should_highlight(req)  = False

tests/core/test_wire.py
```diff
@@ -801,10 +801,59 @@ def test_wire_delete(client, app):
...
+                "body_html": "Story that involves cheese and onions",
+                "slugline": "That's the test slugline cheese",
+                "headline": "Demo Article",
+            }
+        ],
+    )
+    advanced_search_params = parse.quote('{"fields":[],"all":"demo"}')
+    url = f"/wire/search?advanced_search={advanced_search_params}&es_highlight=1"
+    resp = client.get(url)
+    data = json.loads(resp.get_data())
```
petrjasek, line 846: there is no assert on that data

PR #385, opened 2023-05-31
Css refactoring

fritzSF:

PR #389, opened 2023-06-02
Ts migration v2

tomaskikutis:


fritzSF:
@tomaskikutis @petrjasek @MarkLark86 I think that https://github.com/superdesk/newsroom-core/pull/385 has to be merged before this, otherwise a lot of work could be screwed up or lost.

MarkLark86:
There's an e2e test failing, and after running this locally I found the Permissions tab in Company management isn't rendering anything

PR #390, opened 2023-06-05
handle french accents in search

petrjasek:
setup elastic analyzer to handle non ascii characters

CPCN-24


newsroom/agenda/agenda.py
```diff
@@ -131,11 +132,11 @@ class AgendaResource(newsroom.Resource):
     }
 
     # content metadata
-    schema["name"] = events_schema["name"]
+    schema["name"] = metadata_schema["body_html"].copy()
```
MarkLark86, line 135: Probably need to bump the schema versions so elastic is updated https://github.com/superdesk/newsroom-core/blob/develop/newsroom/__init__.py#L21

MarkLark86 (approved):

PR #392, opened 2023-06-06
Fix : Request seat template looks awkward [CPCN-142]

devketanpro:


assets/styles/custom-forms.scss
```diff
@@ -39,6 +39,19 @@ $enable-shadows:true;
   flex: 1 1 auto;
 }
 
+
+.settingsWrap {
+  .input-group {
+   .btn {
+    height: 32px;
+    margin-right: 5px;
```
MarkLark86, line 47: This will not work as intended in RTL formats, also not sure if we should use `settingsWrap` as the parent selector.

@fritzSF @darconny any suggestions here?

darconny: @devketanpro  Yes it should be `margin-inline-end: 5px; `

fritzSF: Hey @devketanpro , I don't think `settingsWrap` i nedded here. Also, instead of the `input-group`  use `tags-list `. As for the button, please use this:
`<button class="icon-button icon-button--primary icon-button--small icon-button--bordered" aria-label="Add Product"><i class="icon--plus"></i></button>`. 
This should fix the layout issues. Thanks!

fritzSF (changes requested):

fritzSF (approved):

PR #393, opened 2023-06-06
[CPCN-198] Support advanced search params in topics

MarkLark86:


MarkLark86:
@petrjasek I changed the search URL param from `advancedSearch` to simply `search`, as the endpoint is `search` and so all usage of the param is `advanced`, otherwise there were differences bettwen Topic and URL and State attribute names

petrjasek (approved):

PR #394, opened 2023-06-06
Fix: More News button is not working. Once clicked, always spinning [NHUB - 293]

devketanpro:


assets/components/cards/render/CardRow.jsx
```diff
@@ -1,26 +1,56 @@
...
 
     componentDidMount() {
         if (this.props.isActive) {
             this.cardElem.scrollIntoView({behavior: 'instant', block: 'end', inline: 'nearest'});
         }
+
+        this.fetchUserData();
+    }
+
+    fetchUserData() {
```
MarkLark86, line 25: We can avoid this, as we should already be populating the user's products etc from the server on initial GET of the home page.
https://github.com/superdesk/newsroom-core/blob/develop/newsroom/wire/views.py#L133

If the required data is not there on page load, you can add it from the server and make sure it is added to the client/redux state

assets/components/cards/render/CardRow.jsx
```diff
@@ -13,14 +13,19 @@ class CardRow extends React.Component {
             this.cardElem.scrollIntoView({behavior: 'instant', block: 'end', inline: 'nearest'});
         }
     }
-
     render() {
+        const {title, product, children, userProducts,userType} = this.props;
+        let moreNews = this.props.moreNews;
+        if(userProducts.length === 0 && userType !== 'administrator'){
```
petrjasek, line 19: pls add space after `if`

newsroom/wire/views.py
```diff
@@ -141,6 +141,7 @@ def get_home_data():
         "cards": cards,
         "products": get_products_by_company(company) if company else [],
         "user": str(user["_id"]) if user else None,
+        "userProducts": user.get("products") if user else None,
```
petrjasek, line 144: this should probably also include company products without seats limit

assets/components/cards/render/CardRow.jsx
```diff
@@ -32,10 +37,18 @@ CardRow.propTypes = {
...
+    userProducts:state.userProducts,
+    userType: state.userType
+});
+
 CardRow.defaultProps = {
     moreNews: true
 };
 
-export default CardRow;
+export default connect(mapStateToProps)(CardRow);
```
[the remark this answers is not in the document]
devketanpro: we have to use connect because this component is called from many places , so sometimes we got `userProducts `and `userType` us `undefined`.

newsroom/wire/views.py
devketanpro: we already include company products here `"products": get_products_by_company(company) if company else [],`

petrjasek: yep sure it's there, but it's not taken into account for that more news button, is it?

devketanpro: yep we are using these products and filter that based on card config.

PR #395, opened 2023-06-08
update fields setting to show full body text in highlight [CPCN-17]

devketanpro:


newsroom/search.py
```diff
@@ -528,9 +529,9 @@ def prefill_search_highlights(self, search, req):
                 },
             )
             elastic_highlight_query["fields"] = {
-                "body_html": {},
-                "headline": {},
-                "slugline": {},
+                "body_html": field_settings,
+                "headline": field_settings,
+                "slugline": field_settings,
```
petrjasek, line 534: pls only configure fields which are selected in the `advanced_search["fields"]` config (or keep all if that config is empty)

newsroom/search.py
```diff
@@ -528,11 +528,15 @@ def prefill_search_highlights(self, search, req):
                     "lenient": True,
                 },
             )
-            elastic_highlight_query["fields"] = {
-                "body_html": field_settings,
-                "headline": field_settings,
-                "slugline": field_settings,
-            }
+            selected_fields = json.loads(advanced_search).get("fields") if advanced_search else []
```
petrjasek, line 531: imo it would be good to do that `json.loads` thing once in the beginning and not each time `advanced_search` is used for something

newsroom/search.py
```diff
@@ -513,12 +513,13 @@ def prefill_search_items(self, search):
     def prefill_search_highlights(self, search, req):
         query_string = search.args.get("q")
         query_string_settings = app.config["ELASTICSEARCH_SETTINGS"]["settings"]["query_string"]
-        advanced_search = search.args.get("advanced")
+        advanced_search = json.loads(search.args.get("advanced")) if search.args.get("advanced") else None
```
petrjasek, line 516: I would simplify it further:
```py
advanced_search = json.loads(search.args.get("advanced")) if search.args.get("advanced") else {}

if (...
 or advanced_search.get("all")
 or advanced_search.get("any")
)

...
selected_field = advanced_search.get("fields") or []
```

petrjasek (approved):

PR #400, opened 2023-06-09
Fix : disable more news button instead of hide full component [NHUB-293]

devketanpro:


assets/components/cards/render/CardRow.jsx
```diff
@@ -22,7 +22,7 @@ class CardRow extends React.Component {
 
         return (
             <div className='row' ref={(elem) => (this.cardElem = elem)}>
-                {moreNews && <MoreNewsButton title={title} product={product} />}
+                {<MoreNewsButton title={title} product={product} moreNews = {moreNews} />}
```
petrjasek, line 25: I suppose there is no need for surrounding `{}`

newsroom/search.py
```diff
@@ -467,8 +467,8 @@ def prefill_search_products(self, search):
                 search.products = get_product_by_id(
                     search.args["product"],
                     product_type=search.section,
-                    company_id=search.company.get("_id"),
```
petrjasek, line 470: you should still check if that product is assigned to user or a company, if not user shouldn't be able to see it

newsroom/search.py
```diff
@@ -548,14 +548,23 @@ def validate_request(self, search):
...
                 # Ensure that all the provided products are permissioned for this request
                 if not all(p in [c.get("_id") for c in search.products] for p in search.args["requested_products"]):
                     abort(404, gettext("Invalid product parameter"))
 
+    def is_product_assigned_to_company(self, search):
+        """Check the product is assigned to a company
+
+        :param SearchQuery search: The search query instance
+        """
+        product_id = search.products[0].get("_id")
```
petrjasek, line 564: imo this doesn't do anything if it's only called when `not len(search.products)`

newsroom/search.py
```diff
@@ -550,12 +550,23 @@ def validate_request(self, search):
                 abort(403, gettext("User does not belong to a company."))
             elif not len(search.products):
                 abort(403, gettext("Your company doesn't have any products defined."))
+            elif len(search.products) and self.is_product_assigned_to_company(search):
```
petrjasek, line 553: it should probably read like `and not is_product_assigned_to_company` instead of returning `True` when it's not assigned, that's confusing

newsroom/search.py
```diff
@@ -550,24 +551,42 @@ def validate_request(self, search):
...
+        """
+        user = data.user
+        company = data.company
+        product = data.args.get("product")
+
+        if user and company and product:
+            company_products_with_zero_seats = [p["_id"] for p in company.get("products", []) if p.get("seats") == 0]
+            user_specific_products = [p["_id"] for p in user.get("products", [])]
+
+            if company_products_with_zero_seats and not ObjectId(product) in user_specific_products:
```
petrjasek, line 576: looks to me you could do:
```python3
return product in user_products or product in company_products_with_zero_seats
```

devketanpro: yep

newsroom/search.py
```diff
@@ -550,12 +551,30 @@ def validate_request(self, search):
...
             elif search.args.get("requested_products"):
                 # Ensure that all the provided products are permissioned for this request
                 if not all(p in [c.get("_id") for c in search.products] for p in search.args["requested_products"]):
                     abort(404, gettext("Invalid product parameter"))
 
+    def is_validate_product(self, data):
+        """
+        Check product is assigned to the company but not to the respective user.
+
+        :param SearchQuery search: The search query instance
```
petrjasek, line 566: seems like a mismatch with the param name

devketanpro: my bad typo

newsroom/search.py
```diff
@@ -550,12 +551,30 @@ def validate_request(self, search):
...
+        """
+        user = data.user
+        company = data.company
+        product = data.args.get("product")
+
+        if user and company and product:
+            company_products_with_zero_seats = [p["_id"] for p in company.get("products", []) if p.get("seats") == 0]
```
petrjasek, line 573: for BC it might be better to just check `if not p.get("seats")` as `seats` might be undefined but then it should handle those as unlimited

petrjasek (approved):

PR #402, opened 2023-06-14
See the content of the body field which is surrounding the first matching search term [CPCN-18]

devketanpro:


assets/wire/utils.js
```diff
@@ -186,6 +186,27 @@ export function shortText(item, length=40, config) {
     return words.slice(0, length).join(' ') + (words.length > length ? '...' : '');
 }
 
+export function shortHighlightedtext(text, maxLength = 40) {
+    const spanRegex = /<span\s+class="es-highlight">([^<]+)<\/span>/i;
+    const match = text.match(spanRegex);
+  
+    if (match) {
+        const startIndex = text.indexOf(match[0]);
```
petrjasek, line 194: isn't this the same as `match.index` ?

petrjasek (commented):
can you provide some samples, ideally write some tests so it's obvious what it does.
imo it won't take the text before the matching string, but that was the idea to get a few words before and after so there is more context.

assets/wire/utils.js
devketanpro: yeah you are right !!

devketanpro:
currently with this changes it looks like this :- 
image

petrjasek:
I think this would be better done on some html tree level, you might get some broken html if you handle that as a text

devketanpro:
now with these changes it looks like this :- 
image

assets/wire/tests/WireApp.spec.js
```diff
@@ -57,3 +58,32 @@ describe('WireApp', () => {
         expect(actions.length).toBe(3);
     });
 });
+
+
+describe('shortHighlightedtext', () => {
+    it('returns truncated text with highlighted span', () => {
```
petrjasek, line 64: can you also add a test case for list and table?

petrjasek (approved):

PR #403, opened 2023-06-14
Fix : enable news more button for company unlimited seats [NHUB-293]

devketanpro:


newsroom/wire/views.py
```diff
@@ -144,6 +144,7 @@ def get_home_data():
         "userProducts": user.get("products") if user else [],
         "userType": user.get("user_type"),
         "company": company_id,
+        "companyProducts": company.get("products") if company else [],
```
petrjasek, line 147: imo we are already having it in the `products` param, don't we?

petrjasek: I see

petrjasek (approved):

PR #407, opened 2023-06-16
Fix : update permission, user type 'adminstrator' access all section in user profile [NHUB-291].

devketanpro:


assets/users/components/EditUser.jsx
```diff
@@ -45,7 +45,7 @@ function EditUserComponent({
         Object.keys(seats[companyId] || {}) :
         products.map((product) => product._id);
     const sections = companyId != null ?
-        companySections[companyId] || [] :
+        (isUserAdmin(user) ? allSections : companySections[companyId] || []) :
```
MarkLark86, line 48: This is only when editing a user. We need to fix the UserProfile so that existing admins can view all sections in the UserProfile (without requiring to edit their user to add this section)

devketanpro: you mean here ?
image

MarkLark86: Yes

devketanpro: got it

MarkLark86 (approved):

PR #413, opened 2023-06-22
STT text and picture genre to be shown on the planned coverages [STTNHUB-199]

devketanpro:


assets/agenda/components/AgendaCoverages.jsx
```diff
@@ -61,7 +61,7 @@ function AgendaCoveragesComponent({item, coverages, wireItems, actions, user, on
                     >
                         <span className={classNames('coverage-item__coverage-icon', WORKFLOW_COLORS[coverage.workflow_status])}>
                             <i className={`icon-small--coverage-${getCoverageIcon(coverage.coverage_type)} me-2`}></i>
-                            <span>{`${getCoverageDisplayName(coverage.coverage_type)}${getSlugline(coverage)}`}</span>
+                            <span>{`${coverage.genre.length !==0 ? gettext(coverage.genre[0].name) : getCoverageDisplayName(coverage.coverage_type)}${getSlugline(coverage)}`}</span>
```
MarkLark86, line 64: In this context, Genre is not the Coverage Type. Coverage type is the type of content that is being planed for (https://github.com/superdesk/newsroom-core/blob/develop/newsroom/web/default_settings.py#L281)

devketanpro: but we have to show that genre name instead of type, IMO that is ticket AC. and with these changes it looks like this 
Screenshot 2023-06-22 at 4 43 52 PM

MarkLark86: Your changes will break the system for other customers.
The genre should be shown, as well as the content type.

MarkLark86: From the ticket:
```I want to see what type of text has been planned (genre)```

The Text is the Content Type, and Genre explains what type of text would be expected

devketanpro: so you mean instead of removing the functionality of display content_type, we add genre also there right ?

devketanpro: > Your changes will break the system for other customers. The genre should be shown, as well as the content type.

I see, I'm sorry, because before I created this PR, I confirmed with Gideon about the changes and shared the screenshots with it.

devketanpro: Hi, @MarkLark86 ,can we do something like this ?
image

devketanpro:
Hi, @MarkLark86 ,can we do something like this ?
image

PR #416, opened 2023-06-26
[CPCN-47] SearchTips Panel

MarkLark86:


assets/styles/index.scss
```diff
@@ -3460,16 +3460,23 @@ a.wire-articles__versions {
         padding: 0;
         margin: 10px 0;
     }
+
```
MarkLark86, line 3463: @fritzSF I had to move these out, as the design that the customer provided had colours inside the dot points and not just in the block text

fritzSF: OK, no problem

assets/search/components/SearchBar.jsx
```diff
@@ -84,17 +84,25 @@ class SearchBar extends React.Component {
                         </div>
                     </form>
                 </div>
-                {this.props.toggleAdvancedSearchPanel == null ? this.props.toggleAdvancedSearchPanel : (
-                    <div className="mx-2 d-flex gap-2">
+                <div className="mx-2 d-flex gap-2">
+                    {this.props.toggleAdvancedSearchPanel == null ? this.props.toggleAdvancedSearchPanel : (
```
petrjasek, line 88: it's a bit confusing, imo would be better to have there explicit `? null : (`

petrjasek (approved):

PR #417, opened 2023-06-28
No Select all button to select all available products when assigning products to a standard user [CPCN-116]

devketanpro:


assets/users/components/EditUser.jsx
```diff
@@ -213,6 +213,11 @@ function EditUserComponent({
                                 title={gettext('Products')}
                                 testId="toggle--products"
                             >
+                                {isCompanyAdmin ? <div className="products-list__heading d-flex justify-content-between align-items-center">
+                                    <button type='button' name='selectAllBtn' className='nh-button nh-button--tertiary nh-button--small' onClick={onChange} >
+                                    Select All
```
petrjasek, line 218: pls use `gettext`

assets/users/reducers.js
```diff
@@ -102,6 +102,8 @@ export default function userReducer(state = initialState, action) {
             } else {
                 user.products = (user.products || []).filter((product) => product._id !== productId);
             }
+        } else if (field.includes('selectAllBtn')) {
+            user.products = state.products.filter((product)=> user.sections[product.product_type]);
```
petrjasek, line 106: not sure about the values you're assigning here, on user we only store `_id` and `section`, don't we?

devketanpro: We preserved `product._id` in `user.products` and used logic if user section is true _(eg:- agenda)_ so that in this case all agenda **products** are selected from the state and then assigned to 'user.products'.

assets/users/components/EditUser.jsx
devketanpro: yep, sorry I missed it

assets/users/reducers.js
petrjasek: yep sure I'm rather wondering if there should be as well:
```ts
.map((product) => {_id: product._id, section: product.product_type})
```
because we don't store the whole product objects in `user.products`

petrjasek (approved):

PR #419, opened 2023-06-30
implement topic folders UI

petrjasek:
CPCN-133

hi @MarkLark86 if you have time to pick any of those that would be great,
but I should have time to handle it when I'm back as well, so no rush.
at the end I've added page reload when closing the profile app instead of
doing some syncing, imo we should either merge those apps or run
the user profile on its own url.

- [x] add search info to topic sidebar
- [x] add e2e


MarkLark86:
@tomaskikutis As Petr and myself worked on this together, I thought it best for someone else to review

petrjasek:
hi @MarkLark86 I've merged some recent changes so pls check

MarkLark86 (approved):

PR #420, opened 2023-06-30
update functionality to showing subject in the agenda list preview based on uiconfig [STTNHUB-200]

devketanpro:
https://github.com/superdesk/newsroom-app-stt/pull/16

### DIFF
--- a/assets/agenda/components/AgendaListItemIcons.jsx
+++ b/assets/agenda/components/AgendaListItemIcons.jsx
@@ -18,8 +18,6 @@ import AgendaListCoverageItem from './AgendaListCoverageItem';
 import AgendaLocation from './AgendaLocation';
 
 import {gettext} from 'utils';
-import {connect} from 'react-redux';
-import {listConfigSelector} from '../../ui/selectors';
 
 class AgendaListItemIcons extends React.Component {
     constructor(props) {
@@ -105,20 +103,18 @@ class AgendaListItemIcons extends React.Component {
                     </div>
                 )}
 
-                <div>
-                    {subject.length != 0 && (
-                        <span className="label label--rounded">
-                            {subject.map((item, index) => (
-                                <span
-                                    key={index}
-                                    className={`subject--${item.qcode}`}
-                                >
-                                    {item.name}
-                                </span>
-                            ))}
-                        </span>
-                    )}
-                </div>
+                {subject.length !== 0 && (
+                    <div>
+                        {subject.map((item, index) => (
+                            <span
+                                key={index}
+                                className={`label label--rounded subject--${item.qcode}`}
+                            >
+                                {item.name}
+                            </span>
+                        ))}
+                    </div>
+                )}
 
                 {state.attachments > 0 && (
                     <div className='d-flex align-items-center wire-articles__item__icons--dashed-border'>
@@ -153,10 +149,6 @@ AgendaListItemIcons.propTypes = {
     listConfig : PropTypes.object,
 };
 
-const mapStateToProps = (state) => ({
-    listConfig: listConfigSelector(state),
-});
-
 AgendaListItemIcons.defaultProps = {isMobilePhone: false};
 
-export default connect(mapStateToProps)(AgendaListItemIcons);
+export default AgendaListItemIcons;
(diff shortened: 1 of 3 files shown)


assets/agenda/components/AgendaListItemIcons.jsx
```diff
@@ -87,6 +105,21 @@ class AgendaListItemIcons extends React.Component {
                     </div>
                 )}
 
+                <div>
+                    {subject.length != 0 && (
+                        <span className="label label--rounded">
```
petrjasek, line 110: I would expect each one to be a separate label, not just 1 label for all. and to have that `subject--` class on the same level so it can be used as a modifier for that

assets/agenda/components/AgendaListItemIcons.jsx
```diff
@@ -117,8 +150,13 @@ AgendaListItemIcons.propTypes = {
...
 };
 
+const mapStateToProps = (state) => ({
+    listConfig: listConfigSelector(state),
+});
+
 AgendaListItemIcons.defaultProps = {isMobilePhone: false};
 
-export default AgendaListItemIcons;
+export default connect(mapStateToProps)(AgendaListItemIcons);
```
petrjasek, line 162: the list config would be better passed from some higher level component, something like `AgendaList`

assets/agenda/components/AgendaListItemIcons.jsx
devketanpro: these are the outer label class which Fritz suggested. and we can modify the label here 
https://github.com/superdesk/newsroom-core/pull/420/files/7347c50aba955604de9bffeec2bfee8ccfc3d44d#diff-4799aff35d2a8cf74c640e3308b36e2b2eadec8a6a4389f3868bd7edf029044fR114

assets/agenda/components/AgendaListItemIcons.jsx
devketanpro: okay

assets/agenda/components/AgendaListItemIcons.jsx
```diff
@@ -57,10 +57,26 @@ class AgendaListItemIcons extends React.Component {
         };
     }
 
+    getItemByschema(props) {
+        const {listConfig} = props;
+        if (listConfig && listConfig.subject && listConfig.subject.scheme) {
```
MarkLark86, line 62: We should probably support the ability to show ALL subjects.
Meaning, if `listConfig.subject.scheme` is defined, then we filter the subjects, otherwise we return all subjects if `listConfig.subject` is defined

assets/agenda/components/AgendaListItemIcons.jsx
```diff
@@ -87,6 +103,19 @@ class AgendaListItemIcons extends React.Component {
                     </div>
                 )}
 
+                {subject.length !== 0 && (
+                    <div>
+                        {subject.map((item, index) => (
+                            <span
+                                key={index}
```
MarkLark86, line 110: Using the array `index` as the `key` is not advisable as it's not a direct unique key to the item itself, it is better to use `item.qcode` in this case.

assets/agenda/components/AgendaListItemIcons.jsx
devketanpro: Okay

assets/agenda/components/AgendaListItemIcons.jsx
```diff
@@ -57,10 +57,28 @@ class AgendaListItemIcons extends React.Component {
...
+        const {scheme} = listConfig && listConfig.subject || {};
+
+        if (scheme) {
+            const subjects = getSubjects(props.item);
+            const filteredSubjects = subjects.filter(item => {
+                return Array.isArray(scheme) ? scheme.includes(item.scheme) : item.scheme === scheme;
+            });
+            return filteredSubjects;
+        }
+        if (listConfig.subject) {
```
MarkLark86, line 71: Here you're assuming `listConfig` is an object, but in line 62 you're covering the scenario if `listConfig` is not defined, in which case this line would throw an exception.

It might be better to surround this entire `getSubjects` section in an outer `if (listConfig && listConfig.subject)`, such as:
```javascript
if ((listConfig || {}).subject != null) {
    const scheme = listConfig.subject.scheme;
    const subjects = getSubjects(props.item);
    return scheme == null ?
        subjects :
        subjects.filter((item) => Array.isArray(scheme) ?
            scheme.includes(item.scheme) :
            item.scheme === scheme
        );
}

return [];
```

PR #421, opened 2023-07-03
Fix : update split pattern for search term highlighted text [CPCN-18]

devketanpro:


assets/wire/utils.js
```diff
@@ -205,7 +205,7 @@ export function shortHighlightedtext(html, maxLength = 40) {
   
     while (node && count < maxLength) {
         const content = node.textContent.trim();
-        const words = content.split(/\s/).filter(w => w);
```
MarkLark86, line 208: Why change from split by `whitespace` to split by `followed by [A-Z]`?

devketanpro: actually that split not give us desired output in the case if search term is found in the last line of the paragraph and paragraph is too long.

MarkLark86: It would be good to have some unit tests for this

devketanpro: sure

MarkLark86 (approved):

PR #422, opened 2023-07-03
instructions for development setup

tomaskikutis:


assets/index.js
```diff
@@ -4,3 +4,12 @@ import 'primereact/resources/primereact.min.css';
 if (isTouchDevice()) {
     document.documentElement.classList.add('no-touch');
 }
+
+
+window.sectionNames = {
```
MarkLark86, line 9: Won't this override the configurable section names from here: https://github.com/superdesk/newsroom-core/blob/develop/newsroom/templates/scripts.html#L8

MarkLark86: Defined in `settings.py` here https://github.com/superdesk/newsroom-core/blob/develop/newsroom/web/default_settings.py#L551

tomaskikutis: I'll remove it, I'm only keeping it for testing until I figure out how to get it running without this. I've marked the PR as draft

DEVELOPMENT.md
```diff
@@ -0,0 +1,74 @@
+# Setting up newsroom for development
+
+## Repositories
+
+Use `develop` branch of `newsroom-core` repository and `master` branch of `superdesk-app`.
```
MarkLark86, line 5: Typo: `superdesk-app` should be `newsroom-app`

MarkLark86: We should also be using the `develop` branch from `newsroom-app`, as that is what `nra-develop` instance is built off. If there is an issue with this branch, we should fix it

DEVELOPMENT.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+WEBPACK_ASSETS_URL=http://localhost:8080
+WEBPACK_SERVER_URL=http://localhost:8080
+SECRET_KEY=newsroom
+```
+
+## Starting local server
+* navigate to `newsroom-app`
+* make sure you're on `master` branch
```
MarkLark86, line 47: `develop` branch

DEVELOPMENT.md
```diff
@@ -0,0 +1,74 @@
...
+* `cd server`
+* create virtual environment `python3 -m venv env`
+* activate virtual environment `source env/bin/activate`
+* install python dependencies - `pip install -r dev-requirements.txt`
+* link local server - `pip install -Ue <path-to-newsroom-core>`
+* create a user - `python manage.py create_user admin@localhost.com admin John Doe true`
```
MarkLark86, line 53: We could base this on what we use in FireQ instances (see [prepopulate](https://github.com/superdesk/fireq/blob/master/tpl/newsroom-app/prepopulate.sh) script).
```bash
python manage.py initialize_data
python manage.py data_upgrade
python manage.py create_user admin@example.com admin Admin Admin true || :
python manage.py schema_migrate || true
```

- This creates the user`admin@example.com` instead of `admin@localhost.com` (same user email as our test instances)
- Running `initialize_data` prepares the system (including elastic)
- Running `data_upgrade` runs any migration scripts
- Running `schema_migrate` runs Mongo/Elastic schema migrations

These last 2 above are important to run when there are changes to the DB schema (see [SCHEMA_VERSIONS](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/__init__.py#L19))

DEVELOPMENT.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Create a subscriber
+
+* Go to Superdesk settings -> Subscribers -> Add new
+* Set "Target Type" to "All"
+* Set "Sequence number settings" to 1 and 10.
+* Click "Add new destination"
+* Set "Format" to "Newsroom NINJS"
+* Set "Delivery type" to "HTTP Push"
+* Set "Resource URL" to "http://localhost:5050/push"
```
MarkLark86, line 67: Also requires to set the `ASSETS URL` and `SECRET_TOKEN`. In this example, these two would be:
* **ASSETS URL**: http://localhost:5050/push_binary
* **SECRET TOKEN**: newsroom

MarkLark86 (approved):

PR #424, opened 2023-07-05
Improve user creation user settings

thecalcc:
CPCN-150


assets/company-admin/components/CompanyAdminApp.tsx
```diff
@@ -184,26 +217,32 @@ class CompanyAdminAppComponent extends React.Component<any, any> {
...
-                                    hideFields={getConfig('allow_companies_to_manage_products') ?
-                                        ['company'] :
-                                        ['company', 'sections', 'products']
-                                    }
-                                />
-                            )}
+                            {
+                                this.props.userToEdit !== null && (
+                                    <EditUser
+                                        original={this.props.usersById[this.props.userToEdit._id ?? ''] || {}}
```
tomaskikutis, line 223: Don't default to an empty string. Maybe the idea is that when you're creating a new user, it is `null`. That would make sense to keep it then. If you convert it to a string there would no longer be a way to differentiate between a real ID and ID missing. Unless you check the length, but that's not a good approach.

assets/company-admin/components/CompanyAdminApp.tsx
```diff
@@ -32,8 +31,38 @@ import {UserListSortFilter} from 'users/components/filters/UserListSortFilter';
...
+    sortDirection: any;
+    setSort: any;
+    toggleSortDirection: any;
+    productId: any;
+    setProductFilter: any;
+    companies: any;
+    totalUsers: any;
+}
+
+class CompanyAdminAppComponent extends React.Component<IProps, any> {
```
tomaskikutis, line 65: state only has a few fields, if you're touching the component already it'd be nice to type state too

assets/users/components/EditUser.tsx
```diff
@@ -314,26 +334,6 @@ function EditUserComponent({
     );
 }
 
-EditUserComponent.propTypes = {
```
tomaskikutis, line 317: move info from propTypes to `IProps`

assets/company-admin/components/CompanyAdminApp.tsx
```diff
@@ -213,38 +252,6 @@ class CompanyAdminAppComponent extends React.Component<any, any> {
     }
 }
 
-CompanyAdminAppComponent.propTypes = {
```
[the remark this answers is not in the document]
thecalcc: I have moved it to IProps

assets/users/components/EditUser.tsx
thecalcc: I have already moved that info to IProps

thecalcc:
FIX CI & E2E before merging

e2e/package-lock.json
tomaskikutis, line 1: add npm rc and disallow package lock in this dir - with package lock it would run old tests or you'd need to regenerate the lock file on every commit

https://github.com/superdesk/superdesk/blob/master/client/.npmrc

e2e/running-e2e-locally.md
```diff
@@ -0,0 +1,31 @@
+# Terminal 1 - starting the client:
+```
+cd e2e
```
tomaskikutis, line 3: looks like npm install step is missing

tomaskikutis (approved):

tomaskikutis (approved):

thecalcc:
Don't merge until ts-migration-v2 is checked by QA.

PR #427, opened 2023-07-11
Fix : Search terms not highlighted completely in wire list view [CPCN-223]

devketanpro:


assets/wire/utils.js
```diff
@@ -218,12 +212,28 @@ export function shortHighlightedtext(html, maxLength = 40) {
   
         node = treeWalker.nextSibling();
     }
-  
     const truncatedText = text.trim();
-    const highlightedText = span.outerHTML;
-    const output = truncatedText + (count > maxLength ? '' : '...');
-    return output.replace(span.textContent, highlightedText);
-}  
+    highlightedSpans = Array.from(parentElement.querySelectorAll('span.es-highlight'));
```
MarkLark86, line 216: Why convert to an Array? The `<element>.querySelectorAll` function returns an instance of [NodeList](https://developer.mozilla.org/en-US/docs/Web/API/NodeList) which already has a `forEach` function defined.

devketanpro: my bad

MarkLark86 (approved):

PR #428, opened 2023-07-13
As a user I want to see the search terms highlighted in the Agenda section [CPCN-197]

devketanpro:


newsroom/search/service.py
```diff
@@ -531,9 +531,14 @@ def prefill_search_highlights(self, search, req):
             selected_field = advanced_search.get("fields") or []
             if not selected_field:
                 elastic_highlight_query["fields"] = {
+                    # wire
                     "body_html": field_settings,
                     "headline": field_settings,
                     "slugline": field_settings,
+                    # Agenda
+                    "description_text": field_settings,
```
MarkLark86, line 539: We should probably also add `definition_long` here (as well as [here](https://github.com/superdesk/newsroom-core/pull/428/files#diff-96efb764531481a220668aafcceb07e7be470513ed3df13e92c55870bb9241b2R599))

devketanpro: sure

MarkLark86 (approved):

PR #429, opened 2023-07-14
Personalize Dashboard MVP

thecalcc:
CPCN-59


petrjasek:
this got closed automatically, reopening

assets/components/PersonalizeHomeModal.tsx
```diff
@@ -0,0 +1,390 @@
...
+                    }
+                </FormToggle> */}
+            </div>
+        ) : (
+            NoSearchMatches()
+        );
+
+        const selectedTopics = (
+            <div className="simple-card__list pt-3">
+                {this.state.selectedTopics.map((topicId: string) => wireTopics.find((topic) => topic._id === topicId)).map((topic) => {
```
tomaskikutis, line 236: Can you make this into a function that wouldn't return nulls and would only contain 2 iterations rather than iterating `wireTopics` as many times as there are selected topic IDs?

assets/components/PersonalizeHomeModal.tsx
```diff
@@ -0,0 +1,356 @@
...
+import CheckboxInput from './CheckboxInput';
+import {IUser, IUserDashboard} from 'interfaces/user';
+import {RadioButtonGroup} from 'features/sections/SectionSwitch';
+import {modalFormInvalid, modalFormValid} from 'actions';
+import {updateUser} from 'users/actions';
+import {getCurrentUser} from 'company-admin/selectors';
+import {IAgendaState} from 'agenda/reducers';
+import {ITopic} from 'interfaces/topic';
+
+interface IStateProps {
```
tomaskikutis, line 17: `IPropsRedux`,  `IPropsFromRedux` or `IPropsConnected` would be a better name.

assets/components/PersonalizeHomeModal.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+type IProps = IStateProps & IDispatchProps & IOwnProps;
+
+interface IState {
+    selectedTopicIds: Array<string>;
+    activeSection: string;
+    searchTerm: string;
+    isLoading: boolean;
+    wireTopics: Array<ITopic>;
+    wireTopicsById: Map<string, ITopic>;
```
tomaskikutis, line 40: `wireTopicsById: Map<ITopic['_id'], ITopic>;` is slightly more explicit.

Another thing - if data is read-only we shouldn't be putting it in state.

tomaskikutis, line 40: We should add `Immutable.js` to the project. Then we could use an ordered map instead of having to have 2 variables for the same data that need to be kept in sycn.

assets/components/PersonalizeHomeModal.tsx
```diff
@@ -0,0 +1,356 @@
...
+};
+
+class PersonalizeHomeModal extends React.Component<IProps, IState> {
+    constructor(props: any) {
+        super(props);
+
+        const [wireTopics, agendaTopics] = partition(this.props.topics, (topic) => topic.topic_type === 'wire');
+        const wireTopicsById = new Map();
+
+        wireTopics.forEach((topic) => wireTopicsById.set(topic._id, topic));
```
tomaskikutis, line 60: you could also do it in immutable way

```
const wireTopicsById = new Map(wireTopics.map((topic) => [topic._id, topic]));
```

assets/components/PersonalizeHomeModal.tsx
```diff
@@ -0,0 +1,356 @@
...
+    isLoading: boolean;
+    wireTopics: Array<ITopic>;
+    wireTopicsById: Map<string, ITopic>;
+}
+
+const MAX_SELECTED_TOPICS = 6;
+
+const getSelectedTopicIds = (user: IUser) => {
+    if (user.dashboards?.length) {
+        return user.dashboards[0].topic_ids || [];
```
tomaskikutis, line 47: if your intent is to fallback to a default value, it's better to use  `??` than `||`. Pipes check for falsy which can be more error prone https://stackoverflow.com/a/63924076

assets/components/PersonalizeHomeModal.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+        this.props.saveUser({dashboards: [newDashboard]});
+
+        this.setState({isLoading: false});
+    }
+
+    getSelectedTopics() {
+        const topics: Array<ITopic> = [];
+
+        this.state.selectedTopicIds.map((topicId) => this.state.wireTopicsById.get(topicId)).map((topic) => {
```
tomaskikutis, line 111:
```
this.state.selectedTopicIds.map((topicId) => this.state.wireTopicsById.get(topicId)).map((topic) => {
                                                                                  // ^ should be forEach
```

assets/components/PersonalizeHomeModal.tsx
```diff
@@ -0,0 +1,356 @@
...
+                        </p>
+                    </figcaption>
+                </div>
+            </div>
+        );
+
+        const filteredTopics = (topics: Array<ITopic>) =>
+            topics.filter(({is_global}) => this.state.activeSection === '1' ? is_global != true : is_global === true);
+
+        const searchMatches = filteredTopics(this.wireTopics ?? [])
```
tomaskikutis, line 156: I see you apply a fallback to an empty array here, but not on line 190. Either we should type it as

```
wireTopics: Array<ITopic> | null;
```

or apply the default in the constructor and keep the type non-nullable.

tomaskikutis (approved):
> remove extra typing [7c5d696](https://github.com/superdesk/newsroom-core/pull/429/commits/7c5d696e9c82f8c01aeff01fa67ec7f0cccd7835)

I'd keep it. Typescript can often infer it, but for function arguments and return type it's useful to type it manually to guard against future changes. TS can sometimes infer too broadly and make it compile when it shouldn't.

tomaskikutis (approved):

PR #431, opened 2023-07-18
Fix : Section filters and products should not be highlighted during search in wire section [NHUB-320]

devketanpro:


newsroom/search/service.py
```diff
@@ -514,37 +513,45 @@ def prefill_search_items(self, search):
...
-                    "description_text": field_settings,
-                    "definition_short": field_settings,
-                    "name": field_settings,
-                    "definition_long": field_settings,
+            for field in fields_to_highlight:
+                search.source["highlight"]["fields"][field] = {
+                    "number_of_fragments": 0,
+                    "highlight_query": {
+                        "query_string": {
+                            "query": query_string or advanced_search.get("all") or advanced_search.get("any"),
```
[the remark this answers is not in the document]
devketanpro: ok

newsroom/search/service.py
```diff
@@ -511,21 +511,33 @@ def prefill_search_items(self, search):
...
+                should_queries.append(query_string(query))
+            all_query = advanced_search.get("all")
+            if all_query:
+                should_queries.append(query_string(all_query))
+            any_query = advanced_search.get("any")
+            if any_query:
+                should_queries.append(query_string(any_query))
+
+            search.query.setdefault("bool", {})
+            search.query["bool"]["should"] = should_queries
```
petrjasek, line 533: there is a helper to create such query, would be better to use that one:
https://github.com/superdesk/newsroom-core/blob/develop/newsroom/search/service.py#L708

devketanpro: Ok I'll check

newsroom/search/service.py
```diff
@@ -550,14 +550,16 @@ def prefill_search_highlights(self, search, req):
...
-                    "highlight_query": {"bool": {"should": should_queries}},
+                    "highlight_query": {"bool": {"filter": highlight_search.query["bool"]["filter"]}},
                 }
 
-            search.source["highlight"]["pre_tags"] = ['<span class="es-highlight">']
-            search.source["highlight"]["post_tags"] = ["</span>"]
-            search.source["highlight"]["require_field_match"] = False
+            highlight_search.source["highlight"]["pre_tags"] = ['<span class="es-highlight">']
+            highlight_search.source["highlight"]["post_tags"] = ["</span>"]
+            highlight_search.source["highlight"]["require_field_match"] = False
```
petrjasek, line 560: not sure if we need this, it should probably only highlight those requested fields

petrjasek (approved):

newsroom/search/service.py
devketanpro: yep

petrjasek (approved):

PR #432, opened 2023-07-18
generate manifest.json file on disk when using webpack dev server

petrjasek:
so it can be used inside docker


MarkLark86 (approved):

MarkLark86:
@petrjasek Why milestone `v3.0` and not `v2.5-beta2` (considering this is merged into `develop`)?

petrjasek:
you're right, got puzzled with those versions

PR #434, opened 2023-07-19
Update DEVELOPMENT.md

petrjasek:


tomaskikutis:
It would be good now to number the steps and mention which steps need to be skipped if backend-inside-docker option is chosen.

petrjasek:
is it more obvious now @tomaskikutis ?

tomaskikutis (approved):
yep, much better now

PR #435, opened 2023-07-19
allow wildcards in advanced search fields

petrjasek:
CPCN-222


MarkLark86 (commented):
We might also need to change the front-end to split keywords when they're grouped together using quotes, such as `"Sydney Weather" Today`: https://github.com/superdesk/newsroom-core/blob/develop/assets/search/components/SearchResultsBar/SearchResultsAdvancedSearchRow.jsx#L10

MarkLark86 (approved):

PR #438, opened 2023-07-24
fix folders dropdown when switching from user to company topic

petrjasek:
CPCN-239

### DIFF
--- a/assets/search/components/TopicEditor.jsx
+++ b/assets/search/components/TopicEditor.jsx
@@ -8,7 +8,7 @@ import {gettext, notify} from 'utils';
 import {canUserEditTopic} from 'topics/utils';
 import types from 'wire/types';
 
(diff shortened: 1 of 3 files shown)


assets/search/components/TopicEditor.jsx
```diff
@@ -127,6 +127,11 @@ class TopicEditor extends React.Component {
                 !get(topic, field) :
                 event.target.value;
 
+            if (field === 'is_global') {
```
MarkLark86, line 130: If the user is currently in `My Topics`, won't this clear the list of user folders for the list view? Similar scenario, if the user is in `Company Topics` and disables the  `Share with my Company` checkbox?

We might need to have both user and company topics loaded in the store, so the list and form can use the respective folder type

petrjasek: yep you are right, will change that

MarkLark86 (approved):

PR #443, opened 2023-07-26
Complete list of countries [NHUB-327]

devketanpro:


assets/components/SelectInput.jsx
```diff
@@ -30,7 +30,10 @@ const SelectInput = ({
                         <option value="">{defaultOption}</option>
                     }
                     {options.map((option) => {
-                        return <option key={option.value} value={option.value}>{gettext(option.text)}</option>;
+                        return <option
```
MarkLark86, line 33: To keep this component simple, it would be best to convert the options before passing them onto this component, so this component only uses `value` and `text`

newsroom/companies/utils.py
```diff
@@ -23,3 +25,12 @@ def get_company_product_ids(company: Company) -> List[ObjectId]:
         ],
         key=lambda o: str(o),
     )
+
+
+def load_countries_list():
+    with open(get_filepath("vocabularies.json")) as f:
```
MarkLark86, line 31: I'm not sure it's good to load a file on every `GET` request to this view, especially when this file won't change during the runtime of the environment.

@petrjasek What are your thoughts on finally having a `vocabularies` resource in Newshub? Or should we load this file in the config, allowing app repo's to override?

MarkLark86 (changes requested):

assets/components/SelectInput.jsx
devketanpro: yep

newsroom/companies/utils.py
MarkLark86: Also, there is no `vocabularies.json` file included in this PR

devketanpro: here we are using superdesk-core `vocabularies.json` file.

petrjasek: I don't see much use for it to be editable in newsroom atm, but maybe some way to define what metadata should be stored for companies/users eventually might require that. for now I would keep it static

petrjasek (commented):
seems like the qcodes we have in superdesk-core don't match the old ones we had hardcoded, so we will need some upgrade script to fix the old values in the db

MarkLark86 (approved):

PR #445, opened 2023-07-26
Fix tab switching in products and global topics

thecalcc:
NHUB-331


tomaskikutis:
let's wait until all occurances are checked

PR #446, opened 2023-07-26
add dashboards field to user schema

petrjasek:
CPCN-240

### DIFF
--- a/newsroom/users/users.py
+++ b/newsroom/users/users.py
@@ -79,6 +79,20 @@ class UsersResource(newsroom.Resource):
             "type": "dict",
             "nullable": True,
         },
+        "dashboards": {
+            "type": "list",
+            "schema": {
+                "type": "dict",
+                "schema": {
+                    "name": {"type": "string"},
+                    "type": {"type": "string"},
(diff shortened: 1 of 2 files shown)


devketanpro (approved):

newsroom/users/users.py
```diff
@@ -79,6 +79,20 @@ class UsersResource(newsroom.Resource):
             "type": "dict",
             "nullable": True,
         },
+        "dashboards": {
```
MarkLark86, line 82: This field would need to be added to [newsroom.users.users.USER_PROFILE_UPDATES](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/users/users.py#L99) to allow users to update their dashboard settings.

If these settings are to be managed via the [/users/<_id>](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/users/views.py#L171) endpoint then `dashboards` will also need to be added to the [UserForm](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/users/forms.py#L27) class, otherwise a separate endpoint will be needed.

PR #447, opened 2023-07-26
Ts migration v2

tomaskikutis:


petrjasek:
hi @MarkLark86 looks good to me now, what about you? would it be an issue merging it before you finish notification changes?

PR #448, opened 2023-07-28
[CPCN-232] fix(ui): Reload items when opening AdvancedSearch panel

MarkLark86:
This is required as we're clearing the TopSearch bar query, and item's weren't reflecting this


petrjasek:
which makes me think if it all makes sense, because if you don't do anything in the advanced modal you might end up on completely different screen

petrjasek (approved):

MarkLark86:
> which makes me think if it all makes sense, because if you don't do anything in the advanced modal you might end up on completely different screen

Yeah I don't like the UX either.

PR #452, opened 2023-08-02
populate user personal dashboard data on home page

petrjasek:
CPCN-249

### DIFF
--- a/newsroom/search/service.py
+++ b/newsroom/search/service.py
@@ -882,5 +882,4 @@ class BaseSearchService(Service):
         search.args["aggs"] = "false"
         self.gen_source_from_search(search)
         internal_req = self.get_internal_request(search)
-        print("REQ", internal_req.args)
         return self.internal_get(internal_req, search.lookup)
--- a/newsroom/search/service.py
+++ b/newsroom/search/service.py
@@ -787,58 +787,17 @@ class BaseSearchService(Service):
         section_filters = get_resource_service("section_filters").get_section_filters_dict()
 
         for topic in topics:
-            search = SearchQuery()
-
             user = users.get(str(topic["user"]))
             if not user:
                 continue
 
-            search.user = user
-            search.is_admin = is_admin(user)
-            search.company = companies.get(str(user.get("company", "")))
+            company = companies.get(str(user.get("company", "")))
 
-            search.query = deepcopy(query)
-            search.section = topic.get("topic_type")
-
-            self.prefill_search_products(search)
-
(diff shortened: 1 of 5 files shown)


newsroom/wire/views.py
```diff
@@ -161,6 +193,7 @@ def get_home_data():
         "topics": topics,
         "ui_config": get_resource_service("ui_config").get_section_config("home"),
         "groups": app.config.get("WIRE_GROUPS", []),
+        "personalizedDashboards": get_personal_dashboards_data(user, company, topics),
```
MarkLark86, line 196: Loading the items on initial load goes against the responsive changes we did previously. At the moment we leave it to the client to request these items, and show a loading spinner until the items arrive on the client.

newsroom/search/service.py
```diff
@@ -866,3 +825,61 @@ def get_matching_topics_for_item(self, topics, users, companies, query):
             raise
 
         return topic_matches
+
+    def get_topic_query(self, topic, user, company, section_filters, query=None):
```
MarkLark86, line 829: I've done something similar with my work on scheduled notifications. And as this is in the typescript branch, we will have merge conflicts. In this case, would it be better to base/merge this into the `develop` branch first then cherry-pick into typescript branch.

This way I can rebase from latest develop merging your changes with mine

newsroom/wire/views.py
```diff
@@ -134,6 +134,38 @@ def delete_dashboard_caches():
         app.cache.delete(f"{HOME_ITEMS_CACHE_KEY}{company['_id']}")
 
 
+def get_personal_dashboards_data(user, company, topics):
+    section_filters = get_resource_service("section_filters").get_section_filters_dict()
+
+    def get_topic_items(topic):
+        query = superdesk.get_resource_service("wire_search").get_topic_query(topic, user, company, section_filters)
```
MarkLark86, line 141: Should we also provide a `size` param as to limit the number of items returned for these homepage cards?

newsroom/wire/views.py
MarkLark86: We are currently doing that loading in the dashboard though.

And I agree about caching, similar to what we have with current dashboard items

newsroom/search/service.py
petrjasek: ok np

newsroom/wire/views.py
petrjasek: I have it set in the `get_items_by_query` method

PR #457, opened 2023-08-04
Fix advanced search panel

thecalcc:
NHUB-331


petrjasek:
imo we don't need it after https://github.com/superdesk/newsroom-core/pull/464

PR #458, opened 2023-08-04
CSS fixes and improvements

fritzSF:


newsroom/templates/login.html
```diff
@@ -2,7 +2,7 @@
 
 {% block contentMain %}
 
-<div class="container-fluid py-5 overflow-auto">
+<div class="container-fluid py-5 overflow-auto nh-login-page__content">
```
MarkLark86, line 5: **Note:** These class additions will also need to be added to customer repos, where these templates are overridden. (such as https://github.com/superdesk/newsroom-app-cp/blob/cn-develop/server/theme/login.html#L5)

MarkLark86 (approved):

newsroom/templates/login.html
fritzSF: Thanks @MarkLark86, I'll keep that in mind.

PR #470, opened 2023-08-10
Fix : If 'Allow companies to manage their own user product permissions' is FALSE, there should be no option to manage Sections and Products on user level [NHUB-330]

devketanpro:


devketanpro:
Don't need a fix , it is not a bug
https://sofab.atlassian.net/browse/NHUB-330?focusedCommentId=123840

PR #474, opened 2023-08-14
User creation settings improvements

thecalcc:
CPCN-150

Reopened in favour of https://github.com/superdesk/newsroom-core/pull/424


assets/users/components/EditUser.tsx
```diff
@@ -334,38 +364,14 @@ function EditUserComponent({
...
-    companySections: PropTypes.object,
-    seats: PropTypes.object,
-};
-
-EditUserComponent.defaultProps = {
-    hideFields: [],
 };
 
-const mapStateToProps = (state: any) => ({
+const mapStateToProps = (state: IReduxStateProps) => ({
```
petrjasek, line 369: not sure about the typing here, shouldn't this `IReduxStateProps` define what is returned from `mapStateToProps` and not what is passed there? it could be the same but in general this seems confusing, that state should be rather defined in the reducer

thecalcc: No I just typed the state object with that interface here since state's type isn't infered from anywhere. The return type of `mapStateToProps` is infered, I could add it explicitly but it won't make a difference.

thecalcc: Eventually yes, we should move it there, but now it would be too time consuming to debug the whole app to get all the possible values of each state prop. I think it's a good middle ground for now

assets/users/components/EditUser.tsx
```diff
@@ -25,16 +25,11 @@ import {getUserStateLabelDetails} from 'company-admin/components/CompanyUserList
...
 const getCompanyOptions = (companies: any) => companies.map((company: any) => ({value: company._id, text: company.name}));
 
-interface IReduxStateProps {
-    allSections?: Array<any>;
-    companySections?: any;
-    seats?: any;
-}
-
-interface IProps extends IReduxStateProps {
+interface IProps extends IUserProfileStore {
```
petrjasek, line 32: I'm not sure about this part, the component won't have access to the whole state, this should be extending the part returned from the mapStateToProps

assets/users/components/EditUser.tsx
```diff
@@ -25,16 +25,11 @@ import {getUserStateLabelDetails} from 'company-admin/components/CompanyUserList
 
 import {companyProductSeatsSelector, companySectionListSelector, sectionListSelector} from 'company-admin/selectors';
 import {IUser} from 'interfaces/user';
+import {IUserProfileStore, IUserProfileStore} from 'user-profile/reducers';
```
petrjasek, line 28: seems like same thing imported 2x

petrjasek (approved):

assets/company-admin/components/CompanyAdminApp.tsx
```diff
@@ -69,7 +69,7 @@ interface IDispatchProps {
     setProductFilter: (productId: any) => any;
 }
 
-interface IOwnProps {
+type IOwnProps = {
```
petrjasek, line 72: why defining it when it's empty?

thecalcc: because I want to see what type I pass to the `connect` function. for the most part documentation

petrjasek: I see, didn't notice you use it there, if you would just pass there `{}` wouldn't it be enough?

thecalcc: It would, but I wanted to have a clear link between connect and the component props. And in the future if we extend this component and decide to pass some props it will be clear and easier to do it.

thecalcc: Technically it would but then it's not obvious what that parameter means. This way we have a direct link between props and that type parameter.

PR #477, opened 2023-08-15
fix server error handling in user form

petrjasek:
CPCN-253

### DIFF
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -10,6 +10,7 @@ import {
     getTimezoneOffset,
     recordAction,
     copyTextToClipboard,
+    errorHandler,
 } from 'utils';
 import {noNavigationSelected, getNavigationUrlParam} from 'search/utils';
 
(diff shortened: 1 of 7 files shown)


assets/users/actions.ts
```diff
@@ -157,8 +157,9 @@ export function postUser() {
                     notify.success(gettext('User created successfully'));
                 }
                 dispatch(fetchUsers());
-            })
-            .catch((error: any) => errorHandler(error, dispatch, setError));
+            }, (error) => {
+                dispatch(setError(error));
```
MarkLark86, line 161: Rather than handling this case different than other errorHandlers, what if we were to improve the `errorHandler` function to check if the `error` is a request response or not, then dispatch `setError`.

Something like:
```typescript
(error.response?.json == null ?
    Promise.resolve(error) :
    error.response.json()
).then((data) => {
    dispatch(setError(data));
})
```

And continue to use `.catch` here rather than the rejected promise callback?

petrjasek: tried to make it a bit more explicit

MarkLark86 (approved):
I don't know why, but we also have this `errorHandler` defined in [wire/actions.ts](https://github.com/superdesk/newsroom-core/blob/develop/assets/wire/actions.ts#L392) and [agenda/actions.ts](https://github.com/superdesk/newsroom-core/blob/develop/assets/agenda/actions.ts#L703) as well

MarkLark86 (approved):

PR #480, opened 2023-08-17
RadioButtonGroup size prop

dzonidoo:


petrjasek (approved):

assets/features/sections/SectionSwitch.tsx
```diff
@@ -24,21 +25,51 @@ interface IProps {
...
+                        'toggle-button--active': section._id === activeOptionId,
+                    });
+
+                    return (
+                        <button
+                            key={section._id}
+                            className={buttonClasses}
+                            onClick={() => switchOptions(section._id)}
+                        >
+                            {gettext(section.name)}
```
tomaskikutis, line 69: @petrjasek we should probably get rid of those gettext calls where variables and not inline strings are being passed?

PR #481, opened 2023-08-18
Add Coverage Not Decided to Coverage Status filter [STTNHUB-198] [STTNHUB-203]

devketanpro:
https://github.com/superdesk/superdesk-stt/pull/40
https://github.com/superdesk/superdesk-planning/pull/1844

### DIFF
--- a/assets/agenda/components/AgendaCoverageExistsFilter.tsx
+++ b/assets/agenda/components/AgendaCoverageExistsFilter.tsx
@@ -27,7 +27,7 @@ function getActiveFilterLabel(filter: any, activeFilter: any) {
     case FILTER_VALUES.NOT_PLANNED:
         return gettext('Not Planned');
     case FILTER_VALUES.MAY_BE:
-        return gettext('May Be');
+        return gettext('Not Decided');
     }
 
     return filter.label;
@@ -54,7 +54,7 @@ function AgendaCoverageExistsFilter ({toggleFilter, activeFilter}: any) {
             >{gettext('Coverage not planned')}
             </button>
             <button
-                key='coverage-not-planned'
+                key='coverage-not-decided'
                 className='dropdown-item'
                 onClick={() => toggleFilter(filter.field, FILTER_VALUES.MAY_BE)}
             >{gettext('Coverage not decided')}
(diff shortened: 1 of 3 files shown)


newsroom/agenda/agenda.py
```diff
@@ -547,11 +547,23 @@ def _filter_terms(filters, item_type):
...
+                must_term_filters.append(
+                    nested_query(
+                        path="coverages",
+                        query={
+                            "bool": {"filter": [{"terms": {"coverages.coverage_status": ["coverage not decided yet"]}}]}
+                        },
+                        name="coverage_status",
+                    )
+                )
             else:
```
petrjasek, line 560: would be good to check the value here explicitly

assets/agenda/components/AgendaCoverageExistsFilter.tsx
```diff
@@ -50,6 +53,12 @@ function AgendaCoverageExistsFilter ({toggleFilter, activeFilter}: any) {
                 onClick={() => toggleFilter(filter.field, FILTER_VALUES.NOT_PLANNED)}
             >{gettext('Coverage not planned')}
             </button>
+            <button
+                key='coverage-not-planned'
```
MarkLark86, line 57: This key should be unique

assets/agenda/components/AgendaCoverageExistsFilter.tsx
```diff
@@ -25,6 +26,8 @@ function getActiveFilterLabel(filter: any, activeFilter: any) {
         return gettext('Planned');
     case FILTER_VALUES.NOT_PLANNED:
         return gettext('Not Planned');
+    case FILTER_VALUES.MAY_BE:
+        return gettext('May Be');
```
MarkLark86, line 30: This is the text that appears in the parent button of the dropdown, which has a different label to the option provided in the dropdown. 
I'm wondering if this should rather be `Not Decided` so the option in the dropdown list and the dropdown button text are relatable for end users

assets/agenda/components/AgendaCoverageExistsFilter.tsx
devketanpro: my bad, I forgot this

assets/agenda/components/AgendaCoverageExistsFilter.tsx
devketanpro: yep, sounds identical

MarkLark86 (approved):

PR #482, opened 2023-08-18
make topic details non editable in preview

petrjasek:
CPCN-260


petrjasek:
the feature itself is around this `readonly` param handling, but also switched the components to typescript which made it a bit bigger

MarkLark86 (approved):

PR #484, opened 2023-08-21
Improve modal style, data, texts, more news button and more

thecalcc:
CPCN-292

### DIFF
--- a/assets/components/cards/render/MoreNewsButton.tsx
+++ b/assets/components/cards/render/MoreNewsButton.tsx
@@ -3,19 +3,25 @@ import PropTypes from 'prop-types';
 import {gettext} from 'utils';
 
 function MoreNewsButton({title, productId, photoUrl, photoUrlLabel, moreNews}: any): any {
-    return ([<div key='heading' className='col-6 col-sm-8'>
-        <h3 className='home-section-heading'>{title}</h3>
-    </div>,
-    <div key='more-news' className='col-6 col-sm-4 d-flex align-items-start justify-content-end'>
-        {moreNews && productId &&
-                <a href={`/wire?product=${productId}`} role='button' className='nh-button nh-button--tertiary mb-3'>
-                    {gettext('More news')}
-                </a>}
-        {photoUrl &&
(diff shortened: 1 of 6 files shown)


assets/components/cards/render/MoreNewsButton.tsx
```diff
@@ -3,19 +3,25 @@ import PropTypes from 'prop-types';
...
+            </div>,
+            <div key='more-news' className='col-6 col-sm-4 d-flex align-items-start justify-content-end'>
+                {moreNews && productId && (
+                    <a href={`/wire?product=${productId}`} role='button' className='nh-button nh-button--tertiary mb-3'>
+                        {gettext('More news')}
+                    </a>
+                )}
+                {photoUrl && (
+                    <a href={photoUrl} target='_blank' rel='noopener noreferrer' role='button' className='nh-button nh-button--tertiary mb-3'>
+                        {gettext(photoUrlLabel)}
```
tomaskikutis, line 19: that was here before, but I'd be good to fix if you're touching the component - the prop should come translated already - we can only pass inline strings to gettext

tomaskikutis (approved):

assets/components/cards/render/MoreNewsButton.tsx
thecalcc: Can't be done since the labels aren't static but they're coming from variables. For now I'd say let's leave it as or just remove the `gettext` since we only pass text from variables.

tomaskikutis: @petrjasek it'd be good to get rid of variables being passed to gettext. In this case it looks like `photoUrlLabel` is coming from the dabatabase - should we simply not translate it? Or is there some hack in newshub as we have in superdesk scripts/core/lang/missing-translations-strings.ts ?

petrjasek: yep I think we can get rid of it, doesn't make much sense

tomaskikutis (approved):

petrjasek:
seems like there a few e2e tests failing after this, will revert it. pls check those first before merging @thecalcc

PR #487, opened 2023-08-22
[NHUB-362] and [NHUB-364]

fritzSF:
- New icons for additional coverage statuses
- New logo for the default Newshub theme


newsroom/static/newshub_logo.svg
petrjasek, line 1: we're importing `logo.svg` in the `logo.html` page, it needs to be updated there (or the new file must be renamed)

petrjasek (approved):

PR #488, opened 2023-08-22
change delete folder message

dzonidoo:
CPCN-313


petrjasek (approved):

petrjasek (changes requested):
actually that `gettext` doesn't work well with those backtick quotes, pls use standard `'`

petrjasek:
almost there, if you switch to single quotes it will be perfect :wink:

petrjasek (approved):

PR #489, opened 2023-08-22
show events starting before selected day

petrjasek:
NHUB-353


petrjasek:
@MarkLark86 it's configurable now on the backend, those UI fixes are needed anyhow

MarkLark86 (approved):

PR #505, opened 2023-08-28
Fix more news button in personal dashboard

thecalcc:


assets/components/cards/render/MoreNewsButton.tsx
```diff
@@ -1,19 +1,44 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import {gettext} from 'utils';
 
-function MoreNewsButton({title, productId, photoUrl, photoUrlLabel, moreNews}: any): any {
+interface IProps {
+    title: string;
+    photoUrl?: string;
+    kind: MoreNewsSearchKind;
```
petrjasek, line 7: just wondering what's the benefit of using this over `kind: "product" | "topic"`

thecalcc: Just a different pattern. Main benefit is that we can extend it in the future without having any ternary conditions in the JSX. It's way shorter this way.

tomaskikutis: Aren't enums useless here? same works with union types [image](https://github.com/superdesk/newsroom-core/assets/2076953/1c33238d-ba01-4a21-8cae-3b7f403b7f47)

thecalcc: Not fully useless, you can access them via <enum>.<option> which is nice. The other thing is that kind is also typed so autofill works and also you'll get an error if you pass an invalid type. Just a preference thing I guess. Let me know if you want me to remove it.

tomaskikutis: As you say, editor autocomplete works with union types as well so I'd say it's _more efficient_ to stick with union types and avoid the need for extra imports. Yes - do update to union types.

assets/components/cards/utils.ts
```diff
@@ -15,9 +15,80 @@ import LargePictureTextCard from 'components/cards/render/LargePictureTextCard';
...
+
+interface ICard1 {
+    _id: '2x2-events';
+    text: string;
+    editComponent: ComponentType<any>;
+    size: number;
+    dashboardComponent: ComponentType<IProps2x2>;
+}
+
+interface ICard2 {
```
petrjasek, line 58: here it would be also better to use something more descriptive than `ICard1` `ICard2` etc

petrjasek (approved):

PR #508, opened 2023-08-29
[CPNHUB-248] Support configuring a Company authentication provider

MarkLark86:


newsroom/types.py
```diff
@@ -59,12 +60,33 @@ class User(UserData):
...
+    is_approved: bool
+
+
+class AuthProviderType(Enum):
+    PASSWORD = "password"
+    GOOGLE_OAUTH = "google_oauth"
+    SAML = "saml"
+
+
+class AuthProvider(TypedDict):
```
petrjasek, line 77: I think it might be useful to have some notion of what does provider support and based on that we could do the checks in the auth workflow, so instead of:
```py
if auth_provider["auth_type"] == AuthProviderType.PASSWORD.value:
    send_token(...)
```
there would be something like:
```
if auth_provider.features.verify_email:
    send_token(...)
```

petrjasek (approved):

petrjasek (approved):

PR #514, opened 2023-08-30
fix missing remove btn in advanced search bar

petrjasek:
avoided duplicated code, there was a bad param in one of the 3 sections

CPCN-318

### DIFF
--- a/assets/search/components/SearchResultsBar/SearchResultsAdvancedSearchRow.tsx
+++ b/assets/search/components/SearchResultsBar/SearchResultsAdvancedSearchRow.tsx
@@ -52,6 +52,19 @@ export function SearchResultsAdvancedSearchRow({
         any: splitTermKeywords(advancedSearchParams.any),
         exclude: splitTermKeywords(advancedSearchParams.exclude),
     };
+
+    const labels = {
+        all: gettext('and'),
+        any: gettext('or'),
+        exclude: gettext('not'),
+    };
+
+    const shades = {
+        all: 'success',
+        any: 'info',
+        exclude: 'alert',
+    };
+
     const fields = get(advancedSearchParams, 'fields') ?? [];
 
     if (!keywords.all.length && !keywords.any.length && !keywords.exclude.length) {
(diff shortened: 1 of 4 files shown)


assets/search/components/SearchResultsBar/SearchResultsAdvancedSearchRow.tsx
```diff
@@ -70,103 +83,38 @@ export function SearchResultsAdvancedSearchRow({
...
-                readOnly={true}
-            />
-        );
-
-        keywords.any.forEach((term, index) => {
+    let key: keyof typeof keywords;
+    for (key in keywords) {
+        // due to the declaration outside key would be the last key
+        // at the end but we need the current value for the onClick callback
+        const field = key;
```
petrjasek, line 90: not exactly happy about this @tomaskikutis @thecalcc are you aware of any better way?

tomaskikutis: You're missing `const`

```
for (const key in keywords) {
     ^
```

With `const` it keeps the context and `onChange` would access correct values.

image

The problem then becomes typescript inferring key as string and not union type. one option is to overwrite types for global `Object.keys` by putting the following to `globals.d.ts`

```
interface ObjectConstructor {
    keys<T extends {}>(o: T): Array<keyof T>;
}
```

In case it causes stuff not to compile somewhere else, we could add a custom object keys implementation that's more strict:

```
function keysStrict<T extends {}>(obj: T): Array<keyof T> {
    return Object.keys(obj) as Array<keyof T>;
}

for (const field of keysStrict(keywords)) {
    ...
}
```

Third option, to use unmodified `Object.keys` and cast the result:
```
for (const field of Object.keys(keywords) as Array<keyof typeof keywords>) {
    ...
}
```

petrjasek: went for the 3rd after all, that seems to be the most straightforward option

assets/search/components/SearchResultsBar/SearchResultsAdvancedSearchRow.tsx
```diff
@@ -70,103 +83,36 @@ export function SearchResultsAdvancedSearchRow({
...
-                text={gettext('or')}
-                operator={true}
-                shade="info"
-                readOnly={true}
-            />
-        );
-
-        keywords.any.forEach((term, index) => {
+    for (const field of Object.keys(keywords) as Array<keyof typeof keywords>) {
+        // due to the declaration outside key would be the last key
```
tomaskikutis, line 87: since you fixed the problem with scope, the comment is no longer correct and should be removed

PR #518, opened 2023-08-31
[NHUB-370] improve: Enable user emails when subscribing to a topic

MarkLark86:


newsroom/topics/topics.py
```diff
@@ -94,6 +99,33 @@ def on_update(self, updates, original):
...
+        self.auto_enable_user_emails(updates, original)
+
+    def auto_enable_user_emails(self, updates, original):
+        if not updates.get("subscribers"):
+            return
+
+        users = get_user_dict()
+        users_service = superdesk.get_resource_service("users")
+        original_subscribers = [subscriber["user_id"] for subscriber in original.get("subscribers") or []]
+        for subscriber in updates["subscribers"]:
```
petrjasek, line 113: imo would be enough to check the current user only, not sure if there would be a usecase to check any other. and while this might be the case anyhow due to checking the original subscribers it might be good to make it explicit so it's more obvious

MarkLark86: If the user has the privilege, they can also enable the subscription for other users, which in this case would also turn it on for them as well

petrjasek: I see, not sure that would be expected if I explicitly stop those to get them later based on some other user action

MarkLark86: Sounds reasonable, 1 users action should not affect another users preferences.

I'll make the change tomorrow

petrjasek (approved):

PR #519, opened 2023-08-31
Initially open some of the dropdowns

thecalcc:
CPCN-319

### DIFF
--- a/assets/components/FormSection.tsx
+++ b/assets/components/FormSection.tsx
@@ -1,8 +1,14 @@
 import React, {useState} from 'react';
-import PropTypes from 'prop-types';
 
-export function FormSection({name, testId, children}) {
-    const [opened, setOpened] = useState(name == null);
+interface IProps {
+    initiallyOpen?: boolean;
+    name?: string;
+    testId: string;
+    children: JSX.Element;
+}
+
+export function FormSection({initiallyOpen, name, testId, children}: IProps) {
+    const [opened, setOpened] = useState(initiallyOpen ?? name == null);
 
     return (
         <div className="nh-flex__column" data-test-id={testId}>
@@ -26,9 +32,3 @@ export function FormSection({name, testId, children}) {
         </div>
     );
 }
-
-FormSection.propTypes = {
-    name: PropTypes.string.isRequired,
-    testId: PropTypes.string,
-    children: PropTypes.node,
-};
--- a/assets/components/FormSection.tsx
+++ b/assets/components/FormSection.tsx
@@ -2,16 +2,16 @@ import React, {useState} from 'react';
 
 interface IProps {
     initiallyOpen?: boolean;
-    name?: string;
-    testId: string;
+    name: string;
+    dataTestId?: string;
     children: JSX.Element;
 }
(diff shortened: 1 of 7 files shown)


petrjasek (approved):

assets/components/FormSection.tsx
```diff
@@ -1,8 +1,14 @@
...
-    const [opened, setOpened] = useState(name == null);
+interface IProps {
+    initiallyOpen?: boolean;
+    name?: string;
+    testId: string;
+    children: JSX.Element;
+}
+
+export function FormSection({initiallyOpen, name, testId, children}: IProps) {
+    const [opened, setOpened] = useState(initiallyOpen ?? name == null);
```
petrjasek, line 11: maybe this should be actually `true` by default, seem to me like it's always the case now

thecalcc: For now yes, all the usages are in places where it has to be open initially, I just thought I'd future proof a little bit by sticking with the common pattern.

petrjasek (approved):
seems like you will have to fix some e2e tests, afaik they were expecting those sections to be closed so there are some extra clicks to toggle it open

assets/components/FormSection.tsx
```diff
@@ -1,8 +1,14 @@
 import React, {useState} from 'react';
-import PropTypes from 'prop-types';
 
-export function FormSection({name, testId, children}) {
-    const [opened, setOpened] = useState(name == null);
+interface IProps {
+    initiallyOpen?: boolean;
+    name?: string;
```
tomaskikutis, line 5: name was marked as required in proptypes

assets/components/FormSection.tsx
```diff
@@ -1,8 +1,14 @@
...
+    testId: string;
```
tomaskikutis, line 6:
```
interface Test {
    'data-test-id'?: string;
}
```

Keep the same name for test IDs

assets/components/FormSection.tsx
tomaskikutis, line 11: since name was marked in mandatory, probably it doesn't make sense to have a condition where you check if it's `null`. Unless proptypes were wrong.

assets/search/components/SearchResultsBar/SearchResultsTopicRow.tsx
```diff
@@ -58,7 +58,7 @@ export function SearchResultsTopicRow({
         );
     }
 
-    if (get(searchParams, 'navigation.length', 0)) {
+    if ((searchParams.navigation?.length ?? 0) > 0 && ((Object.keys(navigations).length ?? 0) > 0)) {
```
tomaskikutis, line 61: `Object.keys` returns an empty array if there are no keys, thus defaulting `length` to zero will never execute

assets/components/FormSection.tsx
thecalcc: This name isn't convenient as you have to destructure props on another line in order to access it with `props['data-test-id']`

tomaskikutis (approved):

PR #528, opened 2023-09-05
If the client doesn't have wire configured don't show dashboard personalization option

thecalcc:
CPCN-323


assets/home/components/HomeApp.tsx
```diff
@@ -230,6 +231,7 @@ class HomeApp extends React.Component<IProps, IState> {
 
     renderContent(children?: any): any {
         const {cards} = this.props;
+        const isWireSectionConfigured = Object.keys(this.props.userSections).find((key) => key === 'wire');
```
tomaskikutis, line 234: Why iterate when you can check it directly? `this.props.userSections['wire'] != null`

Can you use a constant for `wire` string?

tomaskikutis (approved):

PR #529, opened 2023-09-05
User schedule notifications improvement

dzonidoo:
CPCN-297

### DIFF
--- a/assets/user-profile/components/EditNotificationScheduleModal.tsx
+++ b/assets/user-profile/components/EditNotificationScheduleModal.tsx
@@ -10,6 +10,12 @@ import {updateUserNotificationSchedules} from 'user-profile/actions';
 import Modal from 'components/Modal';
 import {TimezoneInput} from 'components/TimezoneInput';
 
+const NOTIFICATION_TIME_OBJECT: React.InputHTMLAttributes<HTMLInputElement> = {
+    type: 'time',
+    min: '00:00',
+    step: '900',
+};
+
 interface IProps {
     modalFormInvalid(): void;
     modalFormValid(): void;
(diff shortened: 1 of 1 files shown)


petrjasek:
imo it would be better to use rather selects, 1 for hours and second for minutes (with just `00, 15, 30, 45` and `00` being the default value), atm it would allow you to select anything and then on save it complains but might be browser specific actually

tomaskikutis:
I did think about that, but decided to suggest the time picker. One reason is out of the box 12/24 hour support and another is that it's less changes since the settings aren't going to be changed that often anyway. We could add a info message to specify the limitation on minutes. We could also reimplement to not use time picker if you think it's worth it.

petrjasek:
ok we can run it by qa and we will see

assets/user-profile/components/EditNotificationScheduleModal.tsx
```diff
@@ -49,13 +53,36 @@ class EditNotificationScheduleModalComponent extends React.Component<IProps, ISt
         });
     }
 
+    onSubmitForm(event: React.FormEvent<HTMLFormElement>) {
+        event.stopPropagation();
+        event.preventDefault();
+
+        if (this.formRef.current == null) {
+            return;
```
tomaskikutis, line 61: it's better to use `throw new Error('ref missing')` so it's easier to distinguish code that is not supposed to run at all

assets/user-profile/components/EditNotificationScheduleModal.tsx
```diff
@@ -69,28 +96,37 @@ class EditNotificationScheduleModalComponent extends React.Component<IProps, ISt
...
+                                step="900"
+                                min="00:00"
                             />
                             <input
                                 type="time"
                                 value={this.state.times[1]}
                                 onChange={(event) => {
                                     this.updateTime(event.target.value, 1);
                                 }}
+                                step="900"
```
tomaskikutis, line 119: move those props into a variable and put at the top of the file

petrjasek (approved):

PR #531, opened 2023-09-07
Fix : Remove 'Contact Us' link on Login header [CPCN-287]

devketanpro:


petrjasek (changes requested):
we need to remove it only for CP but not in general, so would need some `if` there. we can do it probably based on that `config.CONTACT_ADDRESS`

devketanpro:
we need to add `CONTACT_ADDRESS` in the settings, because as I checked it returns `#`

petrjasek:
yep we can set it there to be `None`

petrjasek (approved):

PR #546, opened 2023-09-12
fix user profile notifications not updated on new topic

petrjasek:
reload the page when new topic is saved so the changes done to user profile are also updated.

NHUB-370


assets/search/utils.ts
```diff
@@ -53,3 +54,7 @@ export function getAdvancedSearchFields(context: any) {
 
     return (config.fields || {})[context];
 }
+
+export function getTopicUrl(topic: ITopic) {
+    return `/${topic.topic_type}?topic=${topic._id}`;
+}
```
devketanpro, line 60: end line is missing here

devketanpro (approved):

PR #553, opened 2023-09-14
Fix : Highlighting in the list view is not accurate when using exact phrase search [CPCN-335]

devketanpro:


assets/wire/utils.ts
```diff
@@ -214,10 +214,11 @@ function splitWords(str: string): Array<string> {
         .replace(/ {2,}/g, ' ')
 
         .trim()
-        .split(' ');
+        .split(/(?= [A-Z])/);
```
petrjasek, line 217: can you add a comment what is it supposed to do? and not sure if this will work with french eg. `ÉQUILIBRE`

assets/wire/utils.ts
```diff
@@ -214,10 +214,11 @@ function splitWords(str: string): Array<string> {
...
 }
 
 /**
+ * /(?= [A-Z])/
```
petrjasek, line 221: not sure what does it mean here?

devketanpro: my bad I forgot to remove it

petrjasek (commented):
in general it would be good to show rather more highlights than text, so might make sense to return rather multiple snippets with matches than longer sentence upfront.
would be enough to add a few words before and after each highlight (up to 5 maybe), we don't need full sentences there.

assets/wire/tests/WireApp.spec.tsx
```diff
@@ -122,7 +123,7 @@ describe('shortHighlightedtext', () => {
         const maxLength = 40;
         const output = shortHighlightedtext(html, maxLength);
         expect(output).toEqual(`On olemassa myös sellaisia ihmisiä, joiden silmissä
-        jokainen parisuhde muuttuu <span class="es-highlight">mukaanalkuhuuma</span>  <span class="es-highlight">kestää</span> yleensä 1–2 vuotta...`);
+        jokainen parisuhde muuttuu <span class="es-highlight">mukaanalkuhuuma</span> <span class="es-highlight">kestää</span> yleensä 1–2 vuotta...`);
     });
```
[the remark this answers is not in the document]
petrjasek: also a case like:
```html
<p>some <span class="es-highlight">matching</span> text and <span class="es-highlight">more</span> matching text.</p>
<p>another <span class="es-highlight">matching</span> text.</p>
```
should produce pretty much the same output without those `p` tags

assets/wire/tests/WireApp.spec.tsx
```diff
@@ -125,6 +125,20 @@ describe('shortHighlightedtext', () => {
         expect(output).toEqual(`On olemassa myös sellaisia ihmisiä, joiden silmissä
         jokainen parisuhde muuttuu <span class="es-highlight">mukaanalkuhuuma</span> <span class="es-highlight">kestää</span> yleensä 1–2 vuotta...`);
     });
-       
+
+    it('returns highlight text condition with more tags', () => {
+        const html = '<ul><li><span class="es-highlight">foo</span></li><li><span class="es-highlight">foo</span></li><li>bar</li><li><span class="es-highlight">foo</span></li></ul>';
+        const maxLength = 40;
+        const output = shortHighlightedtext(html, maxLength);
+        expect(output).toEqual('<span class="es-highlight">foo</span>...');
```
petrjasek, line 133: would be good to actually output here:
```html
<span class="es-highlight">foo</span> <span class="es-highlight">foo</span> bar <span class="es-highlight">foo</span>
```
just not sure if we should have line breaks there in between or just spaces

petrjasek: spaces would be probably better or some `...`

assets/wire/utils.ts
```diff
@@ -230,52 +256,39 @@ function splitWords(str: string): Array<string> {
...
-        if (words.length <= remainingCount) {
-            text += content + ' ';
-            count += words.length;
-        } else {
-            text += words.slice(0, remainingCount).join(' ');
-            count += remainingCount;
-        }
+    const div = document.createElement('div');
+    div.innerHTML = html;
+    const highlightSpan = div.querySelector('span.es-highlight');
```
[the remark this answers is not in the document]
devketanpro: we need to develop logic based on the initial span, and if we find the first span, then display the text together with the previous and next siblings, and if there are any more highlighted span in this paragraph, then they will also display.

assets/wire/tests/WireApp.spec.tsx
devketanpro: I see, however I suppose if we handle that situation, it will break in a lengthy paragraph?

petrjasek: yep but you will see more matches which is good, while it's not that long still. showing just the first match is not enough imo

devketanpro: I tried to handle that case, but seems liken other cases are failed

assets/wire/utils.ts
```diff
@@ -258,34 +258,34 @@ function getTextFromNode(node: Node, count: number): { text: string; remainingCo
 export function shortHighlightedtext(html: string, maxLength = 40) {
     const div = document.createElement('div');
     div.innerHTML = html;
-    const highlightSpan = div.querySelector('span.es-highlight');
-    if (!highlightSpan) {
+    const highlightSpans = Array.from(div.querySelectorAll('span.es-highlight'));
+    if (highlightSpans.length === 0) {
         return html;
```
petrjasek, line 263: not sure if this should be the return value because it might be too long

assets/wire/utils.ts
```diff
@@ -258,34 +258,34 @@ function getTextFromNode(node: Node, count: number): { text: string; remainingCo
...
-    let currentNode: Node | null = highlightSpan.previousSibling;
-    while (currentNode) {
-        const {text, remainingCount} = getTextFromNode(currentNode, maxLength);
-        extractedText = text + extractedText;
-        maxLength = remainingCount;
-        currentNode = currentNode.previousSibling;
-    }
+    for (const highlightSpan of highlightSpans) {
+        let currentNode: Node | null = highlightSpan.previousSibling;
+        while (currentNode && currentNode.nodeType === Node.TEXT_NODE) {
```
petrjasek, line 270: when it would need iterating for more than 1 previous node?

assets/wire/utils.ts
```diff
@@ -258,34 +258,34 @@ function getTextFromNode(node: Node, count: number): { text: string; remainingCo
...
+            const {text, remainingCount} = getTextFromNode(currentNode, maxLength);
+            extractedText += text;
+            maxLength = remainingCount;
+            currentNode = currentNode.previousSibling;
+        }
+
+        const tempDiv = document.createElement('div');
+        tempDiv.appendChild(highlightSpan.cloneNode(true));
+        extractedText += tempDiv.innerHTML;
```
petrjasek, line 279: why not use just `extractedText += highlightSpan.outherHTML;` ?

devketanpro: correct , we can use this as well

assets/wire/utils.ts
devketanpro: correct, the `while` loop checking for previous siblings of `highlightSpan` (previous nodes of type `Node.TEXT_NODE`) may not be necessary and can be simplified since I am just thinking in extracting the text before the `highlightSpan`.

assets/wire/tests/WireApp.spec.tsx
```diff
@@ -124,6 +125,20 @@ describe('shortHighlightedtext', () => {
...
+        const output = shortHighlightedtext(html, maxLength);
+        expect(output).toEqual('<span class="es-highlight">foo</span> <span class="es-highlight">foo</span> <span class="es-highlight">foo</span>...');
+    });
+
+    it('returns highlight text condition with more matching words', () => {
+        const html = `<p>some <span class="es-highlight">matching</span> text and <span class="es-highlight">more</span> matching text.</p>
+        <p>another <span class="es-highlight">matching</span> text.</p>`;
+        const maxLength = 40;
+        const output = shortHighlightedtext(html, maxLength);
+        expect(output).toEqual('some <span class="es-highlight">matching</span> text and  text and <span class="es-highlight">more</span> matching text.another <span class="es-highlight">matching</span> text....');
```
petrjasek, line 141: here we should make sure we're not repeating the `text and` that should be there only once. and would be good to add space after `text.`

assets/wire/utils.ts
```diff
@@ -259,28 +245,27 @@ export function shortHighlightedtext(html: string, maxLength = 40) {
...
-        extractedText += highlightSpan.outerHTML;
+        extractedText.push(highlightSpan.outerHTML);
 
-        const nextNode = highlightSpan.nextSibling;
-        if (nextNode && nextNode.nodeType === Node.TEXT_NODE) {
-            const {text, remainingCount} = getTextFromNode(nextNode, maxLength);
-            extractedText += text;
-            maxLength = remainingCount;
+        // Append text after the highlight span
+        if (textAfter && textAfter !== lastWord) {
```
petrjasek, line 263: not sure if we should compare it to `lastWord` here because that would probably skip it in eg.:
```html
<p>foo <span class="es...">bar</span> foo <span class="es...">bar</span> foo</p>
```
would become:
```html
<p>foo <span ...>bar</span> <span ...>bar</span></p>
```
but maybe not? might be something for a test case
in general you want to avoid using `textBefore` from node if that very same node was `nextSibling`
so rather than checking for last word it might be better to store last node

devketanpro: sounds identical

petrjasek (approved):

PR #554, opened 2023-09-14
break lines description

dzonidoo:
STTNHUB-254

### DIFF
--- a/assets/agenda/components/AgendaListItem.tsx
+++ b/assets/agenda/components/AgendaListItem.tsx
@@ -129,8 +129,10 @@ class AgendaListItem extends React.Component<any, any> {
     renderListItem(isMobile: any, children: any) {
         const {item, isExtended, group, planningId, listConfig} = this.props;
         const classes = this.getClassNames(isExtended);
-        const planningItem = (get(item, 'planning_items') || []).find((p: any) => p.guid === planningId) || {};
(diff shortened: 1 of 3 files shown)


tomaskikutis (commented):
Do not call arguments or variables `element` if it's not a HTML element.

`item` is a generic term if you can't come up with anything better. In your case, you are splitting a description that is made of many lines of text. A good name could be `lineOfPlainText` or `lineOfHTML`

assets/agenda/components/AgendaListItem.tsx
```diff
@@ -182,11 +184,20 @@ class AgendaListItem extends React.Component<any, any> {
...
-                                    dangerouslySetInnerHTML={({__html: description})}
-                                /> : <PlainText text={description} />}
-                            </p>
+                            <div className="wire-articles__item__text">
+                                {item.es_highlight && item.es_highlight
+                                    ? (
+                                        description.split('\n').map((lineOfPlainTex: string, index: number) => {
+                                            return (
+                                                <p key={index}>
+                                                    <span dangerouslySetInnerHTML={{__html: lineOfPlainTex}} />
```
tomaskikutis, line 193: why would you do `dangerouslySetInnerHTML` for a line of plain text? :)

tomaskikutis (approved):

PR #567, opened 2023-09-21
Toggle dropdown on item type change in Agenda

thecalcc:
NHUB-407


assets/agenda/components/AgendaApp.tsx
```diff
@@ -383,6 +384,7 @@ const mapStateToProps = (state: any) => ({
     modal: state.modal,
     user: state.user,
     company: state.company,
+    filtersConfig: agendaFiltersConfigSelector(state),
```
petrjasek, line 387: I don't see it being used

tomaskikutis (approved):

PR #568, opened 2023-09-21
Fix : Sorting and Product filter do not have any effect [NHUB-400]

devketanpro:


newsroom/users/views.py
```diff
@@ -79,13 +79,24 @@ def user_profile():
...
         lookup = {"$or": [{"first_name": regex}, {"last_name": regex}]}
 
     if flask.request.args.get("ids"):
         lookup = {"_id": {"$in": (flask.request.args.get("ids") or "").split(",")}}
 
+    if flask.request.args.get("sort"):
+        sort = flask.request.args.get("sort")
+
+    if flask.request.args.get("where"):
+        where = json.loads(flask.request.args.get("where"))
```
petrjasek, line 94: would be good to catch errors

petrjasek (commented):
pls add a test for sorting/filtering

newsroom/users/views.py
```diff
@@ -80,22 +80,26 @@ def user_profile():
...
+            sort = flask.request.args.get("sort")
+
+        where_param = flask.request.args.get("where")
+        if where_param:
+            where = json.loads(where_param)
+            if where.get("company"):
+                lookup = {"company": where["company"]}
+            if where.get("products._id"):
+                lookup = {"products._id": where["products._id"]}
+    except json.JSONDecodeError as e:
```
petrjasek, line 101: would actually put it that `try/catch` only around `json.loads` we don't need it for the whole thing

petrjasek (approved):

PR #575, opened 2023-09-25
How to present event description in list view

dzonidoo:
STTNHUB-254

### DIFF
--- a/assets/agenda/components/AgendaListItem.tsx
+++ b/assets/agenda/components/AgendaListItem.tsx
@@ -136,6 +136,15 @@ class AgendaListItem extends React.Component<any, any> {
         // Show headline for adhoc planning items
         const showHeadline = !item.event && get(item, 'headline.length', 0) > 0;
 
+        const descriptionHTML = description.replace('\n', '<br/>').replace('<br/>', '</p>\n<p>');
+
+        function isHTML(value: string) {
+            const doc = new DOMParser().parseFromString(value, 'text/html');
+            return Array.from(doc.body.childNodes).some(node => node.nodeType === 1);
+        }
+
+        const descriptionArray = isHTML(description) ? descriptionHTML : description;    
+
         return (
             <article key={item._id}
                 className={classes.card}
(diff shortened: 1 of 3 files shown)


assets/agenda/components/AgendaListItem.tsx
```diff
@@ -136,6 +136,15 @@ class AgendaListItem extends React.Component<any, any> {
         // Show headline for adhoc planning items
         const showHeadline = !item.event && get(item, 'headline.length', 0) > 0;
 
+        const descriptionHTML = description.replace('\n', '<br/>').replace('<br/>', '</p>\n<p>');
```
petrjasek, line 139: not sure this would generate valid html in general, why do we need this?
eg. from:
```html
<ul>
  <li>foo</li>
</ul>
```
you will get:
```html
<ul></p>\n<p><li>foo</li></p>\n<p></ul></p>\n<p>
```

assets/agenda/components/AgendaListItem.tsx
```diff
@@ -185,17 +194,27 @@ class AgendaListItem extends React.Component<any, any> {
 
                         {(isMobile || isExtended) && description && (
                             <div className="wire-articles__item__text">
-                                {item.es_highlight && item.es_highlight
+                                {item.es_highlight && item.es_highlight.definition_short
```
petrjasek, line 197: would be better to just remove that extra condition, there might not be highlight for `definition_short` but then it would use the other one

assets/agenda/components/AgendaListItem.tsx
```diff
@@ -185,17 +194,27 @@ class AgendaListItem extends React.Component<any, any> {
...
-                                            );
+                                        descriptionArray.split('\n').map((lineOfHTML: string, index: number) => (
+                                            <p key={index}>
+                                                <span dangerouslySetInnerHTML={{__html: lineOfHTML}} />
+                                            </p>
+                                        ))
+                                    )
+                                    : (
+                                        descriptionArray.split('\n').slice(0, 3).map((plainOfHTML: string, i: number, array: any) => {
+                                            const lastChild: boolean = array.length -1 === i && descriptionArray.length > 3;
```
petrjasek, line 207: wouldn't it be easier to use `:last-child` in css?

assets/agenda/components/AgendaListItem.tsx
```diff
@@ -185,17 +194,27 @@ class AgendaListItem extends React.Component<any, any> {
...
+
+                                            return isHTML(description)
```
petrjasek, line 209: not sure it makes sense to call this for each line to get the same value, given that it does parse html it might be time consuming. would be better to store some value upfront

assets/agenda/components/AgendaListItem.tsx
```diff
@@ -136,6 +136,15 @@ class AgendaListItem extends React.Component<any, any> {
...
         const showHeadline = !item.event && get(item, 'headline.length', 0) > 0;
 
+        const descriptionHTML = description.replace('\n', '<br/>').replace('<br/>', '</p>\n<p>');
+
+        function isHTML(value: string) {
+            const doc = new DOMParser().parseFromString(value, 'text/html');
+            return Array.from(doc.body.childNodes).some(node => node.nodeType === 1);
+        }
+
+        const descriptionArray = isHTML(description) ? descriptionHTML : description;
```
petrjasek, line 146: not sure about the naming there, because neither `descriptionHTML` nor `description` is an array

assets/agenda/components/AgendaListItem.tsx
dzonidoo: not really... for example if you only have two rows of description then you don't need to show three dot on the second one (lastchild in this case)...

assets/agenda/components/AgendaListItem.tsx
dzonidoo: ...and for other one dont need to show full description list... bcz of that i need this extra condition

assets/agenda/components/AgendaListItem.tsx
dzonidoo: we definitely need to talk about this

assets/agenda/components/AgendaListItem.tsx
petrjasek: I see, makes sense then

tomaskikutis:
remains to be tested by @dzonidoo

tomaskikutis (approved):

PR #582, opened 2023-09-28
Fix folder fetching in the case when the user doesn't have a company

thecalcc:
NHUB-420

### DIFF
--- a/assets/user-profile/actions.ts
+++ b/assets/user-profile/actions.ts
@@ -299,8 +299,8 @@ export function fetchFolders() {
         const userTopicsUrl = getFoldersUrl(state, false);
 
         return Promise.all([
-            state.company !== 'None' ? server.get(({_items}: Array<any>) => _items) : Promise.resolve([]),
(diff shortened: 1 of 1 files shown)


tomaskikutis (approved):

assets/user-profile/actions.ts
```diff
@@ -299,8 +299,8 @@ export function fetchFolders() {
         const userTopicsUrl = getFoldersUrl(state, false);
 
         return Promise.all([
-            state.company !== 'None' ? server.get(({_items}: Array<any>) => _items) : Promise.resolve([]),
-            server.get(({_items}: Array<any>) => _items),
```
tomaskikutis:
```
server.get(({_items}: Array<any>) => _items),
                   // ^ this type is for entire argument, not only what you destructure.
```

it should be `server.get(({_items}: {_items: Array<any>}) => _items),`

thecalcc: Actually the URL to fetch from is missign too :D

PR #587, opened 2023-10-02
Add folder icon to every dropdown option. Add dropdown full width option

thecalcc:
CPCN-263

part 2

### DIFF
--- a/assets/search/components/TopicForm.tsx
+++ b/assets/search/components/TopicForm.tsx
@@ -134,7 +134,7 @@ const TopicForm: React.FC<IProps> = ({
                                 }
                             >
                                 <Dropdown
-                                    fullWidth={true}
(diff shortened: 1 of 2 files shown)


assets/search/components/TopicForm.tsx
```diff
@@ -134,6 +134,7 @@ const TopicForm: React.FC<IProps> = ({
                                 }
                             >
                                 <Dropdown
+                                    fullWidth={true}
```
petrjasek, line 137: shouldn't this be actually handled already via `stretch` ?

thecalcc: No the stretch is for the `button` only. Not the options popover.

tomaskikutis (approved):

PR #593, opened 2023-10-05
set timestamp format based on the NH settings

dzonidoo:
CPCN-363

### DIFF
--- a/assets/user-profile/components/EditNotificationScheduleModal.tsx
+++ b/assets/user-profile/components/EditNotificationScheduleModal.tsx
@@ -3,7 +3,7 @@ import {connect} from 'react-redux';
 import moment from 'moment-timezone';
 
 import {IUser} from 'interfaces';
-import {gettext, getScheduledNotificationConfig, TIME_FORMAT_24_HOUR_CLOCK} from 'utils';
+import {gettext, getScheduledNotificationConfig, TIME_FORMAT} from 'utils';
 import {modalFormInvalid, modalFormValid} from 'actions';
 import {updateUserNotificationSchedules} from 'user-profile/actions';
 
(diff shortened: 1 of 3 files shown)


assets/utils.tsx
```diff
@@ -59,6 +59,10 @@ const getCoverageDateTimeFormat = () =>
 const getCoverageDateFormat = () => getLocaleFormat('COVERAGE_DATE_FORMAT');
 
 export const TIME_FORMAT = getTimeFormat();
+export const TIME_FORMAT_24_HOUR_CLOCK: boolean = new Intl.DateTimeFormat(
```
petrjasek, line 62: imo you should rather check for `HH:` in TIME_FORMAT, otherwise this is not using NH settings really

tomaskikutis: > otherwise this is not using NH settings really

It is using NH settings, but I realized we're using locale, and not time format directly which might cause problems if you wanted different time format from locale. Nikola will update it to use the time format

assets/user-profile/components/EditNotificationScheduleModal.tsx
```diff
@@ -52,6 +53,8 @@ class EditNotificationScheduleModalComponent extends React.Component<IProps, ISt
 
         this.onSubmitForm = this.onSubmitForm.bind(this);
         this.formRef = React.createRef<HTMLFormElement>();
+
+        this.TIME_FORMAT = TIME_FORMAT.split(':')[0] === 'HH' ? '24-hours' : '12-hours';
```
tomaskikutis, line 57: would it work to move this as a constant at the top of the file instead to being specific to a class instance?

dzonidoo: yes.. and I will do it

PR #596, opened 2023-10-05
check for user products in the view

petrjasek:
NHUB-329

### DIFF
--- a/newsroom/wire/views.py
+++ b/newsroom/wire/views.py
@@ -14,6 +14,7 @@ from flask_babel import gettext
 from superdesk.utc import utcnow
 from superdesk import get_resource_service
 from superdesk.default_settings import strtobool
+from newsroom.auth.utils import check_user_has_products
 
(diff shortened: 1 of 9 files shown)


newsroom/wire/views.py
```diff
@@ -87,6 +88,10 @@ def get_view_data():
     company_id = str(user["company"]) if user and user.get("company") else None
     user_folders = get_user_folders(user, "wire") if user else []
     company_folders = get_company_folders(company, "wire") if company else []
+    products = get_products_by_company(company, product_type="wire") if company else []
+
+    if user:
```
MarkLark86, line 93: As this method is only used in views decorated with `@login_required`, we could assume that a user should always be available here, otherwise it might be an auth issue.

Do you see any scenario where a user would not be defined here (and in agenda `get_view_data`)?

MarkLark86 (approved):

PR #597, opened 2023-10-06
Quick filters improvements

thecalcc:
NHUB-411

### DIFF
--- a/assets/agenda/components/AgendaCalendarAgendaFilter.tsx
+++ b/assets/agenda/components/AgendaCalendarAgendaFilter.tsx
@@ -14,7 +14,6 @@ export class AgendaCalendarAgendaFilter extends React.PureComponent<any, any> {
         const calendarFilter = {
             label: gettext('Any calendar'),
             field: 'calendar',
-            icon: 'icon-small--calendar',
             itemTypes: ['events', 'combined'],
             isItemActive: (key: any) => (
                 (this.props.activeFilter.calendar || []).includes(key)
@@ -24,7 +23,6 @@ export class AgendaCalendarAgendaFilter extends React.PureComponent<any, any> {
             label: gettext('Any agenda'),
             field: 'agendas',
             nestedField: 'agenda',
-            icon: 'icon-small--calendar',
             itemTypes: ['planning', 'combined'],
             isItemActive: (key: any) => (
                 (this.props.activeFilter.agendas || []).includes(key)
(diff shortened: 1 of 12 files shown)


assets/agenda/components/AgendaCalendarAgendaFilter.tsx
```diff
@@ -45,30 +43,36 @@ export class AgendaCalendarAgendaFilter extends React.PureComponent<any, any> {
...
-                    <React.Fragment>
-                        <h6 className="dropdown-header">{gettext('Events')}</h6>
-                        {calendarItems}
-                        <div className="dropdown-divider" />
-                    </React.Fragment>
-                )}
+                {allCalendarsButton}
+                {!calendarItems.length
+                    ? (
+                        <div className='dropdown-item__empty'>No Calendars available</div>
```
petrjasek, line 68: use `gettext`

assets/components/Dropdown.tsx
```diff
@@ -38,28 +48,39 @@ export function Dropdown({children, isActive, icon, optionLabel, label, value, c
...
-                    {
-                        'nh-dropdown-button--active': isActive,
-                        'nh-dropdown-button--text-only': textOnly,
-                        'nh-dropdown-button--small': small,
-                        'nh-dropdown-button--stretch': stretch,
-                    }
-                )}
+                className={classNames({
+                    'nh-dropdown-button--borderless': borderless,
+                    'nh-dropdown-button': !borderless,
```
petrjasek, line 53: I wonder why this shouldn't be always there, seems like `--borderless` is modifier but the component itself would be missing the main class without it

thecalcc: right, it should always be there

assets/components/Dropdown.tsx
```diff
@@ -85,8 +107,8 @@ Dropdown.propTypes = {
     reset: PropTypes.func,
     small: PropTypes.bool,
     stretch: PropTypes.bool,
-    buttonProps: PropTypes.shape({
-        textOnly: PropTypes.bool,
-    }),
     optionLabel: PropTypes.string,
+    borderless: PropTypes.bool,
+    hideLabel: PropTypes.bool,
+    hideLabelMobile: PropTypes.bool,
```
petrjasek, line 113: are those actually set anywhere? I can't find any usage

thecalcc: Not currently, but I had to add support for them because that are the requirements. So now we'll be using them sometime in the future I guess.

petrjasek (approved):

PR #598, opened 2023-10-08
Drag and drop v2

tomaskikutis:
CPCN-276

https://github.com/superdesk/newsroom-core/assets/2076953/7ab886d3-17df-47a3-bfcc-1a79d6a6e3b4


assets/search/components/TopicList.tsx
```diff
@@ -63,21 +79,103 @@ const TopicList = ({
...
-    const renderedTopics = topics.filter((topic: any) => topic.folder == null).map(renderTopic);
+    const topicsWithoutFolder = topics.filter((topic) => topic.folder == null).map((topic) => renderTopic(topic));
+
+    const sensors = useCustomSensors({activationConstraint: {distance: 10}});
 
     return (
         <>
-            {renderedFolders}
-            {renderedTopics}
+            <DndContext
```
petrjasek, line 103: I wonder if it would be possible to have a reusable component for the context and the overlay which would get the children and `onDragEnd` handler

tomaskikutis: Restoring scroll was the ugly repetetive part IMO, I've moved it into a custom hook. I think not making that component is cleaner for now. We can come back when we have more use cases and see if it would help to abstract it further.

tomaskikutis:
FYI I've found that current method of drag and dropping that is used in e2e test doesn't work, so I'll need to come up with a different approach.

petrjasek (approved):

PR #607, opened 2023-10-10
Improvements

thecalcc:
NHUB-411


tomaskikutis:
Could you explain the purpose and behavior of `resetOptionLabel`?

thecalcc:
> Could you explain the purpose and behavior of `resetOptionLabel`?

Fritz wanted an option so the label of the main button can be separated from the label inside the "reset" button, so I introduced it because of that

thecalcc:
Can you approve if everything is fine? @tomaskikutis

petrjasek (approved):

PR #614, opened 2023-10-17
[NHUB-315] Support Video in featuremedia

MarkLark86:
## Checklist


MarkLark86:
Will fix this in a new PR

PR #615, opened 2023-10-19
[NHUB-315] (2.4.2): Support Audio & Video for Featuremedia

MarkLark86:
## Checklist

### DIFF
--- a/assets/wire/utils.js
+++ b/assets/wire/utils.js
@@ -49,6 +49,35 @@ function getRelatedItemsByType(item, type) {
     return item.type === type ? [item] : Object.values(get(item, 'associations', {}) || {}).filter((assoc) => get(assoc, 'type') === type);
 }
 
+export function getFeatureMedia(item) {
+    if (['picture', 'video', 'audio'].includes(item.type)) {
+        return item;
+    }
+
+    const featured = get(item, 'associations.featuremedia');
+
+    if (featured != null && ['picture', 'video', 'audio'].includes(featured.type)) {
+        return featured;
+    }
+
+    return getBodyPicture(item);
+}
+
+export function getOtherMedia(item) {
+    if (['picture', 'video', 'audio'].includes(item.type)) {
+        return null;
+    }
+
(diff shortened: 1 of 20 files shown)


assets/wire/utils.js
```diff
@@ -49,6 +49,35 @@ function getRelatedItemsByType(item, type) {
     return item.type === type ? [item] : Object.values(get(item, 'associations', {}) || {}).filter((assoc) => get(assoc, 'type') === type);
 }
 
+export function getFeatureMedia(item) {
+    if (['picture', 'video', 'audio'].includes(item.type)) {
+        return item;
+    }
+
+    const featured = get(item, 'associations.featuremedia');
```
tomaskikutis, line 57: it'd be nice to replace lodash get calls with optional chaining so we don't lose typescript support (find usages, rename)

assets/wire/utils.js
```diff
@@ -49,6 +49,35 @@ function getRelatedItemsByType(item, type) {
...
+}
+
+export function getOtherMedia(item) {
+    if (['picture', 'video', 'audio'].includes(item.type)) {
+        return null;
+    }
+
+    return Object.keys(item.associations || {})
+        .filter((key) => (
+            !key.startsWith('editor_') &&
```
tomaskikutis, line 73: what the meaning of `key.startsWith('editor_')`?

tomaskikutis (approved):

assets/wire/utils.js
MarkLark86: This PR is the javascript version for v2.4 of Newshub. The other PR is for typescript (v2.5+) which does not use the `locash.get` function.

assets/wire/utils.js
MarkLark86: If you look at `apps.archive.archive.update_associations` in the backend (https://github.com/superdesk/superdesk-core/blob/develop/apps/archive/archive.py#L212), media items from `fields_meta.body_html.draftjsState.body_html` metadata are stored in `assoications` under the `editor_` prefix

PR #621, opened 2023-10-19
remove 'Sort by' drop down in the Calendar/Agenda section [CPCN-409]

devketanpro:
## Checklist

### DIFF
--- a/assets/search/components/SearchResultsBar/index.tsx
+++ b/assets/search/components/SearchResultsBar/index.tsx
@@ -146,7 +146,7 @@ class SearchResultsBarComponent extends React.Component<any, any> {
                                 </div>
                             )}
                             <div className="navbar__button-group">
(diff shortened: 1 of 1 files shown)


assets/search/components/SearchResultsBar/index.tsx
```diff
@@ -146,7 +146,7 @@ class SearchResultsBarComponent extends React.Component<any, any> {
                                 </div>
                             )}
                             <div className="navbar__button-group">
-                                <Dropdown
+                                {this.props.topicType == 'wire' ? <Dropdown
```
petrjasek, line 149: we usually use `===`

devketanpro: my bad

petrjasek (approved):

PR #627, opened 2023-10-23
check user authorization on api calls

petrjasek:
NHUB-430

## Checklist

### DIFF
--- a/newsroom/auth/utils.py
+++ b/newsroom/auth/utils.py
@@ -97,6 +97,7 @@ def start_user_session(user: User, permanent=False, session=None):
     session["user_type"] = user["user_type"]
     session["auth_ttl"] = utcnow().replace(tzinfo=None) + SESSION_AUTH_TTL
     session.permanent = permanent
+    print("STARTED", session)
 
 
 def clear_user_session(session=None):
--- a/newsroom/auth/utils.py
+++ b/newsroom/auth/utils.py
@@ -97,7 +97,6 @@ def start_user_session(user: User, permanent=False, session=None):
     session["user_type"] = user["user_type"]
     session["auth_ttl"] = utcnow().replace(tzinfo=None) + SESSION_AUTH_TTL
     session.permanent = permanent
-    print("STARTED", session)
 
 
 def clear_user_session(session=None):
--- a/newsroom/auth/views.py
+++ b/newsroom/auth/views.py
@@ -169,7 +169,6 @@ def login_with_token(token):
 @blueprint.route("/logout")
 def logout():
     clear_user_session()
-    print("CLEARED")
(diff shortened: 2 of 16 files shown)


newsroom/auth/utils.py
```diff
@@ -97,6 +97,7 @@ def start_user_session(user: User, permanent=False, session=None):
     session["user_type"] = user["user_type"]
     session["auth_ttl"] = utcnow().replace(tzinfo=None) + SESSION_AUTH_TTL
     session.permanent = permanent
+    print("STARTED", session)
```
MarkLark86, line 100: Remove `print` statement

newsroom/auth/views.py
```diff
@@ -169,6 +169,7 @@ def login_with_token(token):
 @blueprint.route("/logout")
 def logout():
     clear_user_session()
+    print("CLEARED")
```
MarkLark86, line 172: Remove `print` statement

newsroom/users/users.py
```diff
@@ -19,6 +20,32 @@
 from newsroom.companies.utils import get_company_section_names, get_company_product_ids
 
 
+class UserAuthentication(SessionAuth):
+    def authorized(self, allowed_roles, resource, method):
+        if super().authorized(allowed_roles, resource, method):
```
MarkLark86, line 25: Should this rather be `if not super().authorized(allowed_roles, resource, method): return False`?

newsroom/users/users.py
```diff
@@ -19,6 +20,32 @@
...
+        if not request.view_args or not request.view_args.get("_id"):
+            # not a request for a specific user, stop
+            return False
+
+        if request.view_args["_id"] == str(get_user_id()):
+            # current user editing current user
+            return True
+
+        current_user = get_user()
+        if not current_user.get("company") or current_user.get("user_type") != UserRole.COMPANY_ADMIN.value:
```
MarkLark86, line 37: Why return `False` if current user has no company?

MarkLark86, line 37: Should we also allow system admins to modify users?

newsroom/users/users.py
petrjasek: if user is authorized because has the required role needed I return True, here I'm handling the case when user is not authorized via his role but it might be his profile or a profile from his company if he is an admin.

but I should add check for authentication here, that's missing

newsroom/users/users.py
petrjasek: because here the last possible case is it would be company admin for the edited user, but if there is no company then there won't be any link

petrjasek: that's handled by the initial `if authorized(): return True`

MarkLark86 (approved):

PR #630, opened 2023-10-24
Cpcn 403

darconny:
## Checklist


fritzSF (approved):

petrjasek:
do we need this also for command news?

fritzSF:
> do we need this also for command news?

We probably do, @darconny did some genaral changes in the themes nad custom-variables files

PR #638, opened 2023-10-26
fix display of empty topics message

petrjasek:
so it's visible based on selected my/company topics.

CPCN-415

## Checklist

### DIFF
--- a/assets/components/PersonalizeHomeModal.tsx
+++ b/assets/components/PersonalizeHomeModal.tsx
@@ -172,12 +172,14 @@ class PersonalizeHomeModal extends React.Component<IProps, IState> {
 
         const topicSearch = (searchMatches?.length ?? 0) > 0
             ? (
-                <div style={{padding: 4}}>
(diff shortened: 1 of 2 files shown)


assets/components/PersonalizeHomeModal.tsx
```diff
@@ -259,7 +261,7 @@ class PersonalizeHomeModal extends React.Component<IProps, IState> {
                                 {
                                     gettext(
                                         '{{size}} out of 6 topics selected',
-                                        {size: this.state.selectedTopicIds.length}
+                                        {size: this.state.selectedTopicIds.length || '0'}
```
thecalcc, line 264: I think it'd be better if you used `??`

thecalcc (approved):

assets/components/PersonalizeHomeModal.tsx
petrjasek: found the problem why it didn't work, it was in the gettext

PR #643, opened 2023-10-30
[CPCN-27] ui: Support hook for modifying versions label text

MarkLark86:
## Checklist

### DIFF
--- a/assets/api.ts
+++ b/assets/api.ts
@@ -1,0 +1,24 @@
+import {IArticle} from './interfaces';
+import {gettext} from './utils';
+
+export interface INewshubApi {
+    ui: {
+        wire: {
+            getVersionsLabelText(item: IArticle, plural?: boolean): string;
+        };
+    };
+    localization: {
+        gettext(message: string, params?: {[key: string]: string | number}): string;
+    };
+}
+
+function getVersionsLabelText(_item: IArticle, plural?: boolean): string {
+    return plural === true ?
+        gettext('version') :
+        gettext('versions');
+}
+
+export const newshubApi: INewshubApi = {
+    ui: {wire: {getVersionsLabelText: getVersionsLabelText}},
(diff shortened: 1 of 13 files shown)


assets/index.ts
```diff
@@ -1,6 +1,23 @@
...
+export function startApp(): Promise<INewshubApi> {
+    if (started === true) {
+        return Promise.resolve(newshubApi);
+    }
+
+    started = true;
+    return Promise.resolve(newshubApi);
+}
+
+setTimeout(() => {
```
petrjasek, line 19: I'm wondering what's the usecase for this startApp to be called? just to set the `started` flag?

assets/api.ts
```diff
@@ -0,0 +1,24 @@
+import {IArticle} from './interfaces';
+import {gettext} from './utils';
+
+export interface INewshubApi {
```
petrjasek, line 4: maybe we should call it like `INewshubExtensionApi` if we expect it to be overriden, I would expect `INewshubApi` something I can use but not change

assets/api.ts
```diff
@@ -0,0 +1,24 @@
...
+        wire: {
+            getVersionsLabelText(item: IArticle, plural?: boolean): string;
+        };
+    };
+    localization: {
+        gettext(message: string, params?: {[key: string]: string | number}): string;
+    };
+}
+
+function getVersionsLabelText(_item: IArticle, plural?: boolean): string {
```
petrjasek, line 15: I'm wondering if we should have rather the function return some predefined options like:
```ts
getAncestorType(_item: IArticle): 'versions' | 'sequence';
```
and handle those options in core, that would make translations etc simpler

MarkLark86: @petr to keep it simple for now, I'll move this code into core. We can move it out later if desired.

PR #644, opened 2023-10-30
Show search conditionally

thecalcc:
CPCN-50
## Checklist

### DIFF
--- a/assets/wire/components/filters/FilterGroup.tsx
+++ b/assets/wire/components/filters/FilterGroup.tsx
@@ -79,27 +79,29 @@ export default function FilterGroup({group, activeFilter, aggregations, toggleFi
 
     return (
         <NavGroup key={group.field} label={group.label}>
-            <div className="mb-2 search search--small search--with-icon search--bordered m-0">
-                <div className="search__form" role="search" aria-label="search">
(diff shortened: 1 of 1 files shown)


assets/wire/components/filters/FilterGroup.tsx
```diff
@@ -79,27 +79,29 @@ export default function FilterGroup({group, activeFilter, aggregations, toggleFi
...
-                        type="text"
-                        className="search__input form-control"
-                        placeholder={gettext('Search Filters')}
-                        aria-label={gettext('Search Filters')}
-                    />
-                    <div className="search__form-buttons">
-                        <button className="search__button-clear" aria-label={gettext('Clear search')} type="button">
-                            {searchIcon}
-                        </button>
+            {buckets.length > 50 && (
```
petrjasek, line 82: not sure if we should hardcode it or add some config option for that

thecalcc: Idk, Gideon said we should do `50` and in the ticket it said `50` too, but let's discuss this with Fritz and Gideon when Gideon is back

petrjasek (approved):

PR #653, opened 2023-11-03
add new formatter CSV

devketanpro:
## Checklist

### DIFF
--- a/newsroom/agenda/formatters/csv_formatter.py
+++ b/newsroom/agenda/formatters/csv_formatter.py
@@ -1,0 +1,108 @@
+
+from newsroom.formatter import BaseFormatter
+import csv
+import io
+from superdesk.utc import utcnow
+# from typing import isinstance
+
+class CSVFormatter(BaseFormatter):
+    VERSION = "1.0"
+    PRODID = "Newshub"
+    FILE_EXTENSION = "csv"
+    MIMETYPE = "text/csv"
+
+    def format_item(self, item, item_type=None):
+
+        
+        event_item = self.format_event(item)
+        csv_string = io.StringIO()
+        csv_writer = csv.DictWriter(csv_string, delimiter=",", fieldnames=event_item.keys())
+        csv_writer.writeheader()
+        csv_writer.writerow(event_item)  # Write a single row for the provided item
+        csv_string.seek(0)  # Reset the buffer position
+
+        csv_data = csv_string.getvalue().encode('utf-8')  # Encode CSV data as bytes
+
+        return csv_data
+    
+
+    def format_event(self, item):
+
+        location , country = self.parse_location(item)
+
+        return {
+            "Event name": item.get("name", ""),
+            "Description": item.get("definition_long", item.get("definition_short", "")),
+            "Language": item.get("language", ""),
+            "Event start date":item.get("dates", {}).get("start", "").strftime('%Y-%m-%d'),
+            "Event end date": item.get("dates", {}).get("end", "").strftime('%Y-%m-%d'),
+            "Event time":item.get("dates", {}).get("start", "").strftime('%H:%M:%S'),
+            "Event timezone": item.get("dates", {}).get("tz", ""),
+            "Location": location if location else "",
+            "Country": country if country else "",
+            "Subject": self.parse_subject(item),
+            "Website": item.get("event", {}).get("links")[0],
+            "Category": self.parse_category(item),
+            "Event type": item.get("item_type", ""),
+            "Organization name": item.get("event", {}).get("event_contact_info",{})[0].get("organisation"," "),
+            "Contact": self.parse_contact_info(item),
+            "Coverage type": self.parse_coverage(item, "coverage_type"),
+            "Coverage status": self.parse_coverage(item, "coverage_status"),
+        }
+    
+
+    def parse_location(self, item):
+        """
+        parse location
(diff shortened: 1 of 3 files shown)


newsroom/agenda/formatters/csv_formatter.py
```diff
@@ -0,0 +1,99 @@
...
+            "Coverage status": self.parse_coverage(item, "coverage_status"),
+        }
+
+    def format_date(self, item, date_type):
+        date_obj = item.get("dates", {}).get(date_type)
+        if date_obj:
+            return date_obj.strftime("%Y-%m-%d")
+        return ""
+
+    def format_time(self, item, date_type):
```
petrjasek, line 52: I think we need to handle also events without end time and all day events

devketanpro: I see, so If event is `all_day = true` in this case we have to add 'all_day' in  `event time `
and if no `end_time` then show only `start time`  right ?

petrjasek: I guess we can keep it empty when it's all day, for `no_end_time` only show start time

devketanpro: Okay Got it

petrjasek (approved):

petrjasek:
actually I'm wondering now - does it handle multiple items?

devketanpro:
> so if you select multiple events you get 1 csv file with multiple rows?

no we get zip contains multiple csv files

petrjasek:
I think we should change it to be able to get a single csv with multiple items

newsroom/wire/views.py
```diff
@@ -304,6 +304,12 @@ def download(_ids):
         _file.seek(0)
         mimetype = formatter.get_mimetype(item)
         attachment_filename = secure_filename(formatter.format_filename(item))
+    elif formatter.FILE_EXTENSION == "csv" and len(items) != 1:
```
petrjasek, line 307: would be better if the formatter can define it can handle multiple files instead of hardcoding it here for specific one, eg.:
```py
if formatter.multi is True:
    data, filename = formatter.format_items(...)
```

devketanpro: got it, Thanks !

newsroom/agenda/formatters/csv_formatter.py
```diff
@@ -0,0 +1,126 @@
...
+    def format_items(self, items, item_type=None):
+        formatted_event = []
+        for item in items:
+            parse_dates(item)
+            formatted_event.append(self.format_event(item))
+        return self.serialize_to_csv(formatted_event), secure_filename(
+            f"{datetime.now().strftime('%Y-%m-%d-%H:%M:%S')}-{'multi'}.{self.FILE_EXTENSION}"
+        )
+
+    def serialize_to_csv(self, data):
```
petrjasek, line 30: would be good to use something more descriptive than `data`, it's not clear if it should be one item or a list

newsroom/agenda/formatters/csv_formatter.py
```diff
@@ -0,0 +1,126 @@
...
+    FILE_EXTENSION = "csv"
+    MIMETYPE = "text/csv"
+    MULTI = True
+
+    def format_item(self, item, item_type=None):
+        event_item = self.format_event(item)
+        return self.serialize_to_csv(event_item)
+
+    def format_items(self, items, item_type=None):
+        formatted_event = []
```
petrjasek, line 22: this should be rather `formatted_events` imo

newsroom/agenda/formatters/csv_formatter.py
```diff
@@ -0,0 +1,126 @@
...
+                contact.get("honorific", ""),
+                contact.get("first_name", ""),
+                contact.get("last_name", ""),
+                contact.get("organisation", ""),
+                ",".join(contact.get("contact_email", [])),
+                ",".join(contact.get("mobile", [])),
+            ]
+            return ",".join(contact_values)
+
+    def parse_coverage(self, item, field):
```
petrjasek, line 117: in general there is no parsing going on, it's rather formatting, so those names are a bit confusing

newsroom/agenda/formatters/csv_formatter.py
```diff
@@ -0,0 +1,126 @@
+from newsroom.formatter import BaseFormatter
+import csv
+import io
+import arrow
+from werkzeug.utils import secure_filename
+from newsroom.utils import parse_dates
+from datetime import datetime
+
+
+class CSVFormatter(BaseFormatter):
```
petrjasek, line 10: the methods here could use some type definitions

newsroom/agenda/formatters/csv_formatter.py
```diff
@@ -27,21 +28,21 @@ def format_items(self, items, item_type=None):
             f"{datetime.now().strftime('%Y-%m-%d-%H:%M:%S')}-{'multi'}.{self.FILE_EXTENSION}"
         )
 
-    def serialize_to_csv(self, data):
+    def serialize_to_csv(self, items: Union[List[Dict[str, Any]], Dict[str, Any]]) -> bytes:
```
petrjasek, line 31: imo it would be better to have this work with just list, you can call it from `format_event` with `[event]`, that would simplify the types and code in general

PR #656, opened 2023-11-07
Search disapearing fix

thecalcc:
Fix CPCN-50
## Checklist

### DIFF
--- a/assets/wire/components/filters/FilterGroup.tsx
+++ b/assets/wire/components/filters/FilterGroup.tsx
@@ -8,6 +8,10 @@ import FilterItem from './FilterItem';
 import {WithPagination} from 'components/pagination/WithPagination';
 import {searchIcon} from 'search/components/search-icon';
 
+/**
+ * If the filter count is less than 50 we
+ * shouldn't render the search.
+ */
 const SEARCHBAR_THRESHOLD = 50;
 const LIMIT = 5;
 type IBucket = {key: string, doc_count: string};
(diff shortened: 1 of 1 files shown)


thecalcc:
Currently there's a bug, when you type something the buckets are getting filtered by search term and in the case where there are > 50 buckets we hide the searchbar which isn't what we want.

assets/wire/components/filters/FilterGroup.tsx
```diff
@@ -70,16 +70,17 @@ export default function FilterGroup({group, activeFilter, aggregations, toggleFi
...
-    )
+    );
+    const bucketsBySearchTerm = allBuckets
         .sort(compareFunction)
         .filter(({key}: any) => searchTerm.length > 0 ? key.toString().toLocaleLowerCase().includes(searchTerm) : true);
 
     return (
         <NavGroup key={group.field} label={group.label}>
-            {buckets.length > 50 && (
+            {allBuckets.length > 50 && (
```
tomaskikutis, line 83: why `50`? shouldn't you use a `pageSize` variable here?

thecalcc: 50 was in the ticket requirement

assets/wire/components/filters/FilterGroup.tsx
```diff
@@ -8,6 +8,7 @@ import FilterItem from './FilterItem';
 import {WithPagination} from 'components/pagination/WithPagination';
 import {searchIcon} from 'search/components/search-icon';
 
+const SEARCHBAR_THRESHOLD = 50;
```
tomaskikutis, line 11: can you add a comment on what exactly was the requirement?

tomaskikutis (approved):

PR #658, opened 2023-11-09
avoid unique error messages

petrjasek:
so it won't create a new error on sentry every time but can aggregate those together.

## Checklist

### DIFF
--- a/newsroom/notifications/send_scheduled_notifications.py
+++ b/newsroom/notifications/send_scheduled_notifications.py
@@ -8,7 +8,7 @@ from flask import current_app as app
 from superdesk import get_resource_service, Command
 from superdesk.utc import utcnow, utc_to_local
 from superdesk.celery_task_utils import get_lock_id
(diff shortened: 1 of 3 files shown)


newsroom/notifications/send_scheduled_notifications.py
```diff
@@ -76,7 +76,7 @@ def run_schedules(self, force: bool):
                 self.process_schedule(schedule, user, company, now_utc, user_topic_map.get(user["_id"]) or {}, force)
             except Exception as e:
                 logger.exception(e)
-                logger.error(f"{self.log_msg} Failed to run schedule for user {user_id}")
+                logger.error("Failed to run schedule for user", extra={"user": user_id})
```
MarkLark86, line 79: This may work for sentry, but on console logs will no longer output the `user_id` (without changing the formatter)

MarkLark86 (approved):

PR #659, opened 2023-11-13
[CPCN-27] Show segment instead of version for transcripts

MarkLark86:
## Checklist

### DIFF
--- a/assets/wire/actions.ts
+++ b/assets/wire/actions.ts
@@ -524,7 +524,7 @@ export function fetchNewItems() {
 }
 
 export function fetchNext(item: IArticle): Promise<IArticle> {
-    if (!item.nextversion) {
+    if (item.nextversion == null) {
         return Promise.reject();
     }
 
--- a/assets/wire/actions.ts
+++ b/assets/wire/actions.ts
@@ -1,5 +1,6 @@
 import {get, isEmpty} from 'lodash';
 
+import {IArticle} from 'interfaces';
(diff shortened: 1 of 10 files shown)


assets/wire/actions.ts
```diff
@@ -524,14 +523,12 @@ export function fetchNewItems() {
         .then((response: any) => dispatch(setNewItems(response)));
 }
 
-export function fetchNext(item: any) {
-    return () => {
-        if (!item.nextversion) {
-            return Promise.reject();
-        }
+export function fetchNext(item: IArticle): Promise<IArticle> {
+    if (!item.nextversion) {
```
thecalcc, line 527: I think it's better to check this explicitly without using `!`

assets/wire/utils.ts
```diff
@@ -339,3 +339,13 @@ export function getContentTypes(item: IArticle): Set<IContentType> {
 
 export const hasAudio = (item: any) => hasMedia(item, 'audio');
 export const hasVideo = (item: any) => hasMedia(item, 'video');
+
+export function getVersionsLabelText(item: IArticle, plural?: boolean): string {
+    const isTranscript = item.extra?.type === 'transcript';
```
thecalcc, line 344: Not sure if `extra.type` is typed here, in any case extracting 'transcript' to a constant should be better for future reference

thecalcc (approved):

PR #664, opened 2023-11-22
implement change password in user profile

petrjasek:
in case the auth provider supports it.
also move the firebase auth integration to core.

CPCN-206

## Checklist


assets/user-profile/components/profile/UserProfile.tsx
```diff
@@ -202,17 +217,21 @@ class UserProfile extends React.Component<IProps, any> {
...
 const mapDispatchToProps = (dispatch: any) => ({
     saveUser: () => dispatch(saveUser()),
     fetchUser: (userId: IUser['_id']) => dispatch(fetchUser(userId)),
-    onChange: (event: React.ChangeEvent) => dispatch(editUser(event)),
+    onChange: (updates: IUserProfileUpdates) => dispatch(editUser(updates)),
     setError: (errors: {[key: string]: string}) => dispatch(setError(errors)),
     openEditTopicNotificationsModal: () => dispatch(openEditTopicNotificationsModal()),
 });
 
+type IStateProps = typeof mapStateToProps;
```
MarkLark86, line 234: These 2 types aren't used anywhere. Also, the version of Redux we're using doesn't play nicely with typescript (I have updated it in my multiday events PR to the newest version that does)

newsroom/auth/providers.py
```diff
@@ -0,0 +1,64 @@
+import abc
+
+# from type
```
MarkLark86, line 3: Commented line not needed

newsroom/auth/views.py
```diff
@@ -283,3 +308,60 @@ def impersonate_stop():
...
+        if auth_provider.type == AuthProviderType.FIREBASE:
+            if form.data.get("firebase_status"):
+                firebase_status = form.data["firebase_status"]
+                if firebase_status == "OK":
+                    flask.flash(gettext("Your password has been changed."), "success")
+                elif firebase_status == "auth/wrong-password":
+                    flask.flash(gettext("Wrong current password."), "error")
+                else:
+                    logger.warning("Unhandled firebase error %s", firebase_status)
+                    flask.flash(gettext("Could not change password. Please try again later."), "warning")
```
MarkLark86, line 332: I think `Please try again later` might not be good wording here. The user would be thinking if it doesn't work now, how could it work later.
What if we instead suggest to them to contact their administrator?

assets/auth/firebase/reset_password.ts
```diff
@@ -0,0 +1,43 @@
...
+        const data = new FormData(form);
+        const email = data.get('email') as string;
+
+        params.append('email', email);
+        url.search = params.toString();
+
+        sendButton.disabled = true;
+        sendPasswordResetEmail(auth, email, {url: url.toString()})
+            .then(() => {
+                // set `email_sent` to true, so server knows password reset was handled by firebase
```
MarkLark86, line 28: I see `email_sent` is no longer used, as we're checking the auth provider on the server side. Therefor this comment is no longer required

MarkLark86 (approved):
Looks fine to me, only a couple small things

assets/auth/firebase/change_password.ts
```diff
@@ -0,0 +1,35 @@
...
+        const data = new FormData(form);
+        const oldPassword = data.get('old_password') as string;
+        const newPassword = data.get('new_password') as string;
+        const newPassword2 = data.get('new_password2') as string;
+
+        if (newPassword !== newPassword2) {
+            form.submit();
+        }
+
+        signInWithEmailAndPassword(auth, firebaseUserEmail, oldPassword).then((userCredential) => {
```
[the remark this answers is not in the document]
petrjasek: no, but will add there return after submit to make it more explicit

MarkLark86 (approved):

PR #672, opened 2023-11-27
fix picture text card display error with missing picture

petrjasek:
CPCN-488

## Checklist

### DIFF
--- a/assets/components/cards/render/PictureTextCard.tsx
+++ b/assets/components/cards/render/PictureTextCard.tsx
@@ -4,9 +4,16 @@ import CardBody from './CardBody';
 import CardFooter from './CardFooter';
 import CardRow from './CardRow';
 import {ICardProps} from '../utils';
-import {IArticle} from 'interfaces';
+import {IArticle, IListConfig} from 'interfaces';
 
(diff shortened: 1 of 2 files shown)


assets/components/cards/render/PictureTextCard.tsx
```diff
@@ -4,19 +4,20 @@ import CardBody from './CardBody';
 import CardFooter from './CardFooter';
 import CardRow from './CardRow';
 import {ICardProps} from '../utils';
+import {IArticle} from 'interfaces';
 
-const getPictureTextPanel = (item: any, picture: any, openItem: any, withPictures: any, cardId: any, listConfig: any) => {
-    const rendition = withPictures && getThumbnailRendition(picture);
+const getPictureTextPanel = (item: any, picture: IArticle | null, openItem: any, withPictures: boolean, cardId: any, listConfig: any) => {
```
MarkLark86, line 9: I believe we could also have types for the other arguments here.
* item: IArticle;
* openItem(IArticle, string): void;
* cardId: string;
* listConfig: IListConfig

MarkLark86 (approved):

PR #673, opened 2023-11-28
Fix labels

thecalcc:
NHUB-356
## Checklist


MarkLark86:
@thecalcc Please provide a more descriptive commit title than `fix labels`. If you mention what labels are fixed, this will give context to the title and becomes meaningful as to what was changed.

PR #674, opened 2023-11-29
[CPCN-482] and [CPCN-486]

fritzSF:
## Checklist


MarkLark86 (approved):
@fritzSF Is there a need for the `package-lock.json` changes? if not it would be good to remove it from this PR.

Generally if `package.json` doesn't change, we don't need to include changes from the lock file

fritzSF:
> @fritzSF Is there a need for the `package-lock.json` changes? if not it would be good to remove it from this PR.
> 
> Generally if `package.json` doesn't change, we don't need to include changes from the lock file

Hey @MarkLark86. No, no reason for that as far as I know. I will remove it.

PR #675, opened 2023-11-30
Carousel implementation

thecalcc:
## Checklist


tomaskikutis (commented):
remove empty file - `ItemFeatureMedia.tsx`

tomaskikutis:
What about the import? Didn't we say you'd move it to some utils file that is imported in every page?

thecalcc:
> What about the import? Didn't we say you'd move it to some utils file that is imported in every page?

we did but that didn't work, remember when I showed you? Because every page is rendered as a separate app, and also because it just didn't work when I imported it in that utils file

tomaskikutis:
We need to solve it some other way then. What about if you import it inside `BaseApp`?

PR #679, opened 2023-12-04
implement item level expiry based on source

petrjasek:
add `SOURCE_EXPIRY_DAYS` config which is used
to set specific expiry for items and remove those
during expiry cleanup.

CPCN-29

uses https://github.com/superdesk/superdesk-core/pull/2499

### DIFF
--- a/newsroom/wire/items.py
+++ b/newsroom/wire/items.py
@@ -1,8 +1,9 @@
 from copy import deepcopy
 
 import superdesk
+
 from content_api.items.resource import ItemsResource as BaseItemsResource
-from content_api.items.service import ItemsService
+from content_api.items.service import ItemsService as BaseItemsService
 
 from content_api.items_versions.resource import ItemsVersionsResource as BaseItemsVersionsResource
 from content_api.items_versions.service import ItemsVersionsService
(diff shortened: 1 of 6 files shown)


newsroom/wire/items.py
```diff
@@ -15,6 +16,18 @@ class ItemsResource(BaseItemsResource):
     schema["slugline"] = schema["headline"] = schema["body_html"] = schema["description_html"] = metadata_schema[
         "body_html"
     ].copy()
+    schema["expiry"] = {"type": "datetime", "reaonly": True}
+
+
+class ItemsService(BaseItemsService):
+    def get_expired_items(self, expiry_datetime=None, expiry_days=None, max_results=None, include_children=True):
+        # remove old items based on expiry days config
+        for items in super().get_expired_items(expiry_datetime, expiry_days, max_results, include_children):
```
MarkLark86, line 25: Is it desired that an item will expire if **ANY** of the two conditions are met:
* `_updated` <= expiry_datetime
* `expired` <= expiry_datetime

It seems as though the `SOURCE_EXPIRY_DAYS` is limited up to `CONTENT_API_EXPIRY_DAYS` and cannot be longer than that.

newsroom/wire/items.py
```diff
@@ -15,6 +16,18 @@ class ItemsResource(BaseItemsResource):
...
+            yield items
+        # remove items with custom expiry
+        lookup = {"expiry": {"$lt": expiry_datetime}}
+        items = list(self.get_from_mongo(req=None, lookup=lookup))
```
MarkLark86, line 29: We should probably split these out into chunks, using `max_results` and `last_id` similar to what is used in the parent `get_expired_items`, and yielding each `page` of items

newsroom/wire/items.py
```diff
@@ -15,6 +16,18 @@ class ItemsResource(BaseItemsResource):
     schema["slugline"] = schema["headline"] = schema["body_html"] = schema["description_html"] = metadata_schema[
         "body_html"
     ].copy()
+    schema["expiry"] = {"type": "datetime", "reaonly": True}
```
MarkLark86, line 19: Would it be beneficial to have a mongo index for `expiry`?

petrjasek:
made some more changes in sd https://github.com/superdesk/superdesk-core/pull/2501

newsroom/wire/items.py
petrjasek: changed the core expiry to ignore items with expiry set

MarkLark86 (approved):

PR #682, opened 2023-12-06
fix navigation query for users

petrjasek:
NHUB-456 CPCN-469

### DIFF
--- a/tests/core/test_wire.py
+++ b/tests/core/test_wire.py
@@ -1,3 +1,4 @@
+from pytest import fixture
 import pytz
 from flask import json, g, session as server_session
 from datetime import datetime, timedelta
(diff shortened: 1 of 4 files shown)


tests/core/test_wire.py
```diff
@@ -902,3 +907,31 @@ def test_french_accents_search(client, app):
...
+    company = app.data.find_one("companies", req=None, _id=COMPANY_1_ID)
+    assert company
+
+    # add products to user
+    app.data.update(
+        "users", PUBLIC_USER_ID, {"products": [{"section": "wire", "_id": 10}, {"section": "wire", "_id": 11}]}, user
+    )
+
+    # and remove those from company
+    app.data.update("companies", COMPANY_1_ID, {"products": None}, company)
```
MarkLark86, line 925: How can a user have access to a product when the parent company does not? Shouldn't this be a requirement?

petrjasek: in general that is not the case, I've just simplified the setup for the test

petrjasek: updated to test to avoid any confusion

MarkLark86 (approved):

MarkLark86:
@petrjasek Ideally we should plug that permission gap, at a later time maybe!

PR #683, opened 2023-12-06
Agenda search results fix

thecalcc:
NHUB-452

## Checklist

### DIFF
--- a/assets/search/actions.ts
+++ b/assets/search/actions.ts
@@ -8,7 +8,7 @@ import {getNavigationUrlParam, getSearchParams} from './utils';
 import {getLocations, getMapSource} from 'maps/utils';
 
 import {closeModal} from 'actions';
-import {fetchItems, setShareItems} from 'wire/actions';
+import {setShareItems} from 'wire/actions';
 import {createOrUpdateTopic, fetchUser} from 'user-profile/actions';
 
 import {multiSelectTopicsConfigSelector} from 'ui/selectors';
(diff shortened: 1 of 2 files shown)


assets/search/actions.ts
```diff
@@ -22,6 +22,7 @@ import {
 } from './selectors';
 
 import {context} from 'selectors';
+import {fetchItems} from 'agenda/actions';
```
MarkLark86, line 25: Does this now not work in The Wire section? Originally it was calling the `wire/actions/fetchItems` and now its calling the `agenda/actions/fetchItems`. Both those 2 actions do different things.

We would need a better solution here that works for all sections, using dispatching `search/actions/toggleFilter` and `search/actions/setCreatedFilter`

MarkLark86: Maybe we should:
* remove the `fetchItems` from the `toggleFilter` & `setCreatedFilter` dispatch actions
* Add a call to `fetchItems` from where `toggleFilter` & `setCreatedFilter` actions are executed (filter panel etc)

MarkLark86 (changes requested):

assets/search/actions.ts
thecalcc: I checked wire and it's working but yes, I think we need a better solution to clearly define where the search is coming from, I'm just still thinking about it and I'll discuss it with Tomas

MarkLark86: I believe this needs to be fixed sooner rather than later, so we cannot wait for Tomas

MarkLark86: I'm not convinced this is the solution, as the `agenda/actions/fetchItems` runs a query against the Agenda API. So, when a filter is removed from the search results bar, a search against Agenda would happen even in the Wire page

PR #684, opened 2023-12-06
Display duration on Preview and Detail View for a segment of a TV/radio transcript

dzonidoo:
CPCN-31

## Checklist

### DIFF
--- a/assets/wire/components/fields/DurationLabel.tsx
+++ b/assets/wire/components/fields/DurationLabel.tsx
@@ -17,11 +17,11 @@ export function DurationLabel (props: IProps) {
         return null;
     }
 
-    function zeroPad(number: number) {
+    function zeroPad(number: number):string {
         if (number.toString().length < 2) {
             return `0${number}`;
         } else {
-            return number;
+            return number.toString();
         }
     }
 
--- a/assets/wire/components/fields/DurationLabel.tsx
+++ b/assets/wire/components/fields/DurationLabel.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {gettext} from 'utils';
-import {get} from 'lodash';
 
 export function DurationLabel ({item, filterGroupLabels}: any) {
     if (item.extra?.duration == null) {
@@ -28,7 +27,7 @@ export function DurationLabel ({item, filterGroupLabels}: any) {
         return value;
     }
 
-    const label = get(filterGroupLabels, 'duration', gettext('Duration'));
+    const label = filterGroupLabels?.duration ?? gettext('Duration');
     const time = secondsToHHMMSS(item.extra.duration);
     const text = label + ': ' + time;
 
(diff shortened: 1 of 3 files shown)


assets/wire/components/fields/DurationLabel.tsx
```diff
@@ -0,0 +1,43 @@
...
+    
+        return value;
+    }
+
+    const label = get(filterGroupLabels, 'duration', gettext('Duration'));
+    const time = secondsToHHMMSS(item.extra.duration);
+    const text = label + ': ' + time;
+
+    return (
+        <span>{text}</span>
```
petrjasek, line 36: we have some classes on the other fields, shouldn't we use those here as well?

assets/wire/components/fields/DurationLabel.tsx
```diff
@@ -0,0 +1,43 @@
...
+        const h = zeroPad(Math.floor(seconds / 3600));
+        const m = zeroPad(Math.floor(seconds % 3600 / 60));
+        const s = zeroPad(Math.floor(seconds % 3600 % 60));
+
+        const value = `${h}:${m}:${s}`;
+    
+        return value;
+    }
+
+    const label = get(filterGroupLabels, 'duration', gettext('Duration'));
```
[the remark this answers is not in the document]
dzonidoo: so label is just `const label = gettext('Duration');`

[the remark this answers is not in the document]
petrjasek: you could use `filterGroupLabels?.duration ?? : gettext('Duration');`

assets/wire/components/fields/DurationLabel.tsx
dzonidoo: we just use classname for UrgencyLabel in some specific case

assets/wire/components/fields/DurationLabel.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {gettext} from 'utils';
+
+export function DurationLabel ({item, filterGroupLabels}: any) {
```
MarkLark86, line 5: @dzonidoo Please add types as per the PR checklist. This includes adding `duration` to the `IArticle['extra']` (https://github.com/superdesk/newsroom-core/blob/develop/assets/interfaces/content.ts#L26), and avoiding use of `any` and `PropType` and instead use the appropriate `IProps` interface to be used for this component

assets/wire/components/fields/DurationLabel.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function DurationLabel (props: IProps) {
+    const {item, filterGroupLabels} = props;
+
+    if (item.extra?.duration == null) {
+        return null;
+    }
+
+    function zeroPad(number: number) {
```
petrjasek, line 20: imo would be better to make it return string :
`zerePad(number: number) : string {`

petrjasek (approved):

petrjasek (approved):

PR #685, opened 2023-12-06
Display expiry date on Preview and Detail View for a segment of a TV/radio transcript

dzonidoo:
CPCN-30

## Checklist

### DIFF
--- a/assets/wire/components/fields/ExpiryDateLabel.tsx
+++ b/assets/wire/components/fields/ExpiryDateLabel.tsx
@@ -1,15 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import {gettext} from 'utils';
-import {get} from 'lodash';
+import {gettext, formatDate} from 'utils';
 
 export function ExpiryDateLabel ({item, filterGroupLabels}: any) {
     if (item.expiry == null) {
         return null;
     }
 
-    const label = get(filterGroupLabels, 'expiry', gettext('Expiry Date'));
-    const value = item.expiry;
+    const label = filterGroupLabels?.expiry ?? gettext('Expiry Date');
+    const value = formatDate(item.expiry);
     const text = label + ': ' + value;
 
     return (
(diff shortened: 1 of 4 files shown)


assets/wire/components/fields/ExpiryDateLabel.tsx
```diff
@@ -0,0 +1,23 @@
...
+import PropTypes from 'prop-types';
+import {gettext} from 'utils';
+import {get} from 'lodash';
+
+export function ExpiryDateLabel ({item, filterGroupLabels}: any) {
+    if (item.expiry == null) {
+        return null;
+    }
+
+    const label = get(filterGroupLabels, 'expiry', gettext('Expiry Date'));
```
petrjasek, line 11: no get pls

assets/wire/components/fields/ExpiryDateLabel.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const value = item.expiry;
+    const text = label + ': ' + value;
```
petrjasek, line 13: the value should be formatted using `utils.formatDate`

dzonidoo: can u give me more informations? :D

petrjasek: we don't want to display in the UI the server format `2023-12-07T09:53:00` , we have a function for that in utils:
```ts
import {formatDate} from 'utils';
value = formatDate(item.expiry);
```

assets/wire/components/fields/ExpiryDateLabel.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {gettext, formatDate} from 'utils';
+
+export function ExpiryDateLabel ({item, filterGroupLabels}: any) {
```
MarkLark86, line 5: @dzonidoo Please add types as per the PR checklist. This includes adding `expiry` to the `IArticle` (https://github.com/superdesk/newsroom-core/blob/develop/assets/interfaces/content.ts#L13), and avoiding use of `any` and `PropType` and instead use the appropriate `IProps` interface to be used for this component

MarkLark86 (approved):

PR #687, opened 2023-12-07
fix: The date in the downloaded Plain text file is not the event date [CPCN-388]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/template_filters.py
+++ b/newsroom/template_filters.py
@@ -138,29 +138,39 @@ def format_event_datetime(item: dict) -> str:
 
         if is_tbc_item:
             if start.date() != (end.date() if end else None):
-                return lazy_gettext(
-                    f"Event Starts: {formatted_start} - Event Ends: {formatted_end} ({tz}) (Time to be confirmed)"
+                return lazy_gettext("Event starts: {start} - Event Ends: {end} ({tz}) (Time to be confirmed)").format(
+                    start=formatted_start, end=formatted_end, tz=tz
                 )
+
             else:
-                return lazy_gettext(f"Event Starts: {formatted_start} ({tz}) (Time to be confirmed)")
+                return lazy_gettext("Event Starts: {start} ({tz}) (Time to be confirmed)").format(
+                    start=formatted_start, tz=tz
+                )
 
         if start.date() != (end.date() if end else None):
             if all_day or no_end_time and schedule_type in (ScheduleType.REGULAR, ScheduleType.MULTI_DAY):
-                return lazy_gettext(f"Event Starts: {formatted_start} - Event Ends: {formatted_end} ({tz})")
+                return lazy_gettext("Event Starts: {start} - Event Ends: {formatted_end} ({tz})").format(
+                    start=formatted_start, end=formatted_end, tz=tz
+                )
 
         if start.date() == (end.date() if end else None):
             if no_end_time or all_day and schedule_type in (ScheduleType.ALL_DAY, ScheduleType.NO_DURATION):
-                return lazy_gettext(f"Event Starts: {formatted_start} ({tz})")
+                return lazy_gettext("Event Starts: {start} ({tz})").format(start=formatted_start, tz=tz)
 
         if no_end_time and start.date():
-            return lazy_gettext(f"Event Starts: {formatted_start} ({tz})")
+            return lazy_gettext("Event Starts: {start} ({tz})").format(start=formatted_start, tz=tz)
 
         if schedule_type == ScheduleType.REGULAR:
-            return lazy_gettext(
-                f"Event Starts : {formatted_start.split()[1]} to {formatted_end.split()[1]} - On Date: {formatted_start.split()[0]} ({tz})"
+            return lazy_gettext("Event Starts : {start_time} to {end_time} - On Date: {date} ({tz})").format(
+                start_time=formatted_start.split()[1],
+                end_time=formatted_end.split()[1],
+                date=formatted_end.split()[0],
+                tz=tz,
             )
         else:
-            return lazy_gettext(f"Event Starts: {formatted_start} - Event Ends: {formatted_end} ({tz})")
+            return lazy_gettext("Event Starts: {start} - Event Ends: {end} ({tz})").format(
+                start=formatted_start, end=formatted_end, tz=tz
+            )
 
     finally:
         # clear session timezone
--- a/newsroom/template_filters.py
+++ b/newsroom/template_filters.py
@@ -143,15 +143,14 @@ def format_event_datetime(item: dict) -> str:
                     start=formatted_start, tz=tz
                 )
 
-        if end and start.date() != end.date() and schedule_type in (ScheduleType.REGULAR, ScheduleType.MULTI_DAY):
+        if (end and start.date() != end.date()) and schedule_type in (ScheduleType.REGULAR, ScheduleType.MULTI_DAY):
             if all_day or no_end_time:
                 return lazy_gettext("Event Starts: {start} - Event Ends: {end} ({tz})").format(
                     start=formatted_start, end=formatted_end, tz=tz
                 )
 
         if (
-            end
-            and start.date() == end.date()
+            (end and start.date() == end.date())
             or schedule_type in (ScheduleType.ALL_DAY, ScheduleType.NO_DURATION)
             or no_end_time
             or all_day
@@ -163,9 +162,9 @@ def format_event_datetime(item: dict) -> str:
 
         if schedule_type == ScheduleType.REGULAR:
             return lazy_gettext("Event Starts: {start_time} to {end_time} - On Date: {date} ({tz})").format(
-                start_time=formatted_start.split()[1],
-                end_time=formatted_end.split()[1],
-                date=formatted_end.split()[0],
+                start_time=notification_time(start),
+                end_time=notification_time(end),
+                date=notification_date(start),
                 tz=tz,
             )
         else:
--- a/newsroom/template_filters.py
+++ b/newsroom/template_filters.py
@@ -113,13 +113,15 @@ def format_event_datetime(item):
 
             if is_tbc_item:
                 if start.date() != end.date():
-                    return f"Event Starts: {formatted_start} - Event Ends: {formatted_end} (Time to be confirmed)"
+                    return (
+                        f"Event Starts: {formatted_start} - Event Ends: {formatted_end} ({tz}) (Time to be confirmed)"
+                    )
                 else:
-                    return f"Event Starts: {formatted_start} (Time to be confirmed)"
+                    return f"Event Starts: {formatted_start} ({tz}) (Time to be confirmed)"
 
             if start.date() != end.date():
                 if all_day or no_end_time or schedule_type in ("REGULAR", "MULTI_DAY"):
-                    return f"Event Starts: {formatted_start} - Event Ends: {formatted_end}"
+                    return f"Event Starts: {formatted_start} - Event Ends: {formatted_end} ({tz})"
 
             if start.date() == end.date():
                 if no_end_time or all_day or schedule_type in ("ALL_DAY", "NO_DURATION"):
--- a/newsroom/template_filters.py
+++ b/newsroom/template_filters.py
@@ -8,19 +8,27 @@ import logging
 
 from flask import current_app as app
 from eve.utils import str_to_date
-from flask_babel import format_time, format_date, format_datetime, get_locale
+from flask_babel import format_time, format_date, format_datetime, get_locale, lazy_gettext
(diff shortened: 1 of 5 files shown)


newsroom/templates/download_agenda.txt
```diff
@@ -6,7 +6,7 @@
 {% endif %}
 {%- if item.definition_long %}Definition: {{ item.definition_long }}
 {% endif -%}
-On: {{ item.dates.start | datetime_long }}
+On: {{ (item.dates.end or item.dates.start) | datetime_long }}
```
petrjasek, line 9: shouldn't this be rather the other way around?

petrjasek: it's not showing that because of the end time but due to the timezone, it starts at 18:30 and the timezone is +5:30, so it makes that start being on 18th at 00:00 local time (which is what we display in the UI)

devketanpro: I see, I will update it

newsroom/template_filters.py
```diff
@@ -61,6 +62,78 @@ def parse_date(datetime):
...
+    return event_tbc
+
+
+def format_event_datetime(item):
+    datetime = item.get("dates", {})
+    if datetime:
+        tz = datetime.get("tz", get_session_timezone())
+        try:
+            # Set the session timezone
+            set_session_timezone(tz)
```
petrjasek, line 101: I'm not sure but afaik babel caches it so you might have to call refresh after changing it

newsroom/template_filters.py
```diff
@@ -61,6 +62,78 @@ def parse_date(datetime):
...
+                return f"Event Starts: {formatted_start} - Event Ends:{formatted_end} ({tz})"
+            elif no_end_time or schedule_type == "NO_DURATION":
+                return f"Event Starts: {formatted_start} ({tz})"
+            elif schedule_type == "REGULAR":
+                return f"Event Starts: {formatted_start} - Event Ends:{formatted_end} ({tz})"
+            elif schedule_type == "MULTI_DAY":
+                return f"Event Starts :{formatted_start} - Event Ends: {formatted_end} ({tz})"
+            else:
+                print("Not sure about the datetime format", item, schedule_type)
+                return f"Event Starts: {formatted_start} - Event Ends: {formatted_end} ({tz})"
```
petrjasek, line 131: seems like this case is used 4x, so imo we don't need that many ifs

petrjasek (commented):
pls add tests for the format_event_datetime all cases ideally

newsroom/template_filters.py
```diff
@@ -61,6 +62,78 @@ def parse_date(datetime):
     return datetime
 
 
+def get_schedule_type(start, end, all_day, no_end_time):
```
MarkLark86, line 65: Please use Python types for all new code

newsroom/template_filters.py
```diff
@@ -61,6 +62,78 @@ def parse_date(datetime):
...
+    duration = int((end - start).total_seconds() / 60)  # Convert seconds to minutes
+    DAY_IN_MINUTES = 24 * 60 - 1
+
+    if all_day:
+        return "ALL_DAY" if duration == 0 else "MULTI_DAY"
```
MarkLark86, line 70: Seems like an enum could be used for these values, rather than static strings

newsroom/template_filters.py
```diff
@@ -61,6 +62,78 @@ def parse_date(datetime):
...
+def is_item_tbc(item):
+    event_tbc = item.get("event", {}).get("_time_to_be_confirmed", False)
+    if not event_tbc:
+        planning = item.get("planning_items", [])
+        return planning[0].get("_time_to_be_confirmed", False) if planning else False
+    return event_tbc
+
+
+def format_event_datetime(item):
+    datetime = item.get("dates", {})
```
MarkLark86, line 96: The `datetime` name is already defined by the import statement `from datetime import datetime`. It would be a good idea to use a unique name for this variable

newsroom/template_filters.py
```diff
@@ -61,6 +62,78 @@ def parse_date(datetime):
...
+    if datetime:
```
MarkLark86, line 97: Rather than encasing the entire function in an if statement, you could return if `item.get("dates"]` is None

newsroom/template_filters.py
```diff
@@ -61,6 +62,78 @@ def parse_date(datetime):
...
+            all_day = datetime.get("all_day")
+            no_end_time = datetime.get("no_end_time")
+            is_tbc_item = is_item_tbc(item)
+            schedule_type = get_schedule_type(start, end, all_day, no_end_time)
+
+            formatted_start = datetime_long(start)
+            formatted_end = datetime_long(end)
+
+            if is_tbc_item and start != end:
+                return f"Event Starts: {formatted_start} - Event Ends: {formatted_end} (Time to be confirmed)"
```
MarkLark86, line 114: It would be good to use gettext here

newsroom/template_filters.py
devketanpro: I checked but IMO, we don't need to call refresh

newsroom/template_filters.py
```diff
@@ -61,6 +70,84 @@ def parse_date(datetime):
...
+
+def is_item_tbc(item: dict) -> bool:
+    event_tbc = item.get("event", {}).get("_time_to_be_confirmed", False)
+    if not event_tbc:
+        planning = item.get("planning_items", [])
+        return planning[0].get("_time_to_be_confirmed", False) if planning else False
+    return event_tbc
+
+
+def format_event_datetime(item: dict) -> Optional[str]:
```
MarkLark86, line 103: As this is a jinja formatter, we should probably return an empty string instead of `None`

newsroom/template_filters.py
```diff
@@ -61,6 +70,84 @@ def parse_date(datetime):
...
+        all_day = date_info.get("all_day")
+        no_end_time = date_info.get("no_end_time")
+        is_tbc_item = is_item_tbc(item)
+
+        schedule_type = get_schedule_type(start, end, all_day, no_end_time)
+
+        formatted_start = datetime_long(start)
+        formatted_end = datetime_long(end)
+
+        if is_tbc_item:
```
MarkLark86, line 125: It seems like you're not covering the same scenarios as in the front-end. Front end has 9 different texts it can produce, this one only has 5 (see https://github.com/superdesk/newsroom-core/blob/develop/assets/agenda/utils.ts#L1140).

The front-end also has different outputs that contain the time only portion of the date, such as this one: https://github.com/superdesk/newsroom-core/blob/develop/assets/agenda/utils.ts#L1177 `{{startTime}} - {{endTime}} {{startDate}} {{timezone}}`

petrjasek (changes requested):

newsroom/template_filters.py
```diff
@@ -138,29 +138,39 @@ def format_event_datetime(item: dict) -> str:
...
+                return lazy_gettext("Event Starts: {start} ({tz})").format(start=formatted_start, tz=tz)
 
         if no_end_time and start.date():
-            return lazy_gettext(f"Event Starts: {formatted_start} ({tz})")
+            return lazy_gettext("Event Starts: {start} ({tz})").format(start=formatted_start, tz=tz)
 
         if schedule_type == ScheduleType.REGULAR:
-            return lazy_gettext(
-                f"Event Starts : {formatted_start.split()[1]} to {formatted_end.split()[1]} - On Date: {formatted_start.split()[0]} ({tz})"
+            return lazy_gettext("Event Starts : {start_time} to {end_time} - On Date: {date} ({tz})").format(
```
petrjasek, line 164: we should be consistent with the spaces around `:`

devketanpro: my bad

newsroom/template_filters.py
```diff
@@ -61,6 +70,113 @@ def parse_date(datetime):
...
+        is_tbc_item = is_item_tbc(item)
+
+        schedule_type = get_schedule_type(start, end, all_day, no_end_time)
+
+        formatted_start = datetime_long(start)
+        formatted_end = datetime_long(end) if end else None
+
+        if is_tbc_item:
+            if start.date() != (end.date() if end else None):
+                return lazy_gettext("Event Starts: {start} - Event Ends: {end} ({tz}) (Time to be confirmed)").format(
```
petrjasek, line 141: I wonder how would this part work if end is `None`

newsroom/template_filters.py
```diff
@@ -61,6 +70,113 @@ def parse_date(datetime):
...
+
+    if not date_info:
+        return ""
+
+    tz = date_info.get("tz", get_session_timezone())
+    try:
+        # Set the session timezone
+        set_session_timezone(tz)
+
+        start = parse_date(format_datetime(parse_date(date_info.get("start")), "yyyy-MM-dd HH:mm:ss"))
```
petrjasek, line 124: I wonder why we need to parse it 2x?

newsroom/template_filters.py
```diff
@@ -61,6 +70,113 @@ def parse_date(datetime):
...
+        if start.date() == (end.date() if end else None):
+            if no_end_time or all_day and schedule_type in (ScheduleType.ALL_DAY, ScheduleType.NO_DURATION):
+                return lazy_gettext("Event Starts: {start} ({tz})").format(start=formatted_start, tz=tz)
+
+        if no_end_time and start.date():
+            return lazy_gettext("Event Starts: {start} ({tz})").format(start=formatted_start, tz=tz)
+
+        if schedule_type == ScheduleType.REGULAR:
+            return lazy_gettext("Event Starts: {start_time} to {end_time} - On Date: {date} ({tz})").format(
+                start_time=formatted_start.split()[1],
```
petrjasek, line 165: this assumes specific format, but in different locales time might be before the date. we should use configured time format here

newsroom/template_filters.py
devketanpro: my bad

newsroom/template_filters.py
devketanpro: I see, but we used `'DATETIME_FORMAT_LONG'` here, which is probably the same for different locales, isn't it?

petrjasek: it's different, and there might be other things besides date and time in it like `10:00 le 20.12.2023` for fr

devketanpro: I see, can we use these format `TIME_FORMAT_SHORT` for time  and `DATE_FORMAT_SHORT` for date  ?

newsroom/template_filters.py
```diff
@@ -61,6 +70,113 @@ def parse_date(datetime):
...
+                return lazy_gettext("Event Starts: {start} - Event Ends: {end} ({tz}) (Time to be confirmed)").format(
+                    start=formatted_start, end=formatted_end, tz=tz
+                )
+
+            else:
+                return lazy_gettext("Event Starts: {start} ({tz}) (Time to be confirmed)").format(
+                    start=formatted_start, tz=tz
+                )
+
+        if (end and start.date() != end.date()) and schedule_type in (ScheduleType.REGULAR, ScheduleType.MULTI_DAY):
```
petrjasek, line 146: if you use the schedule enum imo it should only use that after to figure out how to present it, but using again the end and start here is imo wrong.
would be good also to check that all schedule [cases are covered](https://typing.readthedocs.io/en/latest/source/unreachable.html#assert-never-and-exhaustiveness-checking)

devketanpro: got it

newsroom/template_filters.py
```diff
@@ -61,6 +70,110 @@ def parse_date(datetime):
...
+            return lazy_gettext("Event Starts: {start_time} to {end_time} - On Date: {date} ({tz})").format(
+                start_time=notification_time(start),
+                end_time=notification_time(end),
+                date=notification_date(start),
+                tz=tz,
+            )
+        else:
+            # Default
+            return lazy_gettext(
+                "Event Starts: {start_time} {start_date} to Event Ends: {end_time} {end_date} ({tz})"
```
petrjasek, line 163: I'm not sure about the `Event starts: ... to Event Ends: ...`, what about:
```
On: {start_time} {start_date} to {end_time} {end_date}
```
or maybe using `Date:` as a prefix? what do you think @MarkLark86

devketanpro: makes sense

MarkLark86: Yeah I agree @petrjasek, maybe just `On: ` makes more sense as it may not always be an Event right!

devketanpro: should we only change the default case, or should we change them all?

petrjasek: all of them

petrjasek (approved):

PR #689, opened 2023-12-08
Fix : Reports that are not working [NHUB-338]

devketanpro:
In this PR, I updated reports `Products per company, Company, Companies per product`.

## Checklist

### DIFF
--- a/newsroom/reports/reports.py
+++ b/newsroom/reports/reports.py
@@ -87,28 +87,24 @@ def get_user_saved_searches():
 def get_company_products():
     """Returns products by company"""
     results = []
-    company_products = {}
     companies = get_entity_dict(query_resource("companies"))
-    products = query_resource("products")
+    products_data = list(query_resource("products"))
+    for company_id, details in companies.items():
+        products = details.get("products", [])
+        company_result = {
+            "_id": str(company_id),
+            "name": details.get("name", ""),
+            "is_enabled": details.get("is_enabled", ""),
+            "products": [],
+        }
+        for product_info in products:
+            product_id = product_info.get("_id")
+            product_data = next((p for p in products_data if p["_id"] == product_id), None)
+            if product_data:
+                product_result = deepcopy(product_data)
+                company_result["products"].append(product_result)
 
-    for product in products:
-        for company_id in product.get("companies", []):
-            company_product = company_products.get(company_id, {})
-            products = company_product.get("products", [])
-            products.append(product)
-            company_product["products"] = products
-            company_products[company_id] = company_product
-
-    for _id, details in company_products.items():
-        if companies.get(_id):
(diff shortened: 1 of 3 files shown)


newsroom/reports/reports.py
```diff
@@ -87,28 +87,24 @@ def get_user_saved_searches():
...
     companies = get_entity_dict(query_resource("companies"))
-    products = query_resource("products")
+    products_data = list(query_resource("products"))
+    for company_id, details in companies.items():
+        products = details.get("products", [])
+        company_result = {
+            "_id": str(company_id),
+            "name": details.get("name", ""),
+            "is_enabled": details.get("is_enabled", ""),
+            "products": [],
```
MarkLark86, line 98: Then you can have the following here instead:
```python
{
  ...,
  "products": [
    product
    for product_id, product in products_data.items()
    if product_id in details.get("products", [])
  ]
}
```

newsroom/reports/reports.py
```diff
@@ -87,28 +87,24 @@ def get_user_saved_searches():
 def get_company_products():
     """Returns products by company"""
     results = []
-    company_products = {}
     companies = get_entity_dict(query_resource("companies"))
-    products = query_resource("products")
+    products_data = list(query_resource("products"))
+    for company_id, details in companies.items():
```
MarkLark86, line 92: `details` here doesn't describe what entity this is talking about. Maybe renaming it to something like `company` would be more meaningful

newsroom/reports/reports.py
```diff
@@ -140,33 +136,29 @@ def get_product_stories():
 def get_company_report():
     """Returns products by company"""
     results = []
-    company_products = {}
-    companies = get_entity_dict(query_resource("companies"))
-    products = query_resource("products")
+    companies = list(query_resource("companies"))
+    products_data = get_entity_dict(query_resource("products"))
```
MarkLark86, line 140: Same as above, using a map of ProductID -> Product would help here

newsroom/reports/reports.py
```diff
@@ -379,6 +371,18 @@ def get_product_company():
     return results
 
 
+def get_companies_id_by_product(product_id):
```
MarkLark86, line 374: Add types to function signature (params & response types)

MarkLark86, line 374: This might be better placed in `newsroom.companies.companies`, in case it might be needed elsewhere later

newsroom/reports/reports.py
devketanpro: here we don't need because of `get_entity_dict `. this function returns in that format
 `{product["_id"]: product for product in products}`

MarkLark86 (approved):

PR #696, opened 2023-12-12
Show products that are for the current company only

thecalcc:
CPCN-449
## Checklist

### DIFF
--- a/assets/company-admin/components/CompanyDetailsProductFilter.tsx
+++ b/assets/company-admin/components/CompanyDetailsProductFilter.tsx
@@ -8,7 +8,10 @@ import {currentCompanySelector, companySectionListSelector} from '../selectors';
 import DropdownFilter from 'components/DropdownFilter';
 
 interface IReduxProps {
-    currentCompany: any;
+    currentCompany: {
+        _id: string;
+        products: Array<{_id: string}>;
+    };
     companySections: any;
 }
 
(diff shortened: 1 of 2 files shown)


thecalcc:
Also fix user creation side panel breaking when no company has been created.

assets/company-admin/components/CompanyDetailsProductFilter.tsx
```diff
@@ -29,9 +29,11 @@ class CompanyDetailsProductFilterComponent extends React.PureComponent<any, any>
 
     getDropdownItems(filter: any) {
         const sectionIds = this.props.companySections[this.props.currentCompany._id].map((section: any) => section._id);
+        const currentCompanyProducts: Array<string> = this.props.currentCompany.products.map(({_id}: any) => _id);
 
         return this.props.products
             .filter((product: any) => sectionIds.includes(product.product_type))
+            .filter((product: any) => currentCompanyProducts.includes(product._id))
```
tomaskikutis, line 36: What about doing inside the existing filter? :D

I'd be also better to use a Set instead of an Array for `currentCompanyProducts` and `sectionIds` so we avoid iterations.

assets/company-admin/components/CompanyDetailsProductFilter.tsx
```diff
@@ -29,9 +29,10 @@ class CompanyDetailsProductFilterComponent extends React.PureComponent<any, any>
 
     getDropdownItems(filter: any) {
         const sectionIds = this.props.companySections[this.props.currentCompany._id].map((section: any) => section._id);
+        const currentCompanyProducts: Set<string> = new Set(this.props.currentCompany.products.map(({_id}: any) => _id));
```
MarkLark86, line 32: If `<component>.propTypes` and `React.PureComponent<any, any>` was replaced with ts interface, this would remove the need for `any` in new code (there's not many props for this one).

thecalcc: Good point, forgot that we're converting everything

assets/company-admin/components/CompanyDetailsProductFilter.tsx
tomaskikutis: can you convert `sectionIds` to a set too?

tomaskikutis (approved):

PR #697, opened 2023-12-12
Fix location from ingested items display

thecalcc:
STTNHUB-187
## Checklist

### DIFF
--- a/assets/agenda/utils.ts
+++ b/assets/agenda/utils.ts
@@ -235,14 +235,19 @@ export function getGeoLocation(item: any) {
  * @return {String}
  */
 export function getLocationString(item: any) {
-    return [
-        get(item, 'location.0.name', get(item, 'location.0.address.title')),
-        get(item, 'location.0.address.line.0'),
-        get(item, 'location.0.address.city') || get(item, 'location.0.address.area'),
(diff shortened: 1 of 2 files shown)


thecalcc:
Mapped & trimmed strings so all strings are consistent after joining.

BEFORE: 
image

AFTER:
image

assets/agenda/utils.ts
```diff
@@ -235,14 +235,19 @@ export function getGeoLocation(item: any) {
...
 export function getLocationString(item: any) {
-    return [
-        get(item, 'location.0.name', get(item, 'location.0.address.title')),
-        get(item, 'location.0.address.line.0'),
-        get(item, 'location.0.address.city') || get(item, 'location.0.address.area'),
-        get(item, 'location.0.address.state') || get(item, 'location.0.address.locality'),
-        get(item, 'location.0.address.postal_code'),
-        get(item, 'location.0.address.country'),
-    ].filter((d: any) => d).join(', ');
+    const location = item.location?.[0];
```
MarkLark86, line 238: Similar changes would need to be done on the back-end too: https://github.com/superdesk/newsroom-core/blob/develop/newsroom/utils.py#L217 (this is mainly for emails)

tomaskikutis (approved):

PR #698, opened 2023-12-12
If a user is being created don't show the resend invite button

thecalcc:
NHUB-461
## Checklist


assets/users/components/EditUser.tsx
```diff
@@ -112,7 +112,7 @@ const EditUserComponent: React.ComponentType<IProps> = (props: IProps) => {
     const isCompanyAdmin = isUserCompanyAdmin(currentUser);
     const company = companies.find((c) => c._id === user.company);
     const userIsAdmin = isUserAdmin(user);
-    const showResendInvite = (user._id == null || user.is_validated !== true) && (
+    const showResendInvite = original?._id != null && (user._id == null || user.is_validated !== true) && (
```
tomaskikutis, line 115: I'd be good to extract it to a separate constant for documentation

```
const newUserIsBeingCreated = original?._id == null;
```

tomaskikutis (approved):

PR #705, opened 2023-12-19
FIX : Real Notifications are not saved on revisiting when switching from Company topics to My topics [CPCN-413]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/topics/topics.py
+++ b/newsroom/topics/topics.py
@@ -81,17 +81,12 @@ class TopicsService(newsroom.Service):
 
         # If ``is_global`` has been turned off, then remove all subscribers
         # except for the owner of the Topic
-        if (
-            original.get("is_global")
-            and "is_global" in updates
-            and not updates.get("is_global")
-            and not updates.get("subscribers")
-        ):
+        if original.get("is_global") and "is_global" in updates and not updates.get("is_global"):
             # First find the subscriber entry for the original user
             subscriber = next(
                 (
                     subscriber
-                    for subscriber in original.get("subscribers", [])
+                    for subscriber in updates.get("subscribers", [])
                     if subscriber["user_id"] == original["user"]
                 ),
                 None,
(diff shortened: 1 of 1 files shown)


petrjasek:
how is it fixing the issue? can you add some description

devketanpro:
>how is it fixing the issue? can you add some description

When a user wants to set `'is_global' = false` **(which means unchecking Share with My Company)** and `subscriber config` **(which means notification)** at the same time, updates of subscribers are not updated because of this capability **(If the ''is_global'' option is set to false, all subscribers are erased)**, which has already been implemented.
So I updated the logic; now, if the user tries to set `'is_global'` to `false` and updates contain subscriber data, subscriber data is successfully stored.

petrjasek:
but wouldn't it keep the other subscribers too in this case?

devketanpro:
> but wouldn't it keep the other subscribers too in this case?

other subscribers? Sorry for missing it. Could you please elaborate?

newsroom/topics/topics.py
```diff
@@ -81,17 +81,12 @@ def on_update(self, updates, original):
...
-            and not updates.get("is_global")
-            and not updates.get("subscribers")
-        ):
+        if original.get("is_global") and "is_global" in updates and not updates.get("is_global"):
             # First find the subscriber entry for the original user
             subscriber = next(
                 (
                     subscriber
-                    for subscriber in original.get("subscribers", [])
+                    for subscriber in updates.get("subscribers", [])
```
petrjasek: not sure it will work now if subscribers are not updated, imo should be like:
```py
for subscriber in (updates.get("subscribers") or original.get('subscribers") or []):
```

devketanpro: I added this :- `for subscriber in updates.get("subscribers", original.get("subscribers", []))`

petrjasek: I see, it's not the same because if you would have `updates["subscribers"] = None` it would raise an error

devketanpro: correct

petrjasek (approved):

PR #708, opened 2023-12-21
Fix folder refresh in the side panel

thecalcc:
NHUB-454
## Checklist

### DIFF
--- a/assets/user-profile/index.ts
+++ b/assets/user-profile/index.ts
@@ -3,11 +3,16 @@ import {initWebSocket} from 'websocket';
 import UserProfileApp from './components/UserProfileApp';
 import {initData, selectMenu, pushNotification} from './actions';
 import {store} from './store';
-import {WIRE_TOPIC_FOLDERS_UPDATED} from 'wire';
 
 if (window.profileData) {
     store.dispatch(initData(window.profileData));
 }
+
+/**
+ * This constant needs to be defined here because of script loading.
+ * If not it will break the tests because of the loading order.
+ */
+export const WIRE_TOPIC_FOLDERS_UPDATED = 'reload-wire-folders';
 
 let previousState = store.getState();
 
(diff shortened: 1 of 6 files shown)


assets/wire/actions.ts
```diff
@@ -506,6 +507,36 @@ export function reloadMyTopics(reloadTopic: any = false) {
     };
 }
 
+export const UPDATE_FOLDERS = 'UPDATE_FOLDERS';
+
+/**
+ * @param {bool} global - fetch company or user folders
+ * @param {bool} skipDispatch - if true it won't replace folders in store
+ */
```
tomaskikutis, line 515: it looks like comments from some other function? where are the params that are being commented?

assets/wire/actions.ts
```diff
@@ -506,6 +507,36 @@ export function reloadMyTopics(reloadTopic: any = false) {
...
+            server.get(userTopicsUrl).then(({_items}: {_items: Array<any>}) => _items),
+        ]).then(([companyFolders, userFolders]) => {
+            dispatch({
+                type: UPDATE_FOLDERS,
+                payload: {
+                    companyFolders: companyFolders,
+                    userFolders: userFolders,
+                },
+            });
+        }).catch((error) => {
```
tomaskikutis, line 533: is there a point to catch it at all here if the only thing you're doing is printing it to the console? Wouldn't it log automatically if promise fails?

tomaskikutis:
what about the checklist?

assets/user-profile/index.ts
```diff
@@ -3,22 +3,30 @@ import {initWebSocket} from 'websocket';
...
 import {initData, selectMenu, pushNotification} from './actions';
 import {store} from './store';
-import {WIRE_TOPIC_FOLDERS_UPDATED} from 'wire';
 
 if (window.profileData) {
     store.dispatch(initData(window.profileData));
 }
 
+/**
+ * This constant needs to be defined here because of script loading.
```
tomaskikutis, line 12: > This constant needs to be defined here

"Here" as opposed to where? Keep in mind when you're writing comments that whoever will read it, will not know about the PR that added the comment. When possible, write it in a way that is understandable by having access only to current commit. Like "this constant needs to be define here, and not in /path/file.ts". If the PR is needed to understand the comment - link to the PR.

tomaskikutis (approved):

PR #709, opened 2023-12-21
Wire list card implementation

thecalcc:
TGA-73
## Checklist

### DIFF
--- a/assets/components/cards/render/WireListCard.tsx
+++ b/assets/components/cards/render/WireListCard.tsx
@@ -12,10 +12,11 @@ import {listConfigSelector} from 'ui/selectors';
 import {DEFAULT_COMPACT_META_FIELDS} from 'wire/components/WireListItem';
 import {IArticle, IListConfig} from 'interfaces';
 import {filterGroupsToLabelMap} from 'search/selectors';
-import {setActive, openItem} from 'wire/actions';
 
 interface IOwnProps {
     item: IArticle;
+    openItem: (item: IArticle, cardId: string) => void;
+    cardId: string;
 }
 
 interface IPropsReduxStore {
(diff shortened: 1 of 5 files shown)


thecalcc:
image

assets/components/cards/render/WireListCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+    render(): React.ReactNode {
+        const cardClassName = classNames('wire-articles__item-wrap col-12 wire-item');
+        const wrapClassName = classNames('wire-articles__item wire-articles__item--wire wire-articles__item--list');
+        const listImage = getImageForList(this.props.item);
+        const compactFields = this.props.listConfig.compact_metadata_fields || DEFAULT_COMPACT_META_FIELDS;
+        const {item, listConfig} = this.props;
+        const isExtended = false;
+
+        // FIXME: Implement the proper subscription overlay
+        const showOverlay = () => showModal(({closeModal}) => {
```
MarkLark86, line 48: This dashboard card should behave in the same way that all other cards behave, i.e. when clicking it should open the wire item in full page preview.

At this present stage, none of the dashboard cards should have any code for the public dashboard feature we will be building.

thecalcc: Ok will change this

assets/components/cards/render/WireListCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    render(): React.ReactNode {
+        const cardClassName = classNames('wire-articles__item-wrap col-12 wire-item');
+        const wrapClassName = classNames('wire-articles__item wire-articles__item--wire wire-articles__item--list');
+        const listImage = getImageForList(this.props.item);
+        const compactFields = this.props.listConfig.compact_metadata_fields || DEFAULT_COMPACT_META_FIELDS;
+        const {item, listConfig} = this.props;
+        const isExtended = false;
+
+        const onItemClick = () => {
```
MarkLark86, line 48: I can see that other Card components use the [ICardProps](https://github.com/superdesk/newsroom-core/blob/develop/assets/components/cards/utils.ts#L40) interface, which provides an `openItem` prop (among others).

It would make it easier to control the functionality if these props were used instead of directing importing/executing the wire actions. This flexibility would be required when reusing these Card components for the public dashboard

MarkLark86 (approved):

PR #711, opened 2023-12-21
Button component

dzonidoo:
NHUB-464

## Checklist


petrjasek:
hi @dzonidoo can you update the PR so we can merge it?

assets/components/Buttons.tsx
petrjasek, line 1: would rename this so it matches the component name like all the other files

assets/components/Buttons.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+export type IPropsVariant = 'primary' | 'secondary' | 'tertiary';
+export type IPropsSize = 'small' | 'medium' | 'large';
+
+export interface IPropsButtonBase {
+    value: string;
```
petrjasek: imo would be better to work with children so it's more like regular button, or call this label because there is no `value` really like on an input

assets/components/Buttons.tsx
```diff
@@ -0,0 +1,58 @@
...
+    onClick(event: React.MouseEvent<HTMLButtonElement, MouseEvent>): void;
+}
+
+interface IPropsSubmit extends IPropsButtonBase {
+    type?: 'submit',
+    onClick?(event: React.MouseEvent<HTMLButtonElement, MouseEvent>): void;
+}
+
+interface IPropsReset extends IPropsButtonBase {
+    type?: 'reset',
```
petrjasek: I guess here the type must be `reset` so should be probably just `type: 'reset'`,
I'm also wondering if we have a usecase where we wouldn't use it with `onClick`, might be better to make it mandatory if not to get this checked.
same applies to submit

assets/components/CloseModalButton.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {gettext} from 'utils';
+
+function CloseModalButton({onClick}: any) {
```
petrjasek, line 5: I wonder what's the usecase for having this one and `CloseButton` which seems to handle modals as well

dzonidoo: this is the iconButton and we still don't have component for that... we will make it soon

assets/components/Buttons.tsx
dzonidoo: i guess 'text' is better name that 'label' in this case?

assets/components/Button.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+interface IPropsButton extends IPropsButtonBase {
+    type?: 'button',
+    onClick(event: React.MouseEvent<HTMLButtonElement, MouseEvent>): void;
+}
+
+interface IPropsSubmit extends IPropsButtonBase {
+    type: 'submit',
+    onClick?(event: React.MouseEvent<HTMLButtonElement, MouseEvent>): void;
```
petrjasek: I would still make this mandatory, if we need to put there empty function in case this is not needed it's probably better than forgetting it when it's needed. same goes for reset button

assets/components/ActionButton.tsx
```diff
@@ -1,43 +1,30 @@
...
-
-    componentWillUnmount() {
-        if (this.elem && this.tooltip) {
-            this.tooltip.dispose();
-        }
-    }
+import {IconButton, IPropsSize, IPropsVariant} from './IconButton';
+
+interface IProps {
+    testId?: string,
```
petrjasek: I see this is `data-test-id` on other components, would be good to make it consistent

assets/agenda/components/AgendaListItem.tsx
```diff
@@ -385,9 +385,8 @@ class AgendaListItem extends React.Component<IProps> {
                 {!this.props.showShortcutActionIcons ? null : this.props.actions.map((action) => action.shortcut && (
                     <ActionButton
                         key={action.name}
-                        className="icon-button icon-button--primary"
+                        variant='primary'
                         action={action}
-                        isVisited={action.visited && action.visited(this.props.user, this.props.item)}
```
petrjasek, line 390: why are we dropping this functionality?

assets/components/ActionButton.tsx
dzonidoo: so change the name?

petrjasek: yes

dzonidoo:
We can merge this now. @fritzSF can confirm that we are dropping this functionality.

petrjasek (approved):

PR #712, opened 2023-12-25
The name and email of the assigned desk or journalist should be displayed in all views [CPCN-503]

devketanpro:
## Checklist

### DIFF
--- a/assets/agenda/components/AgendaCoverages.tsx
+++ b/assets/agenda/components/AgendaCoverages.tsx
@@ -47,79 +47,83 @@ function AgendaCoveragesComponent({item, coverages, wireItems, actions, user, on
                     ))}
                 </div>
             )}
-            {coveragesWithState.map((coverage: any) => (
-                <div
-                    className={classNames(
-                        'coverage-item',
-                        {'coverage-item--clickable': onClick}
-                    )}
-                    key={coverage.coverage_id}
-                    onClick={onClick}
-                    title={onClick ? gettext('Open {{agenda}} in a new tab', window.sectionNames) : onClick}
-                >
+            {coveragesWithState.map((coverage: any) => {
+                const assigneeName = getCoverageAsigneeName(coverage);
+                const deskName = getCoverageDeskName(coverage);
+                return(
                     <div
-                        className='coverage-item__row coverage-item__row--header-row'
-                        title={getCoverageTooltip(coverage)}
+                        className={classNames(
+                            'coverage-item',
+                            {'coverage-item--clickable': onClick}
+                        )}
+                        key={coverage.coverage_id}
+                        onClick={onClick}
+                        title={onClick ? gettext('Open {{agenda}} in a new tab', window.sectionNames) : onClick}
                     >
-                        <span className={classNames('coverage-item__coverage-icon', WORKFLOW_COLORS[coverage.workflow_status])}>
-                            <i className={`icon--coverage-${getCoverageIcon(coverage.coverage_type)}`}></i>
-                        </span>
-
-                        <span className='coverage-item__coverage-heading'>
-                            <span className='fw-medium'>
-                                {`${(coverage.genre?.length ?? 0) > 0 ? gettext(coverage.genre[0].name) : getCoverageDisplayName(coverage.coverage_type)}`}
-                            </span>
-                            {`${getSlugline(coverage)}`}
-                        </span>
-
-                    </div>
-                    {<div className='coverage-item__row align-items-center'>
-                        {getCoverageAsigneeName(coverage) && (
-                            <span className='d-flex text-nowrap pe-1'>
-                                <span className='coverage-item__text-label me-1'>{gettext('assignee')}:</span>
-                                <span>{getCoverageAsigneeName(coverage)}</span>
-                            </span>
-                        )}
-                        {getCoverageAsigneeName(coverage) && getCoverageDeskName(coverage) && ' | '}
-                        {getCoverageDeskName(coverage) && (
-                            <span className='d-flex text-nowrap ps-1'>
-                                <span className='coverage-item__text-label me-1'>{gettext('desk')}:</span>
-                                <span className=''>{getCoverageDeskName(coverage)}</span>
-                            </span>
-                        )}
-                    </div>
-                    }
-                    {coverage.workflow_status !== WORKFLOW_STATUS.COMPLETED && coverage.scheduled != null && (
                         <div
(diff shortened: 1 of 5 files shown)


assets/agenda/utils.ts
```diff
@@ -186,6 +186,14 @@ export function getCoverageDisplayName(coverageType: any) {
         get(coverageTypes, `${coverageType}.name`, coverageType);
 }
 
+export function getCoverageAsigneeName(coverage: any) {
```
MarkLark86, line 189: Please use types, `any` should **not** be used in new code, this includes function arguments and return type.

assets/agenda/utils.ts
```diff
@@ -1015,42 +1023,54 @@ export const getCoverageTooltip = (coverage: any, beingUpdated?: any) => {
     const slugline = coverage.item_slugline || coverage.slugline;
     const coverageType = getCoverageDisplayName(coverage.coverage_type);
     const coverageScheduled = moment(coverage.scheduled);
+    const assignee = getCoverageAsigneeName(coverage) || '';
+    const desk = getCoverageDeskName(coverage) || '';
 
     if (coverage.workflow_status === WORKFLOW_STATUS.DRAFT) {
-        return gettext('{{ type }} coverage {{ slugline }} {{ status_text }}', {
+        return gettext('{{ type }} coverage {{ slugline }} {{ status_text }}, assignee: {{assignee}}, desk: {{desk}}', {
```
MarkLark86, line 1030: How does this look when `assignee` or `desk` is not actually defined? Would this needlessly display
`assignee: , desk: `

assets/agenda/components/AgendaCoverages.tsx
```diff
@@ -71,6 +73,22 @@ function AgendaCoveragesComponent({item, coverages, wireItems, actions, user, on
                         </span>
 
                     </div>
+                    {<div className='coverage-item__row align-items-center'>
+                        {getCoverageAsigneeName(coverage) && (
```
MarkLark86, line 77: `getCoverageAsigneeName`. `getCoverageAsigneeName` and `getCoverageDeskName` are called multiple times for the 1 render. It would be better to store the results of these functions and use the stored response instead.

Also, if none of those 3 responses have a valid value, then `<div className='coverage-item__row align-items-center'>` should not be rendered

MarkLark86 (changes requested):
In the Preview and Detail views, where are the clickable `mailto:` link that opens a new email message in the user's default email client (as per the ticket's A.C.)?

assets/agenda/components/AgendaCoverages.tsx
devketanpro: yep sounds identical

assets/agenda/components/AgendaCoverages.tsx
```diff
@@ -45,63 +48,89 @@ function AgendaCoveragesComponent({item, coverages, wireItems, actions, user, on
...
-                    key={coverage.coverage_id}
-                    onClick={onClick}
-                    title={onClick ? gettext('Open {{agenda}} in a new tab', window.sectionNames) : onClick}
-                >
+            {coveragesWithState.map((coverage: ICoverage) => {
+                const assigneeName = getCoverageAsigneeName(coverage);
+                const deskName = getCoverageDeskName(coverage);
+                const assignedUserEmail = coverage.assigned_user_email;
+                const assignedDeskEmail = coverage.assigned_desk_email;
+                const subject = `Coverage inquiry from NewsPro user: ${item.name || item.slugline}`;
```
MarkLark86, line 56: `gettext` must be used to flag this for translation.

Also `NewsPro` is a CP specific name for Newsroom. We should use the configured [SITE_NAME](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/web/default_settings.py#L182) instead.

This config needs to be passed to the front-end, which can be done in the [scripts.html](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/templates/scripts.html#L2) file, by adding something like `window.siteName = '{{ config.SITE_NAME }}';`.
Don't forget to add `siteName` to the [globals.d.ts](https://github.com/superdesk/newsroom-core/blob/develop/assets/globals.d.ts) file

assets/agenda/utils.ts
```diff
@@ -1015,42 +1023,54 @@ export const getCoverageTooltip = (coverage: any, beingUpdated?: any) => {
     const slugline = coverage.item_slugline || coverage.slugline;
     const coverageType = getCoverageDisplayName(coverage.coverage_type);
     const coverageScheduled = moment(coverage.scheduled);
+    const assigneeText = getCoverageAsigneeName(coverage) ? `, assignee: ${getCoverageAsigneeName(coverage)}` : '';
```
MarkLark86, line 1026: Doing it this way would not flag `assignee: ` nor `desk: ` for translation. Consider a better way to append the desk/assignee details onto the tooltip

MarkLark86 (changes requested):

assets/agenda/utils.ts
devketanpro: you mean like this :
```.ts
let additionalDetails = '';

    if (assignee) {
        additionalDetails += `, ${gettext('assignee')}: ${assignee}`;
    }

    if (desk) {
        additionalDetails += `, ${gettext('desk')}: ${desk}`;
    }

    if (coverage.workflow_status === WORKFLOW_STATUS.DRAFT) {
        return gettext('{{ type }} coverage {{ slugline }} {{ status_text }}{{additionalDetails}}', {
            type: coverageType,
            slugline: slugline,
            status_text: getCoverageStatusText(coverage),
            additionalDetails: additionalDetails,
        });

File: assets/agenda/components/AgendaCoverages.tsx [hunk 6]
```diff
@@ -53,7 +53,7 @@ function AgendaCoveragesComponent({item, coverages, wireItems, actions, user, on
                 const deskName = getCoverageDeskName(coverage);
                 const assignedUserEmail = coverage.assigned_user_email;
                 const assignedDeskEmail = coverage.assigned_desk_email;
-                const subject = `Coverage inquiry from NewsPro user: ${item.name || item.slugline}`;
+                const subject = gettext(`Coverage inquiry from ${window.sitename} user: ${item.name || item.slugline}`);
```
MarkLark86, line 56: This will not work as intended when flagging for translation. You will need to pass the dynamic values as a dictionary to the 2nd argument of `gettext` with the key names used in the translation text

PR #714, opened 2023-12-26
The language config for the date picker calendar in the Agenda section is not picked up anymore

dzonidoo:
NHUB-472

## Checklist

### DIFF
--- a/assets/components/CalendarButton.tsx
+++ b/assets/components/CalendarButton.tsx
@@ -31,7 +31,7 @@ class CalendarButton extends React.Component<any, any> {
         const rootLocale = locale.substring(0, 2);
         const allLocales = locales as any;
 
-        return allLocales[locale] || allLocales[rootLocale];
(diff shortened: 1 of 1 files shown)


assets/components/CalendarButton.tsx
```diff
@@ -24,7 +26,21 @@ class CalendarButton extends React.Component<any, any> {
         prevProps.activeDate === EARLIEST_DATE && this.setState({startDate: moment(this.props.activeDate)});
     }
 
+    getLocale = () => {
+        const locale = window.locale.replace('_', '');
+        const rootLocale = locale.substring(0, 2);
+        const allLocales = locales as any;
+
+        return allLocales[locale] || allLocales[rootLocale];
```
thecalcc, line 34: You can use `??` if `allLocales[rootLocale]` is intended to be a fallback, `||` is a very old way of doing it

PR #719, opened 2024-01-03
Fix : clear user search results in the companyadmin when switch tabs [NHUB-462]

devketanpro:
PR based on latest [issue](https://sofab.atlassian.net/browse/NHUB-462?focusedCommentId=126716) which mentioned on the ticket

## Checklist


petrjasek:
would be good to cover this with some e2e tests

devketanpro:
added e2e tests that covered this [bug](https://sofab.atlassian.net/browse/NHUB-462?focusedCommentId=126716)

thecalcc (approved):

PR #721, opened 2024-01-04
[TGA-74] Feature: Public Dashboard

MarkLark86:
## Checklist


MarkLark86:
@thecalcc I improved the WireList dashboard card by:
* Moving redux/connect to the parent container, instead of a connect to each list item (1 connect/listener vs many)
* Use LargeList instead of CompactList (so slugling, abstract and other metadata are shown)
* Rearranged the imports so the 3rd party imports are at the top, and others are grouped based on the functionality they provide
* Directly use `ICardProps` on the component
* Ability to configure the `size` config of the Card.

You can see these changes in the 2 commits: https://github.com/superdesk/newsroom-core/pull/721/commits/575e93b427304f9be562fbc750f500f36eba132d & https://github.com/superdesk/newsroom-core/pull/721/commits/afe65f74951507829d5c03bfbf455ce237c1854d

thecalcc (approved):

thecalcc:
> @thecalcc I improved the WireList dashboard card by:
> * Use LargeList instead of CompactList (so slugling, abstract and other metadata are shown)

Cool stuff, do you have some screenshots of how it looks now? I just wanted to see it compared to before. And great that now we don't connect each item separately, that's a lot better how you improved it.

MarkLark86:
> > @thecalcc I improved the WireList dashboard card by:
> > 
> > * Use LargeList instead of CompactList (so slugling, abstract and other metadata are shown)
> 
> Cool stuff, do you have some screenshots of how it looks now? I just wanted to see it compared to before. And great that now we don't connect each item separately, that's a lot better how you improved it.

FYI:
image

newsroom/auth/views.py
```diff
@@ -200,7 +208,7 @@ def login_with_token(token):
 @blueprint.route("/logout")
 def logout():
     clear_user_session()
-    return flask.redirect(flask.url_for("auth.login", logout=1))
+    return flask.redirect(flask.url_for("wire.index"))
```
petrjasek, line 211: I think it needs that login with logout to sign off from firebase

newsroom/public/views.py
```diff
@@ -1,9 +1,67 @@
...
+PUBLIC_DASHBOARD_CARDS_CACHE_KEY = "public-dashboard-cards"
+PUBLIC_DASHBOARD_ITEMS_CACHE_KEY = "public-dashboard-items"
+
+
+def get_public_dashboard_config():
+    if app.cache.get(PUBLIC_DASHBOARD_CONFIG_CACHE_KEY):
+        return app.cache.get(PUBLIC_DASHBOARD_CONFIG_CACHE_KEY)
+
+    config = get_resource_service("ui_config").get_section_config("home")
+    app.cache.set(PUBLIC_DASHBOARD_CONFIG_CACHE_KEY, config, timeout=app.config.get("DASHBOARD_CACHE_TIMEOUT", 300))
```
petrjasek, line 23: would be good to use a different timeout so it won't expire all together

newsroom/public/views.py
```diff
@@ -20,7 +20,9 @@ def get_public_dashboard_config():
         return app.cache.get(PUBLIC_DASHBOARD_CONFIG_CACHE_KEY)
 
     config = get_resource_service("ui_config").get_section_config("home")
-    app.cache.set(PUBLIC_DASHBOARD_CONFIG_CACHE_KEY, config, timeout=app.config.get("DASHBOARD_CACHE_TIMEOUT", 300))
+    app.cache.set(
+        PUBLIC_DASHBOARD_CONFIG_CACHE_KEY, config, timeout=app.config.get("PUBLIC_CONTENT_CACHE_TIMEOUT", 240)
```
petrjasek, line 24: what I mean is to use slightly different timeout for each app.cache.set call, maybe we should add some util for that like:
```py
def cache_timeout(timeout: int, extra=60):
    return timeout + random.randrange(1, extra)

app.cache.set(..., timeout=cache_timeout(app.config.get("PUBLIC_CONTENT_CACHE_TIMEOUT", 300))
```

MarkLark86: My reason for keeping the cache keys together are
* Public dashboard cache will purge once every 4 minutes
* Internal dashboard item cache is every 5 minutes
* There are 3 cache keys used for public access
  * GET UI config
  * GET card config
  * GET card items
* Keeping the above resources in sync, no chance of different cache keys having out of sync data
* 2/3 of the caches are for very small units of data (UI config, card config), so additional cache save cost is not much

petrjasek: ok makes sense

petrjasek (approved):

PR #723, opened 2024-01-04
Label component

dzonidoo:
NHUB-468

## Checklist


petrjasek:
hi @dzonidoo can you update the PR?

assets/components/Label.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    text: string;
+    type?: 'default' | 'success' | 'warning' | 'alert' | 'info' | 'highlight';
+    size?: 'small' | 'medium' | 'big';
+    style?: 'fill' | 'hollow' | 'translucent'; // defaults to 'fill'
+    className?: string;
```
petrjasek, line 9: can we avoid that?

assets/components/Label.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    text: string;
+    type?: 'default' | 'success' | 'warning' | 'alert' | 'info' | 'highlight';
+    size?: 'small' | 'medium' | 'big';
+    style?: 'fill' | 'hollow' | 'translucent'; // defaults to 'fill'
```
petrjasek, line 8: I don't see hollow being used at all, if fill is default, I wonder if it wouldn't be better to handle it as param like:
```
default: <Label ... />
translucent: <Label translucent ... />
```
but if this is somehow consistent with other components I don't mind

assets/components/Label.tsx
```diff
@@ -0,0 +1,27 @@
...
+    render() {
+        const classes = classNames('label label--rounded', {
+            'label--default': !this.props.type,
+            [`label--${this.props.type}`]: this.props.type !== undefined,
+            [`label--${this.props.style}`]: this.props.style,
+            [`label--${this.props.size}`]: this.props.size,
+        }, this.props.className);
+
+        return (
+            <span className={classes}>
```
petrjasek, line 22: I wonder why not use `label` tag if it's called a label, in case this is not input label we should probably call it something else

dzonidoo: We don't use the label tag because the purpose of this component is not just to support the input.

assets/components/Label.tsx
dzonidoo: We will be using all types soon

assets/styles/labels.scss
```diff
@@ -51,11 +78,12 @@ $flex-align: (
     @include label;
     --label-height: 20px;
 
-    background-color: transparent;
-    border-style: solid;
-    border-width: 1px;
-    border-color: $gray-mid;
-    color: $gray-mid;
+    @each $name, $color in $label-bg-colors {
+        &.label--#{"" + $name} {
```
petrjasek: imo you don't have to add that `""` there, it was an issue when table keys were actual color names

petrjasek (approved):

PR #725, opened 2024-01-09
[NHUB-474] improve: Add new 'Not Intended' coverage status filter

MarkLark86:
Additionally, adds new config to control which coverage status filters are available and what wording is to be used

## Checklist

@petrjasek This one is for v2.6+ as there was an additional requirements for the filters row here


MarkLark86:
@petrjasek I just realized there's no difference, there'll be a merge conflict with merging release/2.5 into develop anyways

PR #726, opened 2024-01-09
[NHUB-474] improve: Add new 'Not Intended' coverage status filter

MarkLark86:
Additionally, adds new config to control which coverage status filters are available and what wording is to be used

## Checklist

@petrjasek This is the one for <= v2.5


newsroom/agenda/agenda.py
```diff
@@ -576,12 +576,10 @@ def _filter_terms(filters, item_type):
...
                 )
             elif val == ["not planned"]:
-                must_term_filters.append(
+                must_not_term_filters.append(
                     nested_query(
                         path="coverages",
-                        query={
-                            "bool": {"filter": [{"terms": {"coverages.coverage_status": ["coverage not intended"]}}]}
-                        },
+                        query={"bool": {"must": [{"exists": {"field": "coverages"}}]}},
```
petrjasek, line 582: seems like we don't need that `bool.must` wrapper

newsroom/agenda/agenda.py
```diff
@@ -592,6 +590,16 @@ def _filter_terms(filters, item_type):
                         query={"bool": {"must": [{"exists": {"field": "coverages.delivery_id"}}]}},
                     )
                 )
+            elif val == ["not intended"]:
+                must_term_filters.append(
+                    nested_query(
+                        path="coverages",
+                        query={
+                            "bool": {"filter": [{"terms": {"coverages.coverage_status": ["coverage not intended"]}}]}
```
petrjasek, line 598: here it also seems like an extra `bool.filter`

petrjasek (approved):

petrjasek:
hi @MarkLark86 pls don't merge it to 2.5 yet we might only need it in 2.6, should know tomorrow

petrjasek:
@MarkLark86 it's good to go into 2.5

PR #727, opened 2024-01-09
Fix carousel display, item preview from HomeApp

thecalcc:
NHUB-475

## Checklist


tomaskikutis:
does it look the same without overwriting CSS variables?

assets/home/components/HomeApp.tsx
```diff
@@ -22,6 +22,7 @@ import {personalizeHome} from 'agenda/actions';
 import {RadioButtonGroup} from 'features/sections/SectionSwitch';
 import {getCurrentUser} from 'company-admin/selectors';
 import {IPersonalizedDashboardsWithData} from 'home/reducers';
+import '@superdesk/common/dist/src/index.css';
```
tomaskikutis, line 25: wasn't it already imported from base component meaning it wouldn't need additional imports?

thecalcc: it is, but the HomeApp doesn't extend the base, therefor this import wouldn't be available in the homeapp if not explicitly imported as I did

tomaskikutis: What else doesn't extend the base? If this is the only one then fine, but if there are others we should to come up with a way to import it once in the project

assets/am-news/index.ts
```diff
@@ -19,7 +19,7 @@ import {
     toggleNavigation,
     toggleNavigationById,
 } from '../search/actions';
-
+import {render} from 'render-utls';
```
tomaskikutis, line 22: should be `render-utils`?

assets/layout/components/SearchBase.tsx
```diff
@@ -257,7 +275,7 @@ export default class BaseApp extends React.Component<any, any> {
     }
 }
 
-BaseApp.propTypes = {
+SearchBase.propTypes = {
```
tomaskikutis, line 278: if you copied info to IProps interface propTypes can be removed

tomaskikutis (approved):

PR #730, opened 2024-01-10
add coverage related functions to global templates [CPCN-523]

devketanpro:
https://github.com/superdesk/newsroom-app-cp/pull/158

## Checklist

### DIFF
--- a/newsroom/agenda/utils.py
+++ b/newsroom/agenda/utils.py
@@ -172,6 +172,8 @@ def get_coverage_status(coverage: Dict[str, Any]) -> str:
         return "Not decided"
     elif coverage_status == "coverage not intended":
         return "Cancelled"
+    else:
+        return ""
 
 
 def get_item_type(item: Dict[str, Any]) -> Literal["event", "planning"]:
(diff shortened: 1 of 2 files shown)


petrjasek (approved):

newsroom/agenda/utils.py
```diff
@@ -164,6 +164,16 @@ def get_planning_coverages(item, plan_id):
     return [coverage for coverage in item.get("coverages") or [] if coverage.get("planning_id") == plan_id]
 
 
+def get_coverage_status(coverage: Dict[str, Any]) -> str:
+    coverage_status = coverage.get("coverage_status")
+    if coverage_status == "coverage intended":
+        return "Planned"
```
petrjasek, line 170: those would probably need translations

newsroom/agenda/utils.py
```diff
@@ -164,6 +164,18 @@ def get_planning_coverages(item, plan_id):
     return [coverage for coverage in item.get("coverages") or [] if coverage.get("planning_id") == plan_id]
 
 
+def get_coverage_status(coverage: Dict[str, Any]) -> str:
+    coverage_status = coverage.get("coverage_status")
+    if coverage_status == "coverage intended":
+        return gettext("Planned")
+    elif coverage_status in ("coverage not decided yet", "coverage upon request"):
```
petrjasek, line 171: is there any requirement not to have specific status for `on request` coverages?

devketanpro: nope, in the templates, there is not any specific requirement for `on request` .

petrjasek: would keep those separate then

petrjasek (approved):

PR #733, opened 2024-01-15
Images downloaded from NewsPro have low resolution and small size comparing to CP's legacy system [CPCN-514]

devketanpro:
first use `original` rendition if it is present instead of `baseImage` in story_image downloads.

https://github.com/superdesk/newsroom-app-cp/pull/160

## Checklist

### DIFF
--- a/newsroom/wire/formatters/picture.py
+++ b/newsroom/wire/formatters/picture.py
@@ -2,6 +2,7 @@ import mimetypes
 
 from newsroom.wire.formatters.base import BaseFormatter
 from newsroom.wire.utils import get_picture
+from flask import current_app as app
 
 
 class PictureFormatter(BaseFormatter):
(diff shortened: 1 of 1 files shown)


newsroom/wire/formatters/picture.py
```diff
@@ -27,7 +27,8 @@ def format_item(self, item, item_type="items"):
         if not picture:
             raise ValueError("Undefined picture")
 
-        picture_details = picture.get("renditions", {}).get("baseImage", {})
+        renditions = picture.get("renditions", {})
+        picture_details = renditions.get("original") or renditions.get("baseImage", {})
```
petrjasek, line 31: we want to have it configurable, so can you pls add some `DOWNLOAD_RENDITION` config and use it here?

devketanpro: okay

newsroom/wire/formatters/picture.py
```diff
@@ -27,7 +28,8 @@ def format_item(self, item, item_type="items"):
         if not picture:
             raise ValueError("Undefined picture")
 
-        picture_details = picture.get("renditions", {}).get("baseImage", {})
+        renditions = picture.get("renditions", {})
+        picture_details = renditions.get(app.config.get("DOWNLOAD_RENDITION")) or renditions.get("baseImage", {})
```
petrjasek, line 32: would be good to check if `picture_details is None` and somehow handle it, otherwise this might raise an error

petrjasek (approved):

PR #736, opened 2024-01-17
Fix : Coverage status shown wrong in pre and full view and in the list view tooltips [NHUB-480]

devketanpro:
## Checklist

### DIFF
--- a/assets/agenda/utils.ts
+++ b/assets/agenda/utils.ts
@@ -1023,13 +1023,18 @@ export function formatCoverageDate(coverage: ICoverage) {
         parseDate(coverage.scheduled).format(COVERAGE_DATE_TIME_FORMAT);
 }
 
+export function formattedCoverageStatusText(coverage: ICoverage) {
+    const statusText = getCoverageStatusText(coverage).replace('Coverage ', '');
+    return statusText.charAt(0).toUpperCase() + statusText.slice(1);
+}
+
 export const getCoverageTooltip = (coverage: any, beingUpdated?: any) => {
     const slugline = coverage.item_slugline || coverage.slugline;
     const coverageType = getCoverageDisplayName(coverage.coverage_type);
     const coverageScheduled = moment(coverage.scheduled);
     const assignee = getCoverageAsigneeName(coverage);
     const desk = getCoverageDeskName(coverage);
-    const status =  getCoverageStatusText(coverage);
+    const status =  formattedCoverageStatusText(coverage);
     const assignedDetails = [
         assignee ? gettext('assignee: {{name}}', {name: assignee}) : '',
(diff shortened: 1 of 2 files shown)


assets/agenda/utils.ts
```diff
@@ -1041,12 +1042,13 @@ export const getCoverageTooltip = (coverage: any, beingUpdated?: any) => {
             assignedDetails,
         });
     } else if (coverage.workflow_status === WORKFLOW_STATUS.ASSIGNED) {
-        return gettext('Planned {{ type }} coverage {{ slugline }}, expected {{date}} at {{time}} {{assignedDetails}}', {
+        return gettext('{{ status }} {{ type }} coverage {{ slugline }}, expected {{date}} at {{time}} {{assignedDetails}}', {
```
MarkLark86, line 1045: I don't believe this is going to make much sense using the `getCoverageStatusText` here.

For example, a Planned coverage currently states:
`Planning text coverage {{ slugline }}, expected at ....`

After these changes it would state something like:
`Coverage planned text coverage {{ slugline }}, expected at ....` which has the repeating `coverage`

assets/agenda/utils.ts
```diff
@@ -1023,12 +1024,18 @@ export function formatCoverageDate(coverage: ICoverage) {
         parseDate(coverage.scheduled).format(COVERAGE_DATE_TIME_FORMAT);
 }
 
+export function formattedCoverageStatusText(coverage: ICoverage) {
+    const statusText = getCoverageStatusText(coverage).replace('Coverage ', '');
```
MarkLark86, line 1028: This will not work with translated text, `Coverage` is for English only.

IMO a better approach might be to scrap all your changes and place a new if branch [here](https://github.com/superdesk/newsroom-core/blob/develop/assets/agenda/utils.ts#L1036).

```ts
if (coverage.coverage_status !== 'coverage intended') {
    return DRAFT_STATUS_TEXTS[coverage.coverage_status];
} else if (coverage.workflow_status === WORKFLOW_STATUS.DRAFT) {
    ....
}
```

Reason being, if a Coverage is not planned, not intended, on merit/request, then there would be no need to display an expected date/time to the user, because there should not be one for that coverage.

Instead, with the above suggestion, the tooltip would simply state something like `Coverage not planned`

MarkLark86 (changes requested):

assets/agenda/utils.ts
devketanpro: yep got it, Thanks for clarification

MarkLark86 (approved):

MarkLark86:
@devketanpro This should go into the release/2.6 branch and not develop

devketanpro:
created a new PR based on release/2.6 branch

PR #737, opened 2024-01-19
Fix : Coverage status shown wrong in pre and full view and in the list view tooltips [NHUB-480]

devketanpro:
https://github.com/superdesk/newsroom-core/pull/736

## Checklist

### DIFF
--- a/assets/agenda/utils.ts
+++ b/assets/agenda/utils.ts
@@ -48,7 +48,7 @@ const Groupers: any = {
 };
 
 export function getCoverageStatusText(coverage: any) {
-    if (coverage.workflow_status === WORKFLOW_STATUS.DRAFT) {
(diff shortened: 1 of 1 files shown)


assets/agenda/utils.ts
```diff
@@ -48,7 +48,7 @@ const Groupers: any = {
 };
 
 export function getCoverageStatusText(coverage: any) {
-    if (coverage.workflow_status === WORKFLOW_STATUS.DRAFT) {
+    if (coverage.workflow_status === WORKFLOW_STATUS.DRAFT || coverage.coverage_status !== 'coverage intended') {
```
petrjasek, line 51: would use some constant for that `coverage intended` to avoid possible typos

petrjasek (approved):

PR #739, opened 2024-01-19
when looking for matching topics run a query per user

petrjasek:
so it won't fail for other users when there is a bad topic, should also avoid timeouts when checking too many products at the same time.

NHUB-479

## Checklist

### DIFF
--- a/newsroom/search/service.py
+++ b/newsroom/search/service.py
@@ -416,7 +416,8 @@ class BaseSearchService(Service):
             return
 
         current_user = get_user(required=False)
-        search.is_admin = is_admin(current_user)
+        if current_user:
+            search.is_admin = is_admin(current_user)
 
         if search.is_admin and search.args.get("user"):
             search.user = get_resource_service("users").find_one(req=None, _id=search.args["user"])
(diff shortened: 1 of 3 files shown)


newsroom/search/service.py
```diff
@@ -842,56 +843,60 @@ def prepend_embargoed_items_to_response(self, response, req, lookup):
...
-                exc,
-                exc_info=True,
-            )
-            raise
+            try:
+                search_results = self.internal_get(req, None)
+
+                for topic in queried_topics:
+                    if search_results.hits["aggregations"]["topics"]["buckets"][str(topic["_id"])]["doc_count"] > 0:
+                        topic_matches.append(topic["_id"])
```
MarkLark86, line 885: I was thinking, to help with Python memory consumption and performance etc, what if we `yield` the topic here and consume it in [send_topic_notification_emails](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/push.py#L904)?

This way the topic is yielded and processed in the one step, as opposed to constructing the list of topics to later iterate & filter those same topics?

MarkLark86 (approved):

newsroom/search/service.py
petrjasek: makes sense, would do it rather separately as we might need to backport this also into 2.5 so would rather keep it small

PR #742, opened 2024-01-23
Fix: Event description in the CSV file shows blank when downloading from cpcn-uat [CPCN-552]

devketanpro:
handle case where `description_long` is `null`
## Checklist

### DIFF
--- a/newsroom/agenda/formatters/csv_formatter.py
+++ b/newsroom/agenda/formatters/csv_formatter.py
@@ -43,7 +43,7 @@ class CSVFormatter(BaseFormatter):
         event = item.get("event", {})
         return {
             "Event name": item.get("name", ""),
-            "Description": item.get("definition_long", item.get("definition_short", "")),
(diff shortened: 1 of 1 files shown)


newsroom/agenda/formatters/csv_formatter.py
```diff
@@ -43,7 +43,7 @@ def format_event(self, item: Dict[str, Any]) -> Dict[str, Any]:
         event = item.get("event", {})
         return {
             "Event name": item.get("name", ""),
-            "Description": item.get("definition_long", item.get("definition_short", "")),
+            "Description": item.get("definition_long") or item.get("definition_short", ""),
```
petrjasek, line 46: would do it probably `item.get("definition_long") or item.get("definition_short") or ""` so we make sure it's always string

devketanpro: yep

PR #745, opened 2024-01-24
Fix topic being removed from folder on edit

thecalcc:
CPCN-562
## Checklist

### DIFF
--- a/assets/user-profile/actions.ts
+++ b/assets/user-profile/actions.ts
@@ -258,9 +258,8 @@ export function saveFolder(folder: ITopicFolder, data: {name: string}, global?: 
         const state = getState();
         const url = getFoldersUrl(state.company, state.user?._id, global, folder._id);
 
-        const returnFolder = (allFolders: IFoldersUnified, id: string) => {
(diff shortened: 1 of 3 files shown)


tomaskikutis:
What was the cause of the issue?

thecalcc:
> What was the cause of the issue?

We were updating the topic regardless if one was created in the topic editor or not.

assets/user-profile/actions.ts
```diff
@@ -258,20 +258,19 @@ export function saveFolder(folder: ITopicFolder, data: {name: string}, global?:
         const state = getState();
         const url = getFoldersUrl(state.company, state.user?._id, global, folder._id);
 
+        const returnFolder = (allFolders: IFoldersUnified, id: string) =>
```
tomaskikutis, line 261: It's not a good practice to use the word `return` in the function name, because it causes confusion with the built-in return operator. Use get/find instead. I'd call it simply `getFolder`

tomaskikutis (approved):

PR #746, opened 2024-01-24
Fix: the format of information in the CSV file for exported events [CPCN-553]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/agenda/formatters/csv_formatter.py
+++ b/newsroom/agenda/formatters/csv_formatter.py
@@ -42,9 +42,9 @@ class CSVFormatter(BaseFormatter):
         return csv_string.getvalue().encode("utf-8")
 
     def format_event(self, item: Dict[str, Any]) -> Dict[str, Any]:
-        SUBJ_SCHEMAS = app.config.get("AGENDA_CSV_SUBJECT_SCHEMES")
+        subj_schemas = app.config.get("AGENDA_CSV_SUBJECT_SCHEMES", [])
         event = item.get("event", {})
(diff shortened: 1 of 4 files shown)


newsroom/agenda/formatters/csv_formatter.py
```diff
@@ -40,7 +42,9 @@ def serialize_to_csv(self, items: List[Dict[str, Any]]) -> bytes:
         return csv_string.getvalue().encode("utf-8")
 
     def format_event(self, item: Dict[str, Any]) -> Dict[str, Any]:
+        subj_schemas = app.config.get("AGENDA_CSV_SUBJECT_SCHEMES", [])
         event = item.get("event", {})
+        event["subject"] = get_filtered_subject(event.get("subject", {}), subj_schemas)
```
petrjasek, line 47: default value should be rather `[]`

newsroom/agenda/utils.py
```diff
@@ -271,3 +271,17 @@ def push_agenda_item_notification(name, item, **kwargs):
...
+    subject: List[Dict[str, Any]], schemas: List[str]
+) -> Union[List[Dict[str, Any]], Dict[str, Any]]:
+    """
+    Filter subject based on schemas
+    """
+    if not schemas:
+        return subject
+    filtered_items = []
+    for schema in schemas:
+        filtered_items += [item for item in subject if "scheme" in item and item["scheme"] == schema]
```
petrjasek, line 286: instead of doing the for look you can do:
```py
return [subj for subj in subject if subj.get("scheme") in schemas]
```

petrjasek (approved):

PR #751, opened 2024-02-01
Fix : Users unable to view content when clicking on 'Topics' [CPCN-617]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/wire/views.py
+++ b/newsroom/wire/views.py
@@ -16,7 +16,7 @@ from superdesk import get_resource_service
 from superdesk.default_settings import strtobool
 from newsroom.auth.utils import check_user_has_products, is_valid_session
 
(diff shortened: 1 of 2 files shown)


newsroom/wire/views.py
```diff
@@ -108,7 +108,7 @@ def get_view_data() -> Dict:
             for f in app.download_formatters.values()
             if "wire" in f["types"]
         ],
-        "navigations": get_navigations_by_company(company, product_type="wire") if company else [],
+        "navigations": get_navigations_by_user(user, product_type="wire") if user else [],
```
petrjasek, line 111: we need to get user product navigations + navigations for company products which have `0` seats so they are available for all users

petrjasek (approved):

PR #752, opened 2024-02-02
Change ES mapping for urgency and priority to use 'keywords' instead of 'integer' [NHUB-477]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/__init__.py
+++ b/newsroom/__init__.py
@@ -19,8 +19,8 @@ MONGO_PREFIX = "CONTENTAPI_MONGO"
 ELASTIC_PREFIX = "CONTENTAPI_ELASTICSEARCH"
 
 SCHEMA_VERSIONS = {
-    "wire": 4,
-    "agenda": 6,
+    "wire": 3,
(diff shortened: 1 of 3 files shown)


newsroom/__init__.py
```diff
@@ -19,8 +19,8 @@
 ELASTIC_PREFIX = "CONTENTAPI_ELASTICSEARCH"
 
 SCHEMA_VERSIONS = {
-    "wire": 3,
-    "agenda": 5,
+    "wire": 4,
+    "agenda": 6,
```
petrjasek, line 23: not sure if we should bump it, that would trigger reindexing automatically but it's rather an optimization than something that would be needed, so would keep those as it is

petrjasek: what do you think @MarkLark86 ?

MarkLark86: I agree. We should only bump these versions if the system breaks without a re-index.

petrjasek (approved):

PR #792, opened 2024-02-15
FIX : Unable to download CSV file in the Calendar Section [CPCN-626]

devketanpro:
handle 'None' subjects in the events.

## Checklist

### DIFF
--- a/tests/core/test_csv_formatter.py
+++ b/tests/core/test_csv_formatter.py
@@ -97,7 +97,7 @@ def test_csv_formatter_item(client, app):
         "Australia/Sydney",
         "Sydney",
         "Australia",
-        "Statistics & Economic Indicators,Economic Indicators",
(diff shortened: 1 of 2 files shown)


tests/core/test_csv_formatter.py
```diff
@@ -31,6 +31,17 @@
...
     {"name": "Statistics & Economic Indicators", "qcode": "150", "scheme": "event_types", "code": "150"},
     {"name": "Economic Indicators", "qcode": "3", "scheme": "categories", "code": "3"},
+    {
+        "name": None,
+        "qcode": "20001237",
+        "parent": "08000000",
+        "iptc_subject": None,
+        "ap_subject": None,
+        "in_jimi": False,
+        "translations": {"name": {"en-CA": "anniversary", "fr-CA": None}},
```
petrjasek, line 41: would be good to use the translation if there is one for the event language

petrjasek (approved):

PR #809, opened 2024-02-26
Fix : listview rendering and bodyimages rendering in detail preview [CPCN-670]

devketanpro:
PR based on latest comment :- https://sofab.atlassian.net/browse/CPCN-670?focusedCommentId=128070
## Checklist

### DIFF
--- a/assets/wire/components/ItemDetails.tsx
+++ b/assets/wire/components/ItemDetails.tsx
@@ -13,7 +13,8 @@ import {
     isKilled,
     DISPLAY_ABSTRACT,
     isPreformatted,
-    getPictureList
+    getPictureList,
+    getBodyPictureList
 } from 'wire/utils';
 import types from 'wire/types';
 import Content from 'ui/components/Content';
@@ -51,7 +52,7 @@ function ItemDetails({
     const media = getOtherMedia(item);
     const itemType = isPreformatted(item) ? 'preformatted' : 'text';
     const allMedia = getPictureList(item);
-
+    const bodyImages = getBodyPictureList(item);
     return (
         <Content type="item-detail">
             <ContentHeader>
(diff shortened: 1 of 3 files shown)


assets/wire/components/ItemDetails.tsx
```diff
@@ -51,7 +52,7 @@ function ItemDetails({
     const media = getOtherMedia(item);
     const itemType = isPreformatted(item) ? 'preformatted' : 'text';
     const allMedia = getPictureList(item);
```
petrjasek, line 54: imo it would be better to extend this func to do the filtering (and thus avoid that duplication), eg.:
```ts
getPictureList(item, {includeFeatured = false, includeEditor = false})
```

assets/wire/utils.ts
```diff
@@ -108,10 +108,25 @@ function getBodyPicture(item: IArticle): IArticle | null {
     return pictures.length ? pictures[0] : null;
 }
 
-export function getPictureList(item: IArticle): Array<IArticle> {
+export function getPictureList(item: IArticle, options: {includeFeatured?: boolean, includeEditor?: boolean} = {}): IArticle[] {
```
petrjasek, line 111: I think you can just destruct it with:
```ts
export function getPictureList(item: IArticle, {includeFeatured = false, includeEditor = false} = {}): IArticle[] {
```
but thinking about it it might be also useful to export a function which would have the params set like
`getGalleryMedia` that would call `getPictureList(item, {includeFeatured: false, includeEditor: false});`
so we don't have to use the params in preview/detail and we make sure it's consistent

assets/wire/components/WirePreview.tsx
```diff
@@ -50,7 +50,7 @@ class WirePreview extends React.PureComponent<any, any> {
         const {item, user, actions, followStory, topics, previewConfig, downloadMedia, listConfig, filterGroupLabels} = this.props;
         const featureMedia = getFeatureMedia(item);
         const media = getOtherMedia(item);
-        const allMedia = getPictureList(item, {includeFeatured:false, includeEditor: false});
+        const allMedia = getGalleryMedia(item);
```
petrjasek, line 53: `allMedia` is misleading a bit, should be `galleryMedia` or `additionalMedia`

petrjasek (approved):

PR #810, opened 2024-02-27
stop caching products

petrjasek:
to avoid sending emails based on old config

also filtering out topics for sections user does not have access to anymore

CPCN-673 CPCN-647

## Checklist

### DIFF
--- a/newsroom/auth/utils.py
+++ b/newsroom/auth/utils.py
@@ -181,7 +181,7 @@ def revalidate_session_user():
     return is_valid
 
 
-def get_user_sections(user) -> Dict[str, bool]:
+def get_user_sections(user: User) -> Dict[str, bool]:
     if not user:
         return {}
 
@@ -199,7 +199,7 @@ def get_user_sections(user) -> Dict[str, bool]:
     return {}
 
 
-def user_has_section_allowed(user: User, section) -> bool:
+def user_has_section_allowed(user: User, section: str) -> bool:
     sections = get_user_sections(user)
     if sections:
         return sections.get(section, False)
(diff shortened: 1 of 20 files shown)


newsroom/auth/utils.py
```diff
@@ -177,23 +181,26 @@ def revalidate_session_user():
     return is_valid
 
 
-def get_user_sections() -> Dict[str, bool]:
-    user = get_user()
+def get_user_sections(user) -> Dict[str, bool]:
```
MarkLark86, line 184: It would be good to have the `user` argument typed here as well

newsroom/products/products.py
```diff
@@ -55,6 +58,28 @@ def on_deleted(self, doc: Product) -> None:
...
 
+    def create(self, docs):
+        company_products = {}
+        for doc in docs:
+            if doc.get("companies"):
+                for company_id in doc["companies"]:
+                    company_products.setdefault(company_id, []).append(doc)
+        res = super().create(docs)
+        if company_products:
+            warnings.warn("Using deprecated product.companies", DeprecationWarning)
```
MarkLark86, line 69: This was made obsolete in v2.4. Do we still have code using this somewhere?
IMO, it would make for cleaner/stricter code if we removed support for this

MarkLark86 (approved):

newsroom/products/products.py
petrjasek: it got quite complicated to fix all tests, especially those behave ones, otherwise this shouldn't be used, but will need to make some changes in superdesk core steps to make it simpler and that will go into 2.7

PR #825, opened 2024-03-05
Fix: when filter 'planned' coverage status, 'completed' coverages also show up in the results [CPCN-652]

devketanpro:
Only `text` coverages include `delivery_id`. That is why in the `Planned` filter, we are seeing coverages with `workflow status` being `completed` and not having `delivery_id`.

## Checklist


petrjasek:
can you add a test?

petrjasek (approved):

PR #831, opened 2024-03-07
FIX: Update Download method from GET to POST to address URL limiter issue [CPCN-626]

devketanpro:
## Checklist

### DIFF
--- a/assets/server.ts
+++ b/assets/server.ts
@@ -10,10 +10,12 @@ function options(custom: any = {}) {
 function checkStatus(response: Response): Promise<any> {
     if (response.status === 204) {
         return Promise.resolve({});
-    }else if (response.status == 200 && response.headers.get('Content-Disposition')?.includes('attachment')){
+    } else if (response.status >= 200 && response.status < 300) {
+        const contentType = response.headers.get('Content-Type');
+        if (contentType && contentType.includes('application/json')) {
+            return response.json();
+        }
         return Promise.resolve(response);
-    }else if (response.status >= 200 && response.status < 300) {
-        return response.json();
     } else if (response.status === 400) {
         return response.json().then((data: any) => Promise.reject({errorData: data}));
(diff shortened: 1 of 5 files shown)


assets/server.ts
```diff
@@ -10,7 +10,9 @@ function options(custom: any = {}) {
 function checkStatus(response: Response): Promise<any> {
     if (response.status === 204) {
         return Promise.resolve({});
-    } else if (response.status >= 200 && response.status < 300) {
+    }else if (response.status == 200 && response.url && response.url.includes('/download')){
```
petrjasek, line 13: wouldn't do it based on a specific url, rather response mimetype probably

assets/wire/actions.ts
```diff
@@ -416,36 +416,65 @@ export function downloadItems(items: any) {
...
     link.download = '';
     link.href = url;
     link.click();
 }
 
+function initiateDownloadFromBlob(blob: Blob) {
+    const link = document.createElement('a');
+    link.download = '';
+    link.href = URL.createObjectURL(blob);
+    document.body.appendChild(link);
```
petrjasek, line 473: is this append/remove child needed?

petrjasek: I know, just wondering why it's needed here if it's not needed in the other func

assets/server.ts
```diff
@@ -10,7 +10,9 @@ function options(custom: any = {}) {
 function checkStatus(response: Response): Promise<any> {
     if (response.status === 204) {
         return Promise.resolve({});
-    } else if (response.status >= 200 && response.status < 300) {
+    }else if (response.status == 200 && response.headers.get('Content-Disposition')?.includes('attachment')){
+        return Promise.resolve(response);
```
petrjasek, line 14: I would probably make this rather a default behaviour, but we need to do that `.json()` call if the response has `content-type` header `application/json`

petrjasek (approved):

PR #836, opened 2024-03-12
If updates to Planning items are pushed to NH they should be visible in the Agenda list views immediately [NHUB-494]

devketanpro:
## Checklist


assets/agenda/components/AgendaListItemIcons.tsx
```diff
@@ -94,7 +94,7 @@ class AgendaListItemIcons extends React.Component<IProps, IState> {
 
     render() {
         const props = this.props;
-        const state = this.state;
+        const state = this.getUpdatedState();
```
petrjasek, line 97: I wonder why the state is not updated when the props changed, we should rather fix that imo

petrjasek (approved):

PR #839, opened 2024-03-14
FIX : View Content is not greyed out even though the user doesn't have access to the section [CPCN-646]

devketanpro:
## Checklist

### DIFF
--- a/assets/agenda/components/CoverageItemStatus.tsx
+++ b/assets/agenda/components/CoverageItemStatus.tsx
@@ -135,7 +135,8 @@ export default class CoverageItemStatus extends React.Component<any, any> {
                         ref={(elem) => this.elem = elem}
                         href={viewContentEmail ? `mailto:${viewContentEmail}` : undefined}
                         target='_blank'
-                        title={gettext('You don’t have access to this content, please contact {{email}}', {email: viewContentEmail})}
+                        title={viewContentEmail ? gettext('You don’t have access to this content, please contact {{email}}', {email: viewContentEmail}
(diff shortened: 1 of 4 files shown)


assets/agenda/components/CoverageItemStatus.tsx
```diff
@@ -132,7 +132,7 @@ export default class CoverageItemStatus extends React.Component<any, any> {
                         key="contentLink"
                         className='nh-button nh-button--small nh-button--tertiary nh-button--disabled'
                         ref={(elem) => this.elem = elem}
-                        title={gettext('“You don’t have access to this content, please contact test@localhost”')}
+                        title={gettext('You don’t have access to this content, please contact {{email}}', {email: getConfig('view_content_tooltip_email')})}
```
petrjasek, line 135: I wonder if this should be actually link with mailto, otherwise it would be hard to get the email from the tooltip

devketanpro: yep, adding a clickable `mailto:` for the email address would enhance user experience. However, since the ticket does not specify this requirement.

newsroom/web/default_settings.py
```diff
@@ -344,6 +344,7 @@
     "item_actions": {},
     "display_abstract": DISPLAY_ABSTRACT,
     "display_credits": False,
+    "view_content_tooltip_email": "test@localhost.com",
```
petrjasek, line 347: not sure if it makes sense to have some default here, would probably rather hide it if it's not configured

petrjasek (approved):

PR #841, opened 2024-03-16
handle filters with multiple values

tomaskikutis:
CPCN-665

## Checklist


tomaskikutis:
@thecalcc I'm wondering why did you add `clearQuickFilter` function and passed it in multiple places. Wouldn't `toggleFilter` that I ended up using here worked?

thecalcc:
> @thecalcc I'm wondering why did you add `clearQuickFilter` function and passed it in multiple places. Wouldn't `toggleFilter` that I ended up using here worked?

@tomaskikutis did you test if all filters work, in wire and agenda both. I think it had something to do with that. But if they work then we can remove it

thecalcc (approved):

PR #846, opened 2024-03-18
When events is update in the planning-only view, update the event-related planning items [NHUB-494].

devketanpro:
## Checklist

### DIFF
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -591,6 +591,7 @@ export function setAndUpdateNewItems(data: any) {
         }
 
         dispatch(updateItem(item));
+        dispatch(fetchItems());
 
         // Do not use 'killed' items for new-item notifications
         if (item.state === 'killed') {
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -591,8 +591,13 @@ export function setAndUpdateNewItems(data: any) {
         }
 
         dispatch(updateItem(item));
-        dispatch(fetchItems());
 
+        // Fetch related planning items if the updated item has planning_items
+        if (item.item_type === 'event' && item.planning_items && item.planning_items.length > 0) {
+            item.planning_items.forEach((plan: IAgendaItem) => {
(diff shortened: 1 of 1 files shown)


assets/agenda/actions.ts
```diff
@@ -591,6 +591,7 @@ export function setAndUpdateNewItems(data: any) {
         }
 
         dispatch(updateItem(item));
+        dispatch(fetchItems());
```
petrjasek, line 594: why does it need to refetch it? shouldn't it handle the changes via `updateItem`

devketanpro: Nope, this is the case when events contain a related-planning item, so when we change the coverages in the events, here `(dispatch(updateItem(item)))` we get an event item with updated coverages rather than a planning item.
So, in the `planning only` view, associated planning items remain unchanged until we refetch.

petrjasek: I thought the planning items are embedded there, but even if we need to refetch something, I suppose we don't need to refetch everything rigth?

devketanpro: yep, you r right, so in this case, we can do something like fetch an event-related item via planning IDs?

petrjasek: so we get just the planning item in the notification and not the actual event?

devketanpro: we got a full event with an updated planning item in it, but it updates the event only, not its linked planning item state.

devketanpro: so for this we can do something like this, so this updates its related planning item as well.
```.js
item?.planning_items?.map((data:any)=> {
    dispatch(fetchItem(data._id));
})

petrjasek: shouldn't this be rather fixed in the reducer then?

devketanpro: yep, it is more complicated because, in the events, we did not get full data of planning item, so first we need to refetch planning item then update it.

petrjasek (approved):

PR #850, opened 2024-03-20
update check for both local and UTC time in isCoverageForExtraDay function.

devketanpro:
NHUB-494, NHUB-502
## Checklist

### DIFF
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -566,9 +566,10 @@ export const SET_NEW_ITEM = 'SET_NEW_ITEM';
 export function setAndUpdateNewItems(data: any) {
     return function(dispatch: any, getState: any) {
         const item = data.item || {};
+        const state = getState();
 
+        console.log(item)
         if (item.type !== 'agenda') {
-            const state = getState();
 
             // Check if the item is used in the preview or opened agenda item
             // If yes, make it available to the preview
@@ -590,14 +591,15 @@ export function setAndUpdateNewItems(data: any) {
             return Promise.resolve();
         }
 
-        dispatch(updateItem(item));
+        const {itemsById} = state;
+        const prevItem = itemsById[item['_id']];
 
-        // Fetch related planning items if the updated item has planning_items
-        if (item.item_type === 'event' && item.planning_items && item.planning_items.length > 0) {
-            item.planning_items.forEach((plan: IAgendaItem) => {
-                dispatch(fetchItem(plan._id));
-            });
+        // If coverage is updated in the item fetch all items and reintilized group listing.
+        if (prevItem && prevItem.coverages.length !== item.coverages.length) {
+            dispatch(fetchItems());
         }
+
+        dispatch(updateItem(item));
         // Do not use 'killed' items for new-item notifications
         if (item.state === 'killed') {
             return Promise.resolve();
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -598,6 +598,11 @@ export function setAndUpdateNewItems(data: any) {
             dispatch(fetchItems());
         } else {
             dispatch(updateItem(item));
+            if (item.item_type === 'event' && item.planning_items && item.planning_items.length > 0) {
+                item.planning_items.forEach((plan: IAgendaItem) => {
+                    dispatch(fetchItem(plan._id));}
+                );
+            }
         }
         // Do not use 'killed' items for new-item notifications
         if (item.state === 'killed') {
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -568,7 +568,6 @@ export function setAndUpdateNewItems(data: any) {
         const item = data.item || {};
         const state = getState();
 
-        console.log(item)
         if (item.type !== 'agenda') {
 
             // Check if the item is used in the preview or opened agenda item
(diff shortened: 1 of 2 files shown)


assets/agenda/utils.ts
```diff
@@ -215,7 +215,9 @@ export function isWatched(item: IAgendaItem | ICoverage, userId?: IUser['_id']):
  * @return {Boolean}
  */
 export function isCoverageForExtraDay(coverage: ICoverage, dateString: string) {
-    return coverage.scheduled != null && formatDate(moment(coverage.scheduled)) === dateString;
+    return coverage.scheduled != null && (
+        formatDate(moment(coverage.scheduled)) === dateString ||
+        formatDate(moment.utc(coverage.scheduled)) === dateString);
```
petrjasek, line 220: when is that needed? can you provide an example?

devketanpro: Let's consider a scenario where we have an event with planning item and no coverages, and its `planning_date` is set to `"2024-03-19T18:30:00.000Z"`. When we push this event to Newshub, the `display_date` in the agenda will be set to the `planning_date` (referenced [here](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/push.py#L584)). In the agenda list view, the events are grouped by this date, and this particular event falls under the date `"2024-03-19T18:30:00.000Z".`

Now, let's say we add coverage to this event with a scheduled time of `"2024-03-20T19:29:59.000Z".` When we update the event, the `display_date` in the agenda changes accordingly. However, when looking for coverages in the list view, it still refers to the previous planning date, so no coverages appear.

Upon refreshing the page, the grouping is recalculated, and now the event is grouped based on the new display date. item is moved according to the new `display_date` and consequently, the coverages becomes visible in the list view.

Interestingly, when comparing the planning date with the coverage scheduled time using the same date formatting function, such as `formatDate(moment("2024-03-19T18:30:00.000Z")) === formatDate(moment.utc(coverage.scheduled)`), it evaluates to true.

devketanpro: in that dateString `2024-03-19T18:30:00.000Z` which is `Mar 20, 2024`, but coverage scheduled time is `2024-03-20T19:29:59.000Z `which is `Mar 21, 2024`

devketanpro: Check out this one comment: - https://sofab.atlassian.net/browse/NHUB-494?focusedCommentId=128599

devketanpro: I got your point. we have two options to fix this, first option is that if first coverage is added in the previous event, we have to refresh the page to recalculate the event listing based on the new display date (coverage.scheduled) which means refetching all items.
or 
we have to update this logic as
`formatDate(moment(coverage.scheduled)) === dateString || formatDate(moment(item.planning_date)) === dateString`

or first we changed this logic:- 
https://github.com/superdesk/newsroom-core/blob/develop/newsroom/push.py#L572
```.py
if action_type == "add":
    display_dates = get_display_dates(agenda["dates"], agenda["planning_items"])
else:
    display_dates = orig_agenda["display_dates"]
    display_dates += get_display_dates(agenda["dates"], agenda["planning_items"])

agenda["display_dates"] = display_dates
```
in client update logic as 
`formatDate(moment(coverage.scheduled)) === dateString || formatDate(moment(item.display_date[0])) === dateString`

petrjasek: so if the data we get from server are missing something I would handle it there, otherwise if we need to refetch the event when the coverage was added/updated that's also fine

assets/agenda/actions.ts
```diff
@@ -590,14 +591,15 @@ export function setAndUpdateNewItems(data: any) {
...
-        if (item.item_type === 'event' && item.planning_items && item.planning_items.length > 0) {
-            item.planning_items.forEach((plan: IAgendaItem) => {
-                dispatch(fetchItem(plan._id));
-            });
+        // If coverage is updated in the item fetch all items and reintilized group listing.
+        if (prevItem && prevItem.coverages.length !== item.coverages.length) {
+            dispatch(fetchItems());
         }
+
+        dispatch(updateItem(item));
```
petrjasek, line 602: I wonder if this is needed also when it calls `fetchItems`

assets/agenda/actions.ts
```diff
@@ -590,15 +590,15 @@ export function setAndUpdateNewItems(data: any) {
...
         }
 
-        dispatch(updateItem(item));
+        const {itemsById} = state;
+        const prevItem = itemsById[item['_id']];
 
-        // Fetch related planning items if the updated item has planning_items
-        if (item.item_type === 'event' && item.planning_items && item.planning_items.length > 0) {
-            item.planning_items.forEach((plan: IAgendaItem) => {
-                dispatch(fetchItem(plan._id));
```
petrjasek, line 598: I wonder if we're still fine in this case when there is the planning items updated if that wouldn't change the coverages

devketanpro: I checked that part it works fine, If we fetch all the items, they also update its related planning items
so we don't need to fetch its related planing item one by one.

devketanpro: previously, we implemented this for coverages updates

petrjasek: ok

petrjasek (approved):

assets/agenda/actions.ts
devketanpro: my blunders, we require this code; I re-checked and noticed one case is failing.

PR #854, opened 2024-03-21
avoid multiple notifications for recurring events

petrjasek:
CPCN-695

uses https://github.com/superdesk/superdesk-planning/pull/1936

### DIFF
--- a/newsroom/push.py
+++ b/newsroom/push.py
@@ -92,7 +92,7 @@ def push():
     if item.get("type") == "event":
         orig = app.data.find_one("agenda", req=None, _id=item["guid"])
         _id = publish_event(item, orig)
-        notify_new_agenda_item.delay(_id, check_topics=True)
+        notify_new_agenda_item.delay(_id, check_topics=True, is_new=orig is None)
     elif item.get("type") == "planning":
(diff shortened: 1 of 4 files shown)


newsroom/push.py
```diff
@@ -767,6 +767,10 @@ def notify_new_agenda_item(_id, check_topics=True):
     with locked(_id, "agenda"):
         agenda = app.data.find_one("agenda", req=None, _id=_id)
         if agenda:
+            if agenda.get("recurrence_id") and agenda.get("recurrence_id") != _id:
```
MarkLark86, line 770: How will this work if one or more Event(s) are updated in the series that does't include the original?

**Example scenario:**
There is a sporting match that goes for 5 days.
The game is finished on the 3rd day.
The planner cancels the 4th and 5th occurrence of the Event in Superdesk, which subsequently cancels the associated Planning & Coverages for those days.

**Current Functionality**
2 emails would be sent to subscribers, one for each Event that was cancelled

**New Functionality (based on this PR)**
No emails would be sent to subscribers

petrjasek: yep you're right, will ignore the other events only when agenda item is created for the first time

MarkLark86 (approved):

PR #860, opened 2024-03-25
allow multiple auth domains per company

petrjasek:
STTNHUB-330

## Checklist

### DIFF
--- a/data_updates/00015_20240325-150938_companies.py
+++ b/data_updates/00015_20240325-150938_companies.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8; -*-
+# This file is part of Superdesk.
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+#
+# Author  : petr
+# Creation: 2024-03-25 15:09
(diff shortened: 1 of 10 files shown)


data_updates/00015_20240325-150938_companies.py
```diff
@@ -0,0 +1,28 @@
...
+            if company["auth_domain"]:
+                print("Updating company", company["_id"])
+                mongodb_collection.update_one(
+                    {"_id": company["_id"]}, {"$set": {"auth_domains": [company["auth_domain"]]}}
+                )
+
+    def backwards(self, mongodb_collection, mongodb_database):
+        for company in mongodb_collection.find({"auth_domains.0": {"$exists": True}}):
+            mongodb_collection.update_one(
+                {"_id": company["_id"]}, {"$set": {"auth_domain": [company["auth_domains"][0]]}}
```
MarkLark86, line 27: Should the value once rolled back be a list or strings or a single string?

MarkLark86 (approved):

data_updates/00015_20240325-150938_companies.py
petrjasek: single string when rolled back

PR #862, opened 2024-03-26
make personal dashboard type configurable

petrjasek:
and add missing types

CPCN-516

## Checklist


assets/home/components/HomeApp.tsx
```diff
@@ -32,45 +30,13 @@ const modals: any = {
...
-interface IDispatchProps {
-    openItemDetails: (item: any, state: any) => void;
-    fetchItems: () => any;
-    personalizeHome: () => void;
-    actions: Array<IItemAction>;
-    fetchCardExternalItems: (cardId: string, cardLabel: string) => any;
-    fetchCompanyCardItems: () => any;
-    followStory: (item: any) => void;
-    downloadMedia: (href: any, id: any) => any;
+interface IOwnProps {
```
MarkLark86, line 35: `IOwnProps` is a bit misleading here for 2 reasons:
* Typically `ownProps` naming convention is used with Redux (for `map[X]ToProps` functions, props passed in from parent component vs props passed in from Redux state)
* This interface is defining internal state to the component and not a property that is passed in

newsroom/wire/views.py
```diff
@@ -142,7 +142,7 @@ def delete_dashboard_caches():
 def get_personal_dashboards_data(user, company, topics):
     def get_topic_items(topic):
         query = superdesk.get_resource_service("wire_search").get_topic_query(topic, user, company)
-        return list(superdesk.get_resource_service("wire_search").get_items_by_query(query, size=4))
+        return list(superdesk.get_resource_service("wire_search").get_items_by_query(query, size=6))
```
MarkLark86, line 145: Instead of hardcoding the size to 6, could we use the config for the card type itself, falling back to 6 if not defined?

MarkLark86 (approved):

assets/home/components/HomeApp.tsx
petrjasek: right fixed that

newsroom/wire/views.py
petrjasek: I've moved the config now to the backend so it knows which type is used and selects the size according to that

MarkLark86 (approved):

PR #863, opened 2024-03-26
Visual indication for Top stories in list, preview and full view

dzonidoo:
CPCN-715

## Checklist

### DIFF
--- a/assets/agenda/components/TopStoryLabel.tsx
+++ b/assets/agenda/components/TopStoryLabel.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import classNames from  'classnames';
+import {IAgendaItem} from 'interfaces';
+
+interface IProps {
+    item: IAgendaItem;
+    config: any;
+    size?: 'normal' | 'big';
+}
+
+export default function TopStoryLabel({item, config, size}: IProps) {
+    const classes = classNames('label label--fill label--rounded label--top-story', {
+        'label--big': size === 'big',
+        'mb-2': size === 'big',
+    });
+
+    return (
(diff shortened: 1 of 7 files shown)


assets/interfaces/agenda.ts
```diff
@@ -126,6 +126,7 @@ export interface IAgendaItem extends IResourceItem {
     watches?: Array<IUser['_id']>;
     es_highlight?: {[field: string]: Array<string>};
     state: string;
+    subject?: Array<{scheme?: Array<string> | string, topStoryScheme?: string, name: string}>;
```
petrjasek, line 129: this should be rather:
```ts
Array<{scheme?: string; code: string; name: string}>
```

assets/interfaces/configs.ts
```diff
@@ -21,6 +21,7 @@ export type IDisplayFieldsConfig = string |
 export interface IListConfig {
     subject?: {
         scheme?: Array<string> | string;
+        topStoryScheme?: string;
```
petrjasek, line 24: this should be also in the other configs

assets/agenda/components/TopStoryLabel.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import classNames from  'classnames';
+import {IAgendaItem} from 'interfaces';
+
+interface IProps {
+    item: IAgendaItem;
+    config: any;
```
petrjasek, line 7: this should be `IListConfig | IPreviewConfig;`

petrjasek (approved):

PR #868, opened 2024-03-29
configure search bar within filter [CPCN-627]

devketanpro:
## Checklist


assets/interfaces/config.ts
```diff
@@ -49,4 +49,5 @@ export interface IClientConfig {
     system_alert_recipients?: string;
     wire_time_limit_days: number;
     view_content_tooltip_email?: string;
+    searchbar_threshold_value?:number;
```
petrjasek, line 52: there should be a space after `:`

petrjasek (approved):

PR #873, opened 2024-04-03
implement an extension point to modify wire preview content

tomaskikutis:
TGA-80

## Checklist


petrjasek (commented):
what about importing an `extension` file in the `assets/index`  which would be resolved to cwd or to core where it would be empty so we wouldn't have to deal with webpack config in other repos

petrjasek (approved):

PR #877, opened 2024-04-09
convert time to user timezone in emails

petrjasek:
CPCN-753

## Checklist


newsroom/email.py
```diff
@@ -220,45 +225,82 @@ def get_language_template_name(template_name: str, language: str, extension: str
     return fallback_template_name
 
 
+def send_user_email(
```
MarkLark86, line 228: This function is almost identical to `send_template_email`. Would be good to extract the common code out to a util function and use that in both of these send email functions.

MarkLark86 (approved):

PR #878, opened 2024-04-10
FIX : images rendered in the list view [STTNHUB-332]

devketanpro:
Fixed issue where the `shortHighlightedtext` function was unintentionally converting `&lt;` entities to `< `characters, leading to improper rendering of images in the list view.

## Checklist

### DIFF
--- a/assets/wire/tests/WireApp.spec.tsx
+++ b/assets/wire/tests/WireApp.spec.tsx
@@ -148,10 +148,10 @@ describe('shortHighlightedtext', () => {
     });
 
     it('handles the scenario when the highlight text has images', () => {
-        const html = '<p>foo <span class="es-highlight">bar &lt;img src="image1.jpg"&gt; baz</span></p>';
+        const html = 'experience.&lt;img src="img.jpg" alt="Team Vapaus" /&gt;Norion<span class="es-highlight">Nordic</span> financing bank.';
         const maxLength = 40;
         const output = shortHighlightedtext(html, maxLength);
-        expect(output).toEqual('foo <span class="es-highlight">bar &lt;img src="image1.jpg"&gt; baz</span>...');
+        expect(output).toEqual('experience. Norion <span class="es-highlight">Nordic</span> financing bank....');
     });
 });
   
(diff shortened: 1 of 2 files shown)


assets/wire/tests/WireApp.spec.tsx
```diff
@@ -146,5 +146,12 @@ describe('shortHighlightedtext', () => {
         const output = shortHighlightedtext(html, maxLength);
         expect(output).toEqual('foo <span class="es-highlight">bar</span> foo <span class="es-highlight">bar</span> foo...');
     });
+
+    it('handles the scenario when the highlight text has images', () => {
+        const html = '<p>foo <span class="es-highlight">bar &lt;img src="image1.jpg"&gt; baz</span></p>';
+        const maxLength = 40;
+        const output = shortHighlightedtext(html, maxLength);
+        expect(output).toEqual('foo <span class="es-highlight">bar &lt;img src="image1.jpg"&gt; baz</span>...');
```
petrjasek: we should get the expected output without the image like `foo <span class="es-highlight">bar baz</span>`

devketanpro: okay

assets/wire/utils.ts
```diff
@@ -326,7 +326,9 @@ export function shortHighlightedtext(html: string, maxLength = 40) {
 
         lastNode = highlightSpan.nextSibling;
     }
-    return extractedText.join(' ')+(html.length > maxLength ? '...' : ' ');
+    return extractedText.join(' ')
+        .replace(/<img[^>]*src="([^"]*)"[^>]*>/g, ' ')
```
petrjasek, line 333: I wonder if we shouldn't change the code in the for loop, use there `getTextFromNode(highlightSpan)` instead of using the `highlightSpan.outerHTML` and wrap it in the `span` again, otherwise we might have the same issue with other tags

assets/wire/utils.ts
```diff
@@ -289,6 +289,9 @@ function getTextFromNode(node: Node | null): string {
     if (node.nodeType === Node.TEXT_NODE) {
         const textContent = node.textContent || '';
         return textContent.trim();
+    } else if (node.nodeType === Node.ELEMENT_NODE) {
+        const element = node as HTMLElement;
+        return element.innerText.trim();
```
petrjasek, line 294: not sure if we need that trim, does it add some extra spaces?

devketanpro: No, it doesn't add extra spaces; I'm just using this as a precautionary measure.

petrjasek (approved):

PR #880, opened 2024-04-10
implement extension point for copy embed

petrjasek:
replaces https://github.com/superdesk/newsroom-core/pull/873

## Checklist

### DIFF
--- a/assets/ui/components/ArticleBodyHtml.tsx
+++ b/assets/ui/components/ArticleBodyHtml.tsx
@@ -232,7 +232,7 @@ const mapDispatchToProps = (dispatch: any) => ({
     reportCopy: (item: any) => dispatch(selectCopy(item))
 });
 
-const ArticleBodyHtmlConnected: React.ComponentType<IProps> = connect(null, mapDispatchToProps)(ArticleBodyHtmlComponent);
+const ArticleBodyHtmlConnected: React.ComponentType<IOwnProps> = connect(null, mapDispatchToProps)(ArticleBodyHtmlComponent);
 
 // component needs to reinitialize if item changes
-export const ArticleBodyHtml: React.ComponentType<IProps> = (props) => <ArticleBodyHtmlConnected {...props} key={props.item._id} />;
+export const ArticleBodyHtml: React.ComponentType<IOwnProps> = (props) => <ArticleBodyHtmlConnected {...props} key={props.item._id} />;
--- a/assets/ui/components/ArticleBodyHtml.tsx
+++ b/assets/ui/components/ArticleBodyHtml.tsx
@@ -26,10 +26,15 @@ interface IMapDispatchToProps {
     reportCopy(item: IArticle): void;
 }
 
+interface IState {
+    error: boolean;
+}
+
 type IProps = IOwnProps & IMapDispatchToProps;
 
-class ArticleBodyHtmlComponent extends React.PureComponent<IProps> {
-    private getBodyHTML: any;
+const getBodyElement = memoize<(html: string, item: IArticle) => HTMLElement>(_getBodyElement);
+
+class ArticleBodyHtmlComponent extends React.PureComponent<IProps, IState> {
     private bodyRef: React.RefObject<HTMLDivElement>;
 
     constructor(props: any) {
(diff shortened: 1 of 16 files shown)


assets/ui/components/ArticleBodyHtml.tsx
```diff
@@ -187,49 +215,27 @@ class ArticleBodyHtml extends React.PureComponent<any, any> {
...
 const mapDispatchToProps = (dispatch: any) => ({
     reportCopy: (item: any) => dispatch(selectCopy(item))
 });
 
-const component: React.ComponentType<any> = connect(null, mapDispatchToProps)(ArticleBodyHtml);
+const ArticleBodyHtmlConnected: React.ComponentType<IOwnProps> = connect(null, mapDispatchToProps)(ArticleBodyHtmlComponent);
 
-export default component;
+// component needs to reinitialize if item changes
+export const ArticleBodyHtml: React.ComponentType<IOwnProps> = (props) => <ArticleBodyHtmlConnected {...props} key={props.item._id} />;
```
MarkLark86: Using `key=item._id` should be left up to the container component that renders this one, otherwise we're adding an extra branch in the render tree just to add a `key` property.

assets/ui/components/ArticleBodyHtml.tsx
```diff
@@ -187,49 +215,27 @@ class ArticleBodyHtml extends React.PureComponent<any, any> {
...
-        const html = this.getBodyHTML(
-            get(item, 'es_highlight.body_html.length', 0) > 0 ?
-                item.es_highlight.body_html[0] :
-                item.body_html
-        );
-
-        if (!html) {
-            return null;
-        }
+        // preview element will be populated in `componentDidMount`
```
MarkLark86: Moving this functionality outside of `render` to `componentDidMount` means the article body is not updated if a new version is published.  `getBodyHTML` is already memoized so this should not a performance issue

assets/ui/components/ArticleBodyHtml.tsx
```diff
@@ -22,21 +40,51 @@ class ArticleBodyHtml extends React.PureComponent<any, any> {
...
+                item.es_highlight?.body_html[0] :
+                item.body_html
+        );
+
+        if (!html) {
+            return;
+        }
+
+        const prepareWirePreview = extensions.prepareWirePreview ?? ((element) => element);
+        const previewElement = prepareWirePreview(
```
MarkLark86: We're already processing the HTML in the `_updateImageEmbedSources` function. This function manipulates the HTML to modify the `src` attribute of `img, video, audio` tags. The extension point we need provides similar functionality, to inspect and possibly modify the Article's HTML.

So that we can keep these functionalities grouped together, it might be good to add this extension point to `_getBodyHTML` instead, and use the [HTMLContent](https://github.com/superdesk/newsroom-core/blob/develop/assets/components/HTMLContent.jsx) element in the render function so we are not manipulating the DOM directly but instead letting React do that for us (and so `dompurify` is used). Considering `_getBodyHTML` is memoized, it will only get called **if** the Article's HTML is changed.

MarkLark86 (changes requested):

assets/ui/components/ArticleBodyHtml.tsx
petrjasek: removed that

assets/ui/components/ArticleBodyHtml.tsx
petrjasek: it's not using react for rendering to make the embed scripts work

assets/ui/components/ArticleBodyHtml.tsx
petrjasek: we can't use react for rendering due to embeds scripts, we can do some workarounds for certain embeds like we do for iframely but there is no generic fix afaik

MarkLark86: Understood

MarkLark86:
@petrjasek If a new version of the article is published to Newshub, does the preview automatically update? I believe this used to happen, but I'm not sure if that will work anymore.

petrjasek:
@MarkLark86 seems like we're not doing any auto update until user clicks on the refresh button

MarkLark86 (approved):

assets/ui/components/ArticleBodyHtml.tsx
tomaskikutis: Normally, components are expected to react to prop changes and update accordingly. Sometimes code-wise it is simpler to ignore prop changes and to force-reinitialize the component using a key. This is an implementation detail, thus a parent component shouldn't have to know about it.

I don't mind that you changed it Petr, I'm posting this comments so everyone understands the reasoning and we have an opportunity to discuss the pattern.

PR #891, opened 2024-04-22
TBC Label

dzonidoo:
https://sofab.atlassian.net/browse/CPCN-765

## Checklist


assets/agenda/components/AgendaPreview.tsx
```diff
@@ -90,7 +91,10 @@ class AgendaPreview extends React.PureComponent<any, any> {
                         )}
                     >
                         <hgroup>
-                            <TopStoryLabel item={item} config={previewConfig} size='big' />
+                            <div className='d-flex gap-2'>
```
petrjasek: might be useful to have some `LabelGroup` component for this

assets/agenda/components/ToBeConfirmedLabel.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import classNames from  'classnames';
+import {IAgendaItem, IListConfig, IPreviewConfig} from 'interfaces';
+import {isTopStory} from 'agenda/utils';
```
petrjasek: there are a few unused imports

assets/agenda/components/AgendaPreview.tsx
petrjasek: maybe even some `AgendaLabels` component otherwise it seems like duplicated code all over

assets/agenda/components/ToBeConfirmedLabel.tsx
```diff
@@ -0,0 +1,26 @@
...
+import {gettext} from 'utils';
+
+interface IProps {
+    item: IAgendaItem;
+    size?: 'normal' | 'big';
+}
+
+export default function ToBeConfirmedLabel({item, size}: IProps) {
+    const classes = classNames('label label--rounded label--orange', {
```
petrjasek, line 12: not sure if `label--orange` makes sense in regards to theming, `label--warning` might be better

assets/agenda/components/AgendaPreview.tsx
dzonidoo: so do i need to add some LableGroup component in this pr?

assets/agenda/components/ToBeConfirmedLabel.tsx
dzonidoo: ahh i miss that one

assets/agenda/components/ToBeConfirmedLabel.tsx
dzonidoo: its not my class name... i just use the firtz'es one

petrjasek: what do you think @fritzSF ?

assets/agenda/components/AgendaPreview.tsx
dzonidoo: @petrjasek what about this one?

assets/agenda/components/ToBeConfirmedLabel.tsx
fritzSF: @petrjasek I agree with you, but this is how the class names for the labels are set up at the moment. It's nothing new, it is like that for the last six years. We should change to smenatic names at some point for sure.

assets/agenda/components/AgendaPreview.tsx
petrjasek: yep I would do that to begin with

assets/agenda/components/LabelGroup.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    orientation?: 'horizontal' | 'vertical'; // defaults to 'horizontal'
+    justify?: 'start' | 'end' | 'center'; // defaults to 'start'
+    align?: 'start' | 'end' | 'center'; // defaults to 'center'
+    children: React.ReactNode;
+    className?: string;
```
petrjasek: imo this is something we want to avoid in general

assets/agenda/components/LabelGroup.tsx
petrjasek, line 1: I think this should be rather in `ui/components` there is nothing agenda specific

assets/agenda/components/LabelGroup.tsx
dzonidoo: can u be more specific? 😄

petrjasek: allowing some extra classNames used

dzonidoo: now is clear

petrjasek (approved):

assets/agenda/components/AgendaListItem.tsx
```diff
@@ -304,10 +304,10 @@ class AgendaListItem extends React.Component<IProps> {
             >
                 <div className={classes.wrap} tabIndex={0}>
                     <div className={classes.article} key="article">
-                        <LabelGroup>
+                        <Spacer gap='4' noGrow>
```
petrjasek: I would probably still like better that `LabelGroup` with `Spacer` inside, I don't like that duplicated code `gap='4' noGrow` everywhere

dzonidoo: but what will that component do? GroupLabel?

dzonidoo: i dont see the point to have two similar component

petrjasek: it's about code quality, functionality wise it's the same

petrjasek: so if we would like to have there `gap='5'` we won't have to look all over the place, but you just open LabelGroup component and change it there

dzonidoo: so u want to have LabelGroup component back with Spacer inside.. and just copy/paste props from spacer right?

petrjasek: I think the LabelGroup should be without props pretty much, just doing:
```ts
const LabelGroup = ({children}) => (
    <Spacer gap='4' nowrap>{children}</Spacer>
);
```

fritzSF: Hmm, I'm really not sure about that @petrjasek . We might overdo it sometimes with options, but with props, you can allow only a limited number of options defined by the design. The gaps in Label Group should maybe be 4 and 8, that's it. In some other use cases, such as Cards for instance, it needs to be higher. A generic component does not provide any constraints, so a developer can add `<Spacer gap='16' nowrap>{children}</Spacer>` for labels which we don't want.

petrjasek: I'm not saying we shouldn't use the Spacer in other places with different values, it's only for this particular usecase with labels inside

petrjasek (approved):

assets/agenda/components/AgendaListItem.tsx
fritzSF: This would mean that we wouldn't be able to target that specific value without affecting other places where the spacer is used with the same value. The Label Group here is not a good example; it wouldn't be a big deal if we had just one predefined value for it. However, I'm against that approach as a general solution. Sure, there is a place for such a generic component in many applications, but it is not a silver bullet. There is a need to target just one specific element/component in many scenarios. Using such a generic spacer for everything would make that impossible. We very often have to change the default style based on placement, screen size, device, etc. How would you do that if everything is based on the same CSS? 
Anyway, let's move the conversation elsewhere; I will give you multiple examples where this approach will not work. For this specific case, we are good to go either way.

PR #893, opened 2024-04-26
CPCN-194

darconny:
## Checklist

- [x ] This pull request is not adding new forms that use redux


petrjasek:
Wire - Topic -- Delete a folder with content in My Wire Topic (failed)

this makes the dropdown menus in modal not visible

darconny:
@petrjasek should be fixed now, pls check

petrjasek:
I think it is, otherwise the e2e tests would be failing

petrjasek (approved):

PR #896, opened 2024-05-01
[STTNHUB-340] ui: Make AgendaCoverage preview fields configurable

MarkLark86:
## Checklist


assets/agenda/components/preview/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    coverage_update_coming: CoverageUpdateComing,
+    coverage_status: CoverageStatus,
+    coverage_scheduled_status: CoverageScheduledStatus,
+    coverage_ednote: CoverageEdnote,
+    coverage_workflow_reason: CoverageWorkflowReason,
+    coverage_internal_note: CoverageInternalNote,
+    coverage_actions: CoverageActions,
+};
+
+export function registerCoveragePreviewFieldComponent(field: string, component: React.ComponentType<ICoverageMetadataPreviewProps>) {
```
petrjasek: I wonder if we need to have a specific function for just preview, maybe `registerCoverageFieldComponent` would do, I was thinking about something even more generic but I suppose it's useful to be able to provide specific props for coverages/events/plannings/wire so we can probably keep that

petrjasek (approved):

PR #898, opened 2024-05-09
use topic navigation for filtering when sending notifications

petrjasek:
CPCN-763

## Checklist

### DIFF
--- a/newsroom/search/service.py
+++ b/newsroom/search/service.py
@@ -674,7 +674,7 @@ class BaseSearchService(Service):
 
             return ObjectId(product) in user_specific_products or ObjectId(product) in company_products_with_zero_seats
 
-    def apply_section_filter(self, search, filters: Optional[List[SectionFilter]] = None):
+    def apply_section_filter(self, search: SearchQuery, filters: Optional[List[SectionFilter]] = None) -> None:
         """Generate the section filter
 
         :param SearchQuery search: the search query instance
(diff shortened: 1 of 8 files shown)


newsroom/search/service.py
```diff
@@ -949,7 +950,7 @@ def apply_topic_args(self, topic, args=None) -> SearchArgs:
 
         return args
 
-    def get_topic_query(self, topic, user, company, section_filters=None, query=None, args=None):
+    def get_topic_query(self, topic: Topic, user, company, section_filters=None, query=None, args=None):
```
devketanpro: could it be beneficial to add types for all params?

devketanpro (approved):

newsroom/search/service.py
petrjasek: yep added the other ones

newsroom/search/service.py
```diff
@@ -674,7 +674,7 @@ def is_validate_product(self, data):
 
             return ObjectId(product) in user_specific_products or ObjectId(product) in company_products_with_zero_seats
 
-    def apply_section_filter(self, search, filters=None):
+    def apply_section_filter(self, search, filters: Optional[List[SectionFilter]] = None):
```
devketanpro: you missed here :wink: !

devketanpro (approved):

PR #900, opened 2024-05-14
Hiding importance filter from clients that have restricted coverage information [STTNHUB-327]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/utils.py
+++ b/newsroom/utils.py
@@ -670,6 +670,14 @@ def get_groups(company: Optional[Company], groups: List[Group]) -> List[Group]:
     """
     filter agenda groups based on restrict_coverage_info permission
     """
-    if company and company.get("restrict_coverage_info"):
-        return [group for group in groups if "restrict_coverage_info" not in group.get("permissions", [])]
-    return groups
+    return [group for group in groups if has_permissions(group, company)]
+
+
+def has_permissions(group: Group, company: Optional[Company]) -> bool:
+    group_permissions = group.get("permissions", [])
+    if group_permissions:
+        for permission in group_permissions:
+            if company and company.get(permission):
+                return False
+        return True
+    return True
--- a/newsroom/utils.py
+++ b/newsroom/utils.py
@@ -20,7 +20,7 @@ from eve_elastic.elastic import parse_date, ElasticCursor
 from flask import current_app as app, json, abort, request, g, flash, session, url_for
 from flask_babel import gettext
 
-from newsroom.types import PublicUserData, User, Company, Group
+from newsroom.types import PublicUserData, User, Company, Group, Permissions
 from newsroom.template_filters import (
     time_short,
     parse_date as parse_short_date,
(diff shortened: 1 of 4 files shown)


newsroom/utils.py
```diff
@@ -664,3 +664,13 @@ def parse_objectid(value: Union[str, ObjectId]) -> Union[str, ObjectId]:
...
         return value
+
+
+def get_groups(company: Optional[Company], groups: list) -> list:
+    """
+    filter agenda groups based on restrict_coverage_info
+    # STTNHUB - 327
+    """
+    if company and company.get("restrict_coverage_info"):
+        return [item for item in groups if item["field"] != "stturgency"]
```
petrjasek: I wouldn't use `stturgency` in core, rather we might add something like
```py
Permissions = Literal["restrict_coverage_info"]

class Group(TypedDict):
    permissions: List[Permissions]
```
and then in newsroom-app-stt repo in the stturgency group definition:
```py
permissions=["restrict_coverage_info"]
```

petrjasek (changes requested):

newsroom/utils.py
```diff
@@ -666,11 +666,10 @@ def parse_objectid(value: Union[str, ObjectId]) -> Union[str, ObjectId]:
...
-def get_groups(company: Optional[Company], groups: list) -> list:
+def get_groups(company: Optional[Company], groups: List[Group]) -> List[Group]:
     """
-    filter agenda groups based on restrict_coverage_info
-    # STTNHUB - 327
+    filter agenda groups based on restrict_coverage_info permission
     """
     if company and company.get("restrict_coverage_info"):
-        return [item for item in groups if item["field"] != "stturgency"]
+        return [group for group in groups if "restrict_coverage_info" not in group.get("permissions", [])]
```
petrjasek: it would be better to do the logic like:
```py
def get_groups(...):
    return [has_permissions(group, company) for group ...]

def has_permissions(group, company):
    group_permissions = group.get("permissions", [])
    if group_permissions:
        for permission in group_permissions:
            if not company or not company.get(permission):
                return False
        return True
    return True
```

newsroom/utils.py
```diff
@@ -670,6 +670,14 @@ def get_groups(company: Optional[Company], groups: List[Group]) -> List[Group]:
...
-    return groups
+    return [group for group in groups if has_permissions(group, company)]
+
+
+def has_permissions(group: Group, company: Optional[Company]) -> bool:
+    group_permissions = group.get("permissions", [])
+    if group_permissions:
+        for permission in group_permissions:
+            if company and company.get(permission):
+                return False
```
petrjasek: this would return `False` if the company has the permission, it should be the other way around

petrjasek: would be ideal to add some unittest for this method

devketanpro: so this scenario according to our AC, as we have to hide that filter when 
company `permission[restrict_coverage_info] = true`

petrjasek: my bad, I realized now we store it the other way around.
would be good to do something like this for future:
```py
Permissions = Literal["coverage_info"]

def get_company_permissions(company: Company) -> Dict[Permissions, bool]:
    return {"coverage_info": not company or company.get("restrict_coverage_info") != True}

def get_groups(groups, company):
    company_permissions = get_company_permissions(company)
    return [group for group in groups if not group.get("permissions") or all([
        company_permissions[permission] for permission in group["permissions"]
    ])]
```
we might change the way how we store what's enabled for company

petrjasek (approved):

PR #901, opened 2024-05-15
implement ui config for hiding item type filters in agenda

petrjasek:
STTNHUB-238

## Checklist


assets/interfaces/configs.ts
```diff
@@ -79,11 +79,19 @@ export interface IAgendaPreviewConfig extends IPreviewConfig {
...
+    events_only?: boolean;
+    planning_only?: boolean;
+}
+
 export interface IAgendaUIConfig extends IBaseUIConfig {
     _id: 'agenda';
     open_coverage_content_in_same_page?: boolean;
     subnav?: {
-        filters?: Array<string>;
+        filters?: Array<AgendaFilterTypes>;
```
MarkLark86, line 93: It might be good if we were to have these different UI configs documented somewhere, that way PMs are able to get familiar with the different configurations available.

MarkLark86 (approved):

assets/interfaces/configs.ts
petrjasek: added some initial docs, will do the setup on readthedocs later

PR #902, opened 2024-05-15
update superdesk-core

petrjasek:


petrjasek:
hi @eos87 can you check if this will work on mac?

eos87 (approved):
I tested this locally, it works great. Thanks!

PR #904, opened 2024-05-16
[CPCN-177] - Add `Source` column to Content Activity report

eos87:
### Purpose
This PR introduces a new column to the Content Activity report.

### What has changed
- Added a new column (`source`) to the report both UI and CSV export
- Removed the usage of `lodash.get` from the `ContentActivity.tsx` component
- Completed typings for the properties of the component
- Remove the usage of `PropTypes` in favour of Typescript interfaces
- Little code clean up 

### Steps to test
- Checkout the branch
- Run the project and go to Reports > Content Activity
- Run the report
- Notice that the column has been added and everything else is working as expected

### Checklist

Thanks for reviewing! 

Resolves: [CPCN-177]

[CPCN-177]: https://sofab.atlassian.net/browse/CPCN-177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/company-reports/components/ContentActivity.tsx
```diff
@@ -77,12 +121,12 @@ class ContentActivity extends React.Component<any, any> {
...
     }
 
-    componentWillMount() {
+    UNSAFE_componentWillMount() {
         // Fetch the genre & company aggregations to populate those dropdowns
         this.props.fetchAggregations(REPORTS['content-activity']);
     }
 
-    componentWillReceiveProps(nextProps: any) {
+    UNSAFE_componentWillReceiveProps(nextProps: any) {
```
petrjasek: here it seems like it's doing some derived stated, would be good to [avoid it](https://legacy.reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization)

assets/company-reports/components/ContentActivity.tsx
```diff
@@ -77,12 +121,12 @@ class ContentActivity extends React.Component<any, any> {
         };
     }
 
-    componentWillMount() {
+    UNSAFE_componentWillMount() {
```
petrjasek: I think you can use rather `componentDidMount`

petrjasek (commented):
looks good, but would rather fix those 2 lifecycle methods

assets/company-reports/components/ContentActivity.tsx
eos87: yeah, make sense. I'll do the adjustments.

eos87: done

eos87:
Hey @petrjasek, I've made a few changes here: 
- Removed the unnecessary internal state from `ContentActivity.tsx` and used props directly instead 
- Extracted the report filters into a separate component to try to simplify things
- A couple of minor UI/UX improvements discussed earlier today with Fritz and Anna. See screenshots below

Added small label "Select a report" instead of an "empty" dropdown
image

Fixed alignment and overlapping of the date selector. Also removed the double arrows in the dropdown (those at the beginning)
image

Please take another look when you have some time. Thanks!

petrjasek (approved):
looks good

PR #910, opened 2024-05-22
Pause ALL my notifications

dzonidoo:
https://sofab.atlassian.net/browse/CPCN-186

## Checklist


tomaskikutis:
@thecalcc can you review this?

assets/components/NotificationList.tsx
```diff
@@ -3,12 +3,17 @@ import PropTypes from 'prop-types';
...
 import {isTouchDevice} from '../utils';
 import NotificationListItem from './NotificationListItem';
+import {postNotificationSchedule} from 'user-profile/actions';
+
+interface IState {
+    displayItems: boolean;
+    connected: boolean;
+}
 
-class NotificationList extends React.Component<any, any> {
```
thecalcc, line 11: If it's quick, can you type the props as well?

assets/components/NotificationList.tsx
```diff
@@ -127,8 +153,32 @@ class NotificationList extends React.Component<any, any> {
...
+                }
+
+                <span
+                    className={classNames(
+                        'navbar-notifications__inner-circle',
+                        {'navbar-notifications__inner-circle--disconnected': !this.state.connected}
+                    )}
+                    ref={(elem: any) => this.elem = elem}
+                    title={gettext('Notifications')}>
+                    <h3 className="a11y-only">Notification bell</h3>
```
thecalcc: add gettext here, we shouldn't care if it's hidden

assets/notifications/actions.ts
```diff
@@ -1,15 +1,21 @@
...
 
 export const UPDATE_NOTIFICATION_COUNT = 'UPDATE_NOTIFICATION_COUNT';
 export function updateNotificationCount(count: any) {
     return {type: UPDATE_NOTIFICATION_COUNT, count};
 }
 
 export const INIT_DATA = 'INIT_DATA';
-export function initData(data: any) {
-    return {type: INIT_DATA, data};
+export function initData(notificationData: any, profileData: any) {
```
thecalcc, line 12: can you type props here?

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    onSubmitForm(event: React.FormEvent<HTMLFormElement>) {
+        event.stopPropagation();
+        event.preventDefault();
+
+        if (this.formRef.current == null) {
+            throw new Error('ref missing');
+        }
+
```
thecalcc, line 57: is this meant to stop the function execution? What does throwing an error serve us here?

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+        const inputs = this.formRef.current.querySelectorAll('input') as NodeListOf<HTMLInputElement>;
+
+        const hasErrors = [...inputs].some(input => {
+            return input.checkValidity() !== true;
+        });
```
thecalcc: what's the reason here to spread inputs?

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+        const stateClone: IState = {};
+
+        stateClone[state] = event?.target.value;
+        this.setState(stateClone);
+    }
+
+    disabledMinOption(state: string | undefined) {
+        if (state != '' && state != undefined) {
+            const newMaxDate = new Date(state);
+            newMaxDate.setDate(newMaxDate.getDate() + 1);
```
thecalcc: you can do this in one line, but if you think readability is better this way, let's keep it.

Add `1` to a constant that explains what its purpose is, we can't have magic numbers

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+            return newMaxDate.toISOString().split('T')[0];
+        } else {
+            return disabledOptionDateFrom;
+        }
+    }
+
+    disabledMaxOption(state: string | undefined) {
+        if (state != '' && state != undefined) {
+            const newMaxDate = new Date(state);
+            newMaxDate.setDate(newMaxDate.getDate() - 1);
```
thecalcc: same goes for here, I bet 1 is some kind of offset, I understand it but others would need to know also

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+                    >
+                        <div className='d-flex align-items-center gap-5'>
+                            <div className="d-flex flex-column align-items-start">
+                                <label htmlFor="date-from">{gettext('From')}</label>
+                                <input
+                                    id="date-from"
+                                    type="date"
+                                    name="date-from"
+                                    className="form-control"
+                                    onChange={(event) => this.updateDate(event, 'pauseFrom')}
```
thecalcc: looks like we're expecting a return from this function which isn't the case, can you add curly braces to it?

thecalcc (approved):

tomaskikutis (changes requested):
@thecalcc you shouldn't approve a PR until your comments are addressed/answered.

I'll also block the merging so after you finish the review, I'll take a look too.

thecalcc:
> @thecalcc you shouldn't approve a PR until your comments are addressed/answered.
> 
> I'll also block the merging so after you finish the review, I'll take a look too.

I clicked approve by mistake, my bad

thecalcc (commented):
Would component prop types take too long to be properly typed?

dzonidoo:
Actually yes...

thecalcc:
> Actually yes...

@dzonidoo ok got it. What about not adding new redux forms, is it possible that we don't use redux?

dzonidoo:
@thecalcc its not.. we need it

assets/components/NotificationList.tsx
```diff
@@ -3,12 +3,17 @@ import PropTypes from 'prop-types';
...
 import NotificationListItem from './NotificationListItem';
+import {postNotificationSchedule} from 'user-profile/actions';
+
+interface IState {
+    displayItems: boolean;
+    connected: boolean;
+}
 
-class NotificationList extends React.Component<any, any> {
+class NotificationList extends React.Component<any, IState> {
```
tomaskikutis: Add types for props. It doesn't have to be perfect - it's better to leave some `any` if you can't understand it from usage - than not have any types at all. Adding types for existing stuff is nice to have, but typing your new code is a must. And you can't type your new code if an interface for props doesn't exist at all.

assets/components/NotificationList.tsx
```diff
@@ -70,45 +75,66 @@ class NotificationList extends React.Component<any, any> {
...
-                    )}
-                    ref={(elem: any) => this.elem = elem}
-                    title={gettext('Notifications')}>
-                    <h3 className="a11y-only">Notification bell</h3>
-                    <i className='icon--alert' onClick={this.toggleDisplay} />
-                </span>
-
-                {!this.state.displayItems ? null : this.props.count === 0 ? (
+        const notificationPopUp = () => {
+            if (this.props.fullUser.notification_schedule?.pauseFrom != '' && this.props.fullUser.notification_schedule?.pauseFrom != '') {
```
tomaskikutis: you're checking the same thing twice

assets/components/NotificationList.tsx
```diff
@@ -70,45 +75,66 @@ class NotificationList extends React.Component<any, any> {
...
                     <div className="notif__list dropdown-menu dropdown-menu-right show">
                         <div className='notif__list__header d-flex'>
                             <span className='notif__list__header-headline ms-3'>{gettext('Notifications')}</span>
                         </div>
-                        <div className='notif__list__message'>
-                            {gettext('No new notifications!')}
+
+                        <div className='d-flex flex-column gap-2 p-3'>
+                            <div className='nh-container nh-container__text--alert p-2'>
+                                {gettext('All notifications are paused until {{pauseTo}}', {pauseTo: formatDate(this.props.fullUser.notification_schedule?.pauseTo)})}
```
tomaskikutis: use more clear placeholders inside `gettext` - `All notifications are paused until {{date}}` .

Also you shouldn't be using optional chaining here, because you already checked the value not to be null before.

assets/user-profile/actions.ts
```diff
@@ -356,6 +364,26 @@ export function updateUserNotificationSchedules(schedule: Omit<IUser['notificati
     };
 }
 
+export function postNotificationSchedule(userId: string, schedule: Omit<IUser['notification_schedule'], 'last_run_time'>):Promise<void> {
+    return server.post(`/users/${userId}/notification_schedules`, schedule)
+        .then(() => {
+            notify.success(gettext('pause notification updated'));
```
tomaskikutis: It's not correct gramatically. Use `Notifications paused` / `Notifications resumed`

assets/components/NotificationList.tsx
```diff
@@ -127,8 +153,33 @@ class NotificationList extends React.Component<any, any> {
...
                             ))
                         )}
+                    </div>;
+                }
+            }
+        };
+        
+        return (
+            <div className="navbar-notifications__inner">
+                <h3 className="a11y-only">Notification Bell</h3>
```
tomaskikutis: This needs to be translated. Also call it `Notifications` and extract it into a constant because there are multiple usages.

assets/components/NotificationList.tsx
```diff
@@ -127,8 +153,33 @@ class NotificationList extends React.Component<any, any> {
...
+                        {'navbar-notifications__inner-circle--disconnected': !this.state.connected}
+                    )}
+                    ref={(elem: any) => this.elem = elem}
+                    title={gettext('Notifications')}
+                >
+                    <h3 className="a11y-only">{gettext('Notification bell')}</h3>
+                    <i className='icon--alert' onClick={this.toggleDisplay} />
+                </span>
+
+                {this.state.displayItems && notificationPopUp()}
```
tomaskikutis: inline the code from `notificationPopUp`

assets/user-profile/components/profile/UserProfile.tsx
```diff
@@ -180,6 +182,21 @@ class UserProfile extends React.PureComponent<IProps> {
                                     </button>
                                 </div>
                             </div>
+                            
+                            <div className='row'>
```
tomaskikutis, line 188: [this](https://github.com/superdesk/newsroom-core/assets/2076953/ab5c2427-6ba8-42ce-ac04-2e7cfbe9a136) part of the design is not implemented

assets/notifications/actions.ts
```diff
@@ -108,3 +114,20 @@ export function pushNotification(push: any): any {
         }
     };
 }
+
+export function updateUserNotificationPause() {
```
tomaskikutis: you need to improve naming here, this function is not doing anything related to pausing notificatons

assets/components/NotificationList.tsx
```diff
@@ -70,45 +75,66 @@ class NotificationList extends React.Component<any, any> {
...
+
+                        <div className='d-flex flex-column gap-2 p-3'>
+                            <div className='nh-container nh-container__text--alert p-2'>
+                                {gettext('All notifications are paused until {{pauseTo}}', {pauseTo: formatDate(this.props.fullUser.notification_schedule?.pauseTo)})}
+                            </div>
+
+                            <button
+                                type="button"
+                                className="nh-button nh-button--small nh-button--tertiary"
+                                onClick={() => {
```
tomaskikutis: adjust this to be

```
onClick={() => {
    this.props.resumeNotifications();
}}
```

and update props accordingly - meaning you would execute the actual logic from a parent component.

assets/components/NotificationList.tsx
```diff
@@ -70,45 +75,66 @@ class NotificationList extends React.Component<any, any> {
...
-                                onClick={this.props.clearAll}>{gettext('Clear All')}
+                                onClick={this.props.clearAll}
+                            >
+                                {gettext('Clear All')}
                             </button>
                         </div>
+
+                        <div className='p-3'>
+                            <div className='nh-container nh-container__text--info p-2'>
+                                {gettext('All notifications are set to be paused from {{pauseFrom}} to {{pauseTo}}', {pauseFrom: formatDate(this.props.fullUser.notification_schedule?.pauseFrom), pauseTo: formatDate(this.props.fullUser.notification_schedule?.pauseTo)})}
```
tomaskikutis: You need to be more careful with optional chaining. If you use it - it means the value can be undefined. And if you call `formatDate(undefined)` what result do you expect in that case? What kind of string the user would see then?

assets/notifications/actions.ts
```diff
@@ -108,3 +114,20 @@ export function pushNotification(push: any): any {
...
+    return (dispatch: any, getState: any) => {
+        const fullUser = getState().fullUser;
+        return dispatch(fetchUser(fullUser._id));
```
tomaskikutis: Avoid nested dispatches. Inline `server.get` here instead of extracting it to another function.

assets/notifications/components/NotificationsApp.tsx
```diff
@@ -26,13 +27,17 @@ class NotificationsApp extends React.Component<any, any> {
                 clearAll={this.props.clearAll}
                 loadNotifications={this.props.loadNotifications}
                 loading={this.props.loading}
+                fullUser={this.props.fullUser}
+                user={this.props.user}
```
tomaskikutis: what's the purpose to pass both - user and fullUser?

assets/user-profile/actions.ts
```diff
@@ -356,6 +364,26 @@ export function updateUserNotificationSchedules(schedule: Omit<IUser['notificati
     };
 }
 
+export function postNotificationSchedule(userId: string, schedule: Omit<IUser['notification_schedule'], 'last_run_time'>):Promise<void> {
```
tomaskikutis: move this to some helpers file. now it's inside user store(actions/reducer). Stores should be isolated and not import stuff from each other. Especially since this is not an action nor reducer, but a helper function.

assets/user-profile/actions.ts
```diff
@@ -356,6 +364,26 @@ export function updateUserNotificationSchedules(schedule: Omit<IUser['notificati
...
+    return server.post(`/users/${userId}/notification_schedules`, schedule)
+        .then(() => {
+            notify.success(gettext('pause notification updated'));
+        });
+}
+
+export function updateUserNotificationPause(schedule: Omit<IUser['notification_schedule'], 'last_run_time'>) {
```
tomaskikutis: This function is almost identical to `updateUserNotificationSchedules`. Use the one that existed already and add some arguments if you need it to be slightly different.

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+    modalFormInvalid(): void;
+    modalFormValid(): void;
+    updateUserNotificationPause(schedule: Omit<IUser['notification_schedule'], 'last_run_time'>): void;
+    data: {
+        user: IUser;
+    };
+}
+
+interface IState {
+    pauseFrom?: string;
```
tomaskikutis: if you are applying defaults in the constructor, should properties not be mandatory then?

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+        const inputs = this.formRef.current.querySelectorAll('input') as NodeListOf<HTMLInputElement>;
+
+        const hasErrors = [...inputs].some(input => {
+            return input.checkValidity() !== true;
+        });
+
+        if (hasErrors === true) {
+            this.formRef.current.reportValidity();
+        } else {
+            this.props.updateUserNotificationPause(this.state);
```
tomaskikutis: check types of this function, I pass 'abc' as argument and it still compiles.

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    updateDate(event: any, state: 'pauseFrom' | 'pauseTo') {
+        const stateClone: IState = {};
```
tomaskikutis: where are you cloning anything? naming should be better here.

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+        });
+
+        if (hasErrors === true) {
+            this.formRef.current.reportValidity();
+        } else {
+            this.props.updateUserNotificationPause(this.state);
+        }
+    }
+
+    updateDate(event: any, state: 'pauseFrom' | 'pauseTo') {
```
tomaskikutis: don't call the argument `state` not to confuse it with component state.

Also `'pauseFrom' | 'pauseTo'` isn't a state of anything so naming is definitelly off.

I see you use `any` type here as well - go over the entire PR and ensure you are not using `any` in your newly written code.

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+                                <input
+                                    id="date-to"
+                                    type="date"
+                                    name="date-to"
+                                    className="form-control"
+                                    onChange={(event) => {
+                                        this.updateDate(event, 'pauseTo');
+                                    }}
+                                    value={this.state.pauseTo}
+                                    min={this.disabledMinOption(this.state.pauseFrom)}
```
tomaskikutis: improve naming - "min option" can neither be enabled nor disabled.

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+                <div className="nh-container nh-container--highlight rounded--none">
+                    <form
+                        ref={this.formRef}
+                        onSubmit={(event) => this.onSubmitForm(event)}
+                    >
+                        <div className='d-flex align-items-center gap-5'>
+                            <div className="d-flex flex-column align-items-start">
+                                <label htmlFor="date-from">{gettext('From')}</label>
+                                <input
+                                    id="date-from"
```
tomaskikutis: use variables for IDs

assets/user-profile/components/PauseNotificationModal.tsx
tomaskikutis: There are a few reasons for this. One is to stop the execution in case it is `null` and another is to tell compiler that if execution is not stopped, it should be treated as `non-null`.

We could also achieve this using `if/else` or early return. But - it is preferrable to throw errors when we're guarding against invalid application states.

assets/user-profile/components/PauseNotificationModal.tsx
tomaskikutis: `querySelectorAll` returns a collection which is not an array and doesn't have array methods. @dzonidoo use `Array.from(inputs)` instead.

assets/user-profile/components/PauseNotificationModal.tsx
tomaskikutis: when `0` and `1` are used, most of the times it's obvious what's being done thus 0/1 are not considered magic numbers.

tomaskikutis (commented):
Thanks for taking a look @thecalcc, I'll take over from here. I wish you did a more in-depth job, I still have lots of comments after your review.

@dzonidoo there's a bug I found not related to this PR - do apply the following patch to fix it. (it was crashing when notification schedule isn't set)
image

tomaskikutis:
> What about not adding new redux forms, is it possible that we don't use redux?

where did you see Redux being used for forms @thecalcc ? If data that you gather from a form is being sent to a redux action, it doesn't make it a redux based form. Those are when validation and form data while typing live inside redux store.

assets/helpers/notification.tsx
```diff
@@ -0,0 +1,10 @@
+import {IUser} from 'interfaces';
+import server from 'server';
+import {gettext, notify} from 'utils';
+
+export function postNotificationSchedule(userId: string, schedule: Omit<IUser['notification_schedule'], 'last_run_time'>, message: string):Promise<void> {    
+    return server.post(`/users/${userId}/notification_schedules`, schedule)
+        .then(() => {
+            notify.success(gettext(message));
```
tomaskikutis: Don't pass a variable to `gettext`. You should be receiving a translated string as an argument here already.

assets/components/NotificationList.tsx
```diff
@@ -72,7 +91,7 @@ class NotificationList extends React.Component<any, any> {
     render() {
         return (
             <div className="navbar-notifications__inner">
-                <h3 className="a11y-only">Notification Bell</h3>
+                <h3 className="a11y-only">{gettext('{{label}}', {label: NOTIFICATION_ARIA_LABEL})}</h3>
```
tomaskikutis: Placeholders aren't being translated by design. Thus it doesn't make to use `gettext` at all if the only thing you have inside a string is a placeholder.

assets/components/NotificationList.tsx
```diff
@@ -85,63 +104,96 @@ class NotificationList extends React.Component<any, any> {
...
-                        ) : (
-                            this.props.notifications.map((notification: any) => (
-                                <NotificationListItem
-                                    key={get(this.props.items, `${notification.item}._id`, 'test')}
-                                    item={get(
-                                        this.props.items,
-                                        `${notification.item}`,
-                                        get(notification, 'data.item', {})
+                {this.state.displayItems
+                    && (() => {
```
tomaskikutis: move `this.state.displayItems` check inside the function as a first block. It will reduce indentation/nesting

`if (this.state.displayItems !== true) {
    return null;
}`

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+                </div>
+            </Modal>
+        );
+    }
+}
+
+const mapDispatchToProps = (dispatch: any) => ({
+    modalFormInvalid: () => dispatch(modalFormInvalid()),
+    modalFormValid: () => dispatch(modalFormValid()),
+    updateUserNotificationPause: (schedule: Omit<IUser['notification_schedule'], 'last_run_time'>, message: string) => (
```
tomaskikutis: adjust naming so if you only dispatch `updateUserNotificationSchedule` here, what about updating the prop name to be the same?

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+        user: IUser;
+    };
+}
+
+interface IState {
+    pauseFrom: string;
+    pauseTo: string;
+}
+
+const disabledOptionDateFrom = new Date().toISOString().split('T')[0];
```
tomaskikutis: improve naming, what about `dateToday`?

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+        } else {
+            this.props.updateUserNotificationPause(this.state, gettext('Notifications paused'));
+        }
+    }
+
+    updateDate(event: React.ChangeEvent<HTMLInputElement>, pauseType: 'pauseFrom' | 'pauseTo') {
+        this.setState({...this.state, [pauseType]: event.target.value});
+    }
+
+    disabledPrevOptions(state: string | undefined) {
```
tomaskikutis: Improve naming - this method does not disable anything - only returns a date. A good name would describe what kind of date does it return.

assets/components/NotificationList.tsx
```diff
@@ -85,63 +104,96 @@ class NotificationList extends React.Component<any, any> {
...
+                                </div>;
+                            }
+                        }})()
+                    }
             </div>
         );
     }
 }
 
-NotificationList.propTypes = {
```
tomaskikutis, line 137: when you're removing something, search to make sure you're not leaving pieces. For example here you left the import and static declaration inside the class.

assets/components/NotificationList.tsx
```diff
@@ -85,63 +104,96 @@ class NotificationList extends React.Component<any, any> {
...
+                        if (this.props.fullUser.notification_schedule?.pauseFrom != '' && this.props.fullUser.notification_schedule?.pauseTo != '') {
+                            return (
+                                <div className="notif__list dropdown-menu dropdown-menu-right show">
+                                    <div className='notif__list__header d-flex'>
+                                        <span className='notif__list__header-headline ms-3'>{gettext('Notifications')}</span>
+                                    </div>
+            
+                                    <div className='d-flex flex-column gap-2 p-3'>
+                                        <div className='nh-container nh-container__text--alert p-2'>
+                                            {gettext('All notifications are paused until {{date}}', {date: formatDate(this.props.fullUser.notification_schedule?.pauseTo)})}
```
tomaskikutis: when I comment on stuff, do check other usages not only the line I posted on. See my other comment about passing undefined to `formatDate`

assets/components/NotificationList.tsx
```diff
@@ -85,63 +104,96 @@ class NotificationList extends React.Component<any, any> {
...
+                                            className="button-pill ms-auto me-3"
+                                            onClick={this.props.clearAll}
+                                        >
+                                            {gettext('Clear All')}
+                                        </button>
+                                    </div>
+            
+                                    <div className='p-3'>
+                                        <div className='nh-container nh-container__text--info p-2'>
+                                            {gettext('All notifications are set to be paused from {{pauseFrom}} to {{pauseTo}}', {pauseFrom: formatDate(this.props.fullUser.notification_schedule.pauseFrom), pauseTo: formatDate(this.props.fullUser.notification_schedule.pauseTo)})}
```
tomaskikutis: improve naming of placeholders, use date1, date2 or dateFrom, dateTo. it's important that translators understand that placeholder is a date.

assets/notifications/components/NotificationsApp.tsx
```diff
@@ -54,6 +66,11 @@ const mapDispatchToProps = (dispatch: any) => ({
     clearNotification: (id: any) => dispatch(deleteNotification(id)),
     clearAll: () => dispatch(deleteAllNotifications()),
     loadNotifications: () => dispatch(loadNotifications()),
+    resumeNotifications: (userId: string) => {
```
tomaskikutis: improve naming.

assets/notifications/components/NotificationsApp.tsx
```diff
@@ -26,13 +32,19 @@ class NotificationsApp extends React.Component<any, any> {
...
+                    );
+                }}
             />,
         ];
     }
 }
 
 NotificationsApp.propTypes = {
-    user: PropTypes.string,
+    fullUser: PropTypes.object,
```
tomaskikutis: convert to interface

assets/search/components/TopicEditor.tsx
```diff
@@ -596,6 +597,7 @@ const mapDispatchToProps = (dispatch: any) => ({
         dispatch(loadMyWireTopic(topic._id)),
     setTopicEditorFullscreen: (fullscreen: boolean) => dispatch(setTopicEditorFullscreen(fullscreen)),
     openEditTopicNotificationsModal: () => dispatch(openEditTopicNotificationsModal()),
+    openPauseNotificationModal: () => dispatch(openPauseNotificationModal()),
```
tomaskikutis: Is this used anywhere?

assets/user-profile/components/EditNotificationScheduleModal.tsx
```diff
@@ -190,8 +190,8 @@ class EditNotificationScheduleModalComponent extends React.Component<IProps, ISt
 const mapDispatchToProps = (dispatch: any) => ({
     modalFormInvalid: () => dispatch(modalFormInvalid()),
     modalFormValid: () => dispatch(modalFormValid()),
-    updateUserNotificationSchedules: (schedule: Omit<IUser['notification_schedule'], 'last_run_time'>) => (
-        dispatch(updateUserNotificationSchedules(schedule))
+    updateUserNotificationSchedules: (schedule: Omit<IUser['notification_schedule'], 'last_run_time'>, message: string) => (
```
tomaskikutis: Naming should be in singular, review all names in your PR. why are you accepting an argument for a message if there's only one usage here?

assets/user-profile/components/profile/UserProfile.tsx
```diff
@@ -180,6 +184,45 @@ class UserProfile extends React.PureComponent<IProps> {
...
+                                        </div>
+                                    )
+                                    : (
+                                        <div className="nh-container nh-container--highlight mb-3 mt-1">
+                                            <button
+                                                type="button"
+                                                className="nh-button nh-button--small nh-button--tertiary"
+                                                onClick={this.props.openPauseNotificationModal}
+                                            >
+                                                {gettext('Pause All Notification')}
```
tomaskikutis: should be `Pause all notifications`.

assets/user-profile/components/PauseNotificationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+                        ref={this.formRef}
+                        onSubmit={(event) => this.onSubmitForm(event)}
+                    >
+                        <div className='d-flex align-items-center gap-5'>
+                            <div className="d-flex flex-column align-items-start">
+                                <label htmlFor={DATE_ID}>{gettext('From')}</label>
+                                <input
+                                    id={DATE_ID}
+                                    type="date"
+                                    name={DATE_ID}
```
tomaskikutis: what's the reason you're setting name field?

assets/components/NotificationList.tsx
```diff
@@ -70,9 +85,11 @@ class NotificationList extends React.Component<any, any> {
     }
 
     render() {
+        const notificationArePaused: boolean = new Date(this.props.fullUser.notification_schedule?.pauseFrom ?? '') < new Date();
```
tomaskikutis: You should't be creating a date from an empty string. If it's not defined, simply set `notificationArePaused` to `false` and compare dates only if it's defined.

tomaskikutis (commented):
also check one other comment I left unresolved

assets/user-profile/components/PauseNotificationModal.tsx
tomaskikutis: use `schedule: Partial<IUser['notification_schedule']>` for schedule type. It will solve typing issues.

tomaskikutis (approved):

PR #915, opened 2024-05-27
fix agenda list loading

petrjasek:
only display days with some event starting
so the newly loaded events are added at the bottom. plus make agenda page size configurable

CPCN-818

## Checklist

### DIFF
--- a/assets/agenda/components/AgendaList.tsx
+++ b/assets/agenda/components/AgendaList.tsx
@@ -428,7 +428,7 @@ class AgendaList extends React.Component<IProps, IState> {
                 }}
                 onScroll={this.props.onScroll}
             >
-                {this.props.groupedItems.map((group) => (
+                {this.props.groupedItems.filter((group) => group.items.length > 0).map((group) => (
                     <React.Fragment key={group.date}>
                         <div className='wire-articles__header' key={`${group.date}header`}>
                             {this.getListGroupDate(group)}
--- a/assets/agenda/components/AgendaList.tsx
+++ b/assets/agenda/components/AgendaList.tsx
@@ -414,6 +414,7 @@ class AgendaList extends React.Component<IProps, IState> {
     }
 
     render() {
+        const lastGroupWithItems = this.props.groupedItems.findLastIndex((group) => group.items.length > 0);
         return (
             <div
                 className={classNames('wire-articles wire-articles--list', {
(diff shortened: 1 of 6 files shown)


assets/agenda/components/AgendaList.tsx
```diff
@@ -428,7 +428,7 @@ class AgendaList extends React.Component<IProps, IState> {
                 }}
                 onScroll={this.props.onScroll}
             >
-                {this.props.groupedItems.map((group) => (
+                {this.props.groupedItems.filter((group) => group.items.length > 0).map((group) => (
```
MarkLark86: I think this might cause an issue to not show the `hiddenItems` for dates where no other `items` are shown.

Example:
```
Date 1:
items: 1
hiddenItems: 1

Date 2:
items: 0
hiddenItems: 1

Date 3:
items: 1
hiddenItems: 1

Date 4:
items: 0
hiddenItems: 1
```

With this change `Date 2` would not be shown, which I believe may cause confusion.

What we could do instead is use the date of the last group that has items to be shown (disregarding hidden items) to determine which is the last date in the list that should be shown. As the list of items received from the server are paginated and sorted by date, this might suffice. What do you think?

In the above example, the last shown date would be for `Date 3` (as that is the last group with an item count), which would show `Date 2` but not `Date 4`.

assets/agenda/components/AgendaList.tsx
```diff
@@ -414,6 +414,7 @@ class AgendaList extends React.Component<IProps, IState> {
     }
 
     render() {
+        const lastGroupWithItems = this.props.groupedItems.findLastIndex((group) => group.items.length > 0);
```
MarkLark86, line 417: We have the [groupItems](https://github.com/superdesk/newsroom-core/blob/develop/assets/agenda/utils.ts#L699) function in agenda/utils.ts that provides the logic around grouping.

Moving this filtering code to that function would help by having a single source of truth, if the group is in the store then it should be rendered.

Or is there another reason to filter groups in the view separately to the store?

MarkLark86 (approved):
In any case, the outcome of your changes are agreeable.

assets/agenda/components/AgendaList.tsx
petrjasek: I'm not sure if it should be there, the items still belong to that group we just don't want to see the group yet, so imo it's rather for the rendering

PR #916, opened 2024-05-28
update logic based on all_day flag [CPCN-815]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/utils.py
+++ b/newsroom/utils.py
@@ -199,7 +199,11 @@ def get_agenda_dates(agenda: Dict[str, Any], date_paranthesis: bool = False) -> 
     all_day = agenda.get("dates", {}).get("all_day", False)
 
     if all_day:
-        return "{}".format(date_short(start))
+        return (
+            "{}".format(date_short(start))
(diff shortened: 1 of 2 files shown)


newsroom/utils.py
```diff
@@ -196,9 +196,9 @@ def date_short(datetime):
 def get_agenda_dates(agenda: Dict[str, Any], date_paranthesis: bool = False) -> str:
     start = parse_date_str(agenda.get("dates", {}).get("start"))
     end = parse_date_str(agenda.get("dates", {}).get("end"))
+    all_day = agenda.get("dates", {}).get("all_day", False)
 
-    if start.time() == datetime.min.time() or end.time() == datetime.min.time():
-        # if start time or end time is not present
+    if all_day:
```
petrjasek, line 201: all day can be set also for multi day events

devketanpro: so in this case, we have to show both dates `start and end` right ?

petrjasek: exactly

petrjasek (approved):

PR #917, opened 2024-05-29
Fix handling of all-day events to display exact dates without timezone adjustments [CPCN-815]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/utils.py
+++ b/newsroom/utils.py
@@ -200,9 +200,9 @@ def get_agenda_dates(agenda: Dict[str, Any], date_paranthesis: bool = False) -> 
 
     if all_day:
         return (
-            "{}".format(date_short(start))
+            "{}".format(start.strftime("%d/%m/%Y"))
             if start.date() == end.date()
-            else "{} - {}".format(date_short(start), date_short(end))
+            else "{} - {}".format(start.strftime("%d/%m/%Y"), end.strftime("%d/%m/%Y"))
         )
 
     if start + timedelta(minutes=DAY_IN_MINUTES) < end:
--- a/newsroom/utils.py
+++ b/newsroom/utils.py
@@ -18,7 +18,7 @@ from bson.errors import InvalidId
 from eve.utils import config, ParsedRequest
 from eve_elastic.elastic import parse_date, ElasticCursor
 from flask import current_app as app, json, abort, request, g, flash, session, url_for
-from flask_babel import gettext, _get_current_context
+from flask_babel import gettext
 
 from newsroom.types import PublicUserData, User, Company, Group, Permissions
 from newsroom.template_filters import (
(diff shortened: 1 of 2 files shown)


newsroom/utils.py
```diff
@@ -200,9 +200,9 @@ def get_agenda_dates(agenda: Dict[str, Any], date_paranthesis: bool = False) ->
 
     if all_day:
         return (
-            "{}".format(date_short(start))
+            "{}".format(start.strftime("%d/%m/%Y"))
```
petrjasek: we should use config for the format

petrjasek: I mean an existing one https://github.com/superdesk/newsroom-core/blob/develop/newsroom/web/default_settings.py#L315

devketanpro: yep, however this configuration is flask-babel specific, and we removed the use of babel from here because we were having timezone issues with it. Also, `strftime` does not accept this format.

petrjasek: you can use `flask_babel.format_date` here just make sure to set `rebase=False` there:
https://python-babel.github.io/flask-babel/#flask_babel.format_date

devketanpro: Thanks

newsroom/utils.py
```diff
@@ -199,12 +200,10 @@ def get_agenda_dates(agenda: Dict[str, Any], date_paranthesis: bool = False) ->
...
-        format = app.config.get("ALL_DAY_DATE_FORMAT", "%d/%m/%Y")
-        return (
-            "{}".format(start.strftime(format))
-            if start.date() == end.date()
-            else "{} - {}".format(start.strftime(format), end.strftime(format))
-        )
+        format = get_client_format("NOTIFICATION_EMAIL_DATE_FORMAT") or "dd/MM/yyyy"
+        _start = format_datetime(start, format, rebase=False)
+        _end = format_datetime(end, format, rebase=False)
+        return "{}".format(_start) if start.date() == end.date() else "{} - {}".format(_start, _end)
```
petrjasek: seems like `return _start if start.date() == ...` would do

newsroom/utils.py
```diff
@@ -199,12 +200,10 @@ def get_agenda_dates(agenda: Dict[str, Any], date_paranthesis: bool = False) ->
...
     if all_day:
-        format = app.config.get("ALL_DAY_DATE_FORMAT", "%d/%m/%Y")
-        return (
-            "{}".format(start.strftime(format))
-            if start.date() == end.date()
-            else "{} - {}".format(start.strftime(format), end.strftime(format))
-        )
+        format = get_client_format("NOTIFICATION_EMAIL_DATE_FORMAT") or "dd/MM/yyyy"
+        _start = format_datetime(start, format, rebase=False)
+        _end = format_datetime(end, format, rebase=False)
```
petrjasek: might be better to use `format_date` here so it's more obvious what is going on

petrjasek (approved):

PR #920, opened 2024-05-30
As a user viewing search results in Large List View or Compact List View in the Wire section, I want to see a tag/label indicating if an item is an ADVISORY or an ALERT or a PRESS RELEASE or is the LATEST on a major story.

dzonidoo:
CPCN-20

## Checklist

### DIFF
--- a/assets/wire/components/WireLabel.tsx
+++ b/assets/wire/components/WireLabel.tsx
@@ -2,28 +2,33 @@ import React from 'react';
 import classNames from  'classnames';
 import {IArticle} from 'interfaces';
 import {wireLabel} from 'agenda/utils';
+import {gettext} from 'utils';
 
 interface IProps {
     item: IArticle;
 }
 
 export default function WireLabel({item}: IProps) {
-    const classes = classNames('label label--fill label--rounded');
-
     const subjectList = item.subject || [];
     const wireItemSubjects = subjectList.filter(wireLabel);
 
     return (
         wireItemSubjects.length > 0 ? (
             <div>
-                {wireItemSubjects.map(subject => (
-                    <span
-                        className={classes + ` label-wire--${subject.code}`}
-                        key={subject.name}
-                    >
-                        {subject.name}
-                    </span>
-                ))}
+                {wireItemSubjects.map(subject => {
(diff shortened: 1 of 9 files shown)


assets/wire/components/WireLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    const subjectList = item.subject || [];
+    const wireItemSubjects = subjectList.filter(wireLabel);
+
+    return (
+        wireItemSubjects.length > 0 ? (
+            <div>
+                {wireItemSubjects.map(subject => (
+                    <span
+                        className={classes + ` label-wire--${subject.code}`}
```
tomaskikutis: do not use dynamic classnames

assets/wire/components/WireLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+                        key={subject.name}
```
tomaskikutis: it's safer to use an ID as a key. In this context, `qcode` is an ID

assets/wire/components/WireLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+                    >
+                        {subject.name}
```
petrjasek: we will need to translate those unlike the other subjects, so pls use `{gettext(subject.name)}` here (the translations will be populated based on backend code)

tomaskikutis: how would `gettext` help in translating strings that sit in the DB?

petrjasek: those strings will be populated on incoming items via some backend code, so we will have them available for extraction and then later for usage

tomaskikutis: > we will have them available for extraction and then later for usage

would marking for extraction happen on the back-end? Maybe we could return the translated ones from back-end too?

petrjasek: yep it would, but to return it translated based on user language would require some processing when it's fetched and that would be quite complicated

tomaskikutis (approved):

petrjasek (approved):

PR #921, opened 2024-05-31
Introduce a time based switch for the archive search [NHUB-508]

devketanpro:
## Checklist


newsroom/wire/search.py
```diff
@@ -56,20 +66,27 @@ class WireSearchResource(newsroom.Resource):
 def versioncreated_range(created):
     _range = {}
     offset = int(created.get("timezone_offset", "0"))
-    if created.get("created_from"):
+    if created.get("date_from"):
```
MarkLark86: Keep in mind that if you change the variable name for a filter, then the `My Topic` & `Company Topic` resources will need to be updated, to reflect this change in variable name (so a migration script will be required).

Otherwise these saved topics will not apply the date filters

devketanpro: got it !

assets/utils.tsx
```diff
@@ -703,6 +703,11 @@ export function getCreatedSearchParamLabel(created: any) {
         } else {
             return {from: formatDate(created.from)};
         }
+    } else if (created.date_filter){
+        const value = created.date_filter.split('_').map(
```
petrjasek: that should be rather a lookup in the config for matching value and using the label from there

assets/wire/components/filters/FiltersTab.tsx
```diff
@@ -91,7 +91,15 @@ class FiltersTab extends React.Component<any, any> {
             )
         );
 
-        this.setState({createdFilter: created});
+        this.setState({createdFilter: created}, () => {
+            const defaultFilterExists = this.props.dateFilters?.some((filter: {
```
petrjasek: is this checking if there is a default filter at all or if the currently selected filter is the default one? would improve the naming of the constant to make this more obvious

assets/wire/components/filters/NavCreatedPicker.tsx
```diff
@@ -1,55 +1,102 @@
...
-        setCreatedFilter({from: createdFilter.from === value ? null : value, to: null});
-    };
+function NavCreatedPicker({setCreatedFilter, createdFilter, context, wireFilters}: any) {
+    const [showCustomRange, setShowCustomRange] = useState(false);
+
+    useEffect(() => {
+        if (createdFilter.date_filter === 'custom_date') {
+            setShowCustomRange(true);
+        } else {
+            setShowCustomRange(false);
```
petrjasek: this block could be just `setShowCustomRange(createdFilter.date_filter === 'custom_date')` but imo this should be done already in the `useState` call so it's not triggering rerender right away

newsroom/wire/search.py
```diff
@@ -290,6 +318,34 @@ def apply_request_filter(self, search, highlights=True):
...
+            if date_filter == "custom_date":
+                pass
+            elif date_filter == "last_week":
+                start_of_last_week, end_of_last_week = get_start_and_end_of_last_week()
+                search.query["bool"]["must"].append(
+                    {
+                        "range": {
+                            "versioncreated": {
+                                "gte": start_of_last_week.isoformat(),
+                                "lt": end_of_last_week.isoformat(),
```
petrjasek: I think you can just use here:
```py
{"gte": "now-1w/w", "lt": "now/w", "time_zone": app.config.get("DEFAULT_TIMEZONE")}
```

newsroom/wire/search.py
```diff
@@ -290,6 +318,34 @@ def apply_request_filter(self, search, highlights=True):
...
+                            }
+                        }
+                    }
+                )
+            else:
+                query = self.get_date_filter_query(date_filter)
+                if query:
+                    search.query["bool"]["must"].append({"range": {"versioncreated": {"gte": query, "lt": "now+1d/d"}}})
```
petrjasek: I don't think it needs any `lt` part

newsroom/wire/search.py
```diff
@@ -290,6 +318,34 @@ def apply_request_filter(self, search, highlights=True):
...
+        else:
+            default_time_filter = next((f for f in self.get_time_filters() if f["default"]), None)
+            if default_time_filter:
```
petrjasek, line 344: what if the default filter is `last_week` ?

newsroom/wire/search.py
```diff
@@ -25,6 +27,13 @@ def get_aggregations():
     return app.config.get("WIRE_AGGS", {})
 
 
+class TimeFilter(TypedDict):
+    name: str
+    default: bool
+    query: str
```
petrjasek: might be good idea if the query is set as:
```py
class DateRangeQuery(TypedDict, total=False):
    gt: str
    gte: str
    lt: str
    lte: str
    time_zone: str
```
and then we could just apply whatever is set there

assets/utils.tsx
```diff
@@ -704,10 +704,8 @@ export function getCreatedSearchParamLabel(created: any) {
             return {from: formatDate(created.from)};
         }
     } else if (created.date_filter){
-        const value = created.date_filter.split('_').map(
-            (word: string) => word.charAt(0)
-                .toUpperCase() + word.slice(1).toLowerCase()).join(' ');
-        return {relative: gettext(value)};
+        const value = dateFilters?.find((filter: any) => filter?.filter == created?.date_filter);
+        return {relative: gettext(value?.name)};
```
petrjasek: we shouldn't use gettext here, client should get it already translated

newsroom/wire/search.py
```diff
@@ -319,32 +327,33 @@ def apply_request_filter(self, search, highlights=True):
...
+                    date_range_query = {"gte": query, "time_zone": app.config.get("DEFAULT_TIMEZONE")}
         else:
-            default_time_filter = next((f for f in self.get_time_filters() if f["default"]), None)
+            default_time_filter: TimeFilter = next((f for f in self.get_time_filters() if f["default"]), None)
             if default_time_filter:
-                default_query = default_time_filter["query"]
-                search.query["bool"]["must"].append(
-                    {"range": {"versioncreated": {"gte": default_query, "lt": "now+1d"}}}
-                )
+                if default_time_filter["filter"] == "last_week":
```
petrjasek: I see this code duplicated, that should be avoided.

newsroom/wire/search.py
```diff
@@ -319,32 +327,33 @@ def apply_request_filter(self, search, highlights=True):
...
+                    date_range_query: DateRangeQuery = {
+                        "gte": "now-1w/w",
+                        "lt": "now/w",
+                        "time_zone": app.config.get("DEFAULT_TIMEZONE"),
+                    }
+                else:
+                    default_query = default_time_filter["query"]
+                    date_range_query = {"gte": default_query, "time_zone": app.config.get("DEFAULT_TIMEZONE")}
```
petrjasek: what I meant was that `default_query` from settings would be `DateRangeQuery` instance so then it could be used as is inside range query:
```py
search.query["bool"]["must"].append({"range": {"versioncreated": time_filter["query"]}})
```

newsroom/web/default_settings.py
```diff
@@ -806,8 +806,22 @@
...
+#:
+#: .. versionadded: 2.8
+#:
 WIRE_TIME_FILTERS = [
-    {"name": "Today", "filter": "today", "default": False, "query": "now/d"},
-    {"name": "Last Week", "filter": "last_week", "default": False, "query": "now-1w/w"},
-    {"name": "Last 30 days", "filter": "last_30_days", "default": True, "query": "now-30d/d"},
+    {"name": "Today", "filter": "today", "default": False, "query": {"gte": "now/d"}},
+    {
+        "name": "Last Week",
```
petrjasek: here it should use the `lazy_gettext` to get those translated

assets/interfaces/common.ts
```diff
@@ -80,3 +80,13 @@ export interface IOccurStatus {
     qcode: string;
     label: string;
 }
+
+export interface IDateFilters{
+    name: string,
+    default: boolean,
+    query?: Array<{}>
```
petrjasek: not sure about this type

assets/interfaces/common.ts
```diff
@@ -80,3 +80,13 @@ export interface IOccurStatus {
...
+    filter: Array<['today', 'last_week', 'last_30_days', '']>
```
petrjasek: and here the filter names could be different if there will be some client specific ones defined, that should be rather a string

newsroom/wire/search.py
```diff
@@ -25,6 +27,21 @@ def get_aggregations():
...
+    lt: str
+    lte: str
+    time_zone: Optional[str]
+
+
+class TimeFilter(TypedDict):
+    name: str
+    default: bool
+    query: DateRangeQuery
+    filter: Literal["today", "last_week", "last_30_days", ""]
```
petrjasek: those are set via config, so this can be any string in general

newsroom/wire/search.py
```diff
@@ -271,6 +295,28 @@ def has_permissions(self, item, ignore_latest=False):
...
+        """Retrieve the time filters from app config."""
+        return app.config.get("WIRE_TIME_FILTERS", [])
+
+    def get_date_filter_query(self, filter_name: str) -> Optional[DateRangeQuery]:
+        """Get the query for the given filter name."""
+        time_filters = self.get_time_filters()
+        for time_filter in time_filters:
+            if time_filter["filter"] == filter_name:
+                query = time_filter["query"]
+                query["time_zone"] = app.config.get("DEFAULT_TIMEZONE")
```
petrjasek: it should use probably `query.setdefault("time_zone", app.config.get("DEFAULT_TIMEZONE"))` if it's possible to specify it also via config

newsroom/wire/search.py
```diff
@@ -271,6 +295,28 @@ def has_permissions(self, item, ignore_latest=False):
...
+                return query
+        return None
+
+    def get_date_range_query(self, date_filter: str) -> Optional[DateRangeQuery]:
+        if date_filter != "custom_date":
+            query = self.get_date_filter_query(date_filter)
+            if query:
+                query["time_zone"] = app.config.get("DEFAULT_TIMEZONE")
```
petrjasek: seems like this is already done in `get_date_filter_query`

assets/interfaces/common.ts
```diff
@@ -84,8 +84,8 @@ export interface IOccurStatus {
 export interface IDateFilters{
     name: string,
     default: boolean,
-    query?: Array<{}>
-    filter: Array<['today', 'last_week', 'last_30_days', '']>
+    query?: string,
```
petrjasek: I think it's not used on the client right? so you might as well remove it

devketanpro: yep

petrjasek (approved):

PR #929, opened 2024-06-03
implement backend for pause notifications

petrjasek:
CPCN-833 CPCN-186

## Checklist

### DIFF
--- a/newsroom/users/users.py
+++ b/newsroom/users/users.py
@@ -1,3 +1,4 @@
+import pytz
 import bcrypt
 import newsroom
 import superdesk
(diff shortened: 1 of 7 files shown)


newsroom/users/users.py
```diff
@@ -405,6 +406,17 @@ class UserApprovalUpdates(TypedDict, total=False):
             updated_user.update(user_updates)
             send_token(updated_user, token_type="new_account", update_token=False)
 
+    def user_has_paused_notifications(self, user) -> bool:
```
MarkLark86: Type for `user` missing.

MarkLark86 (approved):

PR #930, opened 2024-06-04
[CPCN-178] Implement a new report for Company Admins

eos87:
### Purpose
Implement a new report for Company Admins. The report shows the number of saved My topics and Company topics per user in their company.

### What has changed
- Added the new report
- Completed some typings in the frontend code
- Small adjustment to `__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` usage so store name will be recognised in the browser terminal
- Adjusted template filters according to user type 

### How to test
- Checkout the branch and run the project
- Log in as a Company Admin user and observe that the user is now able to see the reports section 
- Enter the reports section and select the only report in the dropdown an run it
- Please double check if the calculations are correct

### Checklist

### Screenshots
This is how it looks
image

~**Note:** I will add a few tests cases tomorrow.~

Thanks for checking!

Resolves: [CPCN-178]

[CPCN-178]: https://sofab.atlassian.net/browse/CPCN-178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/company-reports/actions.ts
+++ b/assets/company-reports/actions.ts
@@ -14,6 +14,7 @@ export const REPORTS_NAMES = {
     'COMPANY_NEWS_API_USAGE': 'company-news-api-usage',
     'PRODUCT_COMPANIES': 'product-companies',
     'EXPIRED_COMPANIES': 'expired-companies',
+    'COMPANY_AND_USER_SAVED_SEARCHES': 'company-and-user-saved-searches',
 };
 
 
@@ -28,6 +29,7 @@ export const REPORTS: any = {
     [REPORTS_NAMES.COMPANY_NEWS_API_USAGE]: '/reports/company-news-api-usage',
     [REPORTS_NAMES.PRODUCT_COMPANIES]: '/reports/product-companies',
     [REPORTS_NAMES.EXPIRED_COMPANIES]: '/reports/expired-companies',
+    [REPORTS_NAMES.COMPANY_AND_USER_SAVED_SEARCHES]: '/reports/company-and-user-saved-searches',
 };
 
 function getReportQueryString(currentState: any, next: any, exportReport: any, notify: any) {
--- a/assets/company-reports/actions.ts
+++ b/assets/company-reports/actions.ts
@@ -1,6 +1,7 @@
 import {errorHandler, getItemFromArray, getDateInputDate, notify, gettext} from '../utils';
 import server from '../server';
 import {get, cloneDeep} from 'lodash';
+import {type ICompanyReportsData} from './types';
 
 export const REPORTS_NAMES = {
     'COMPANY_SAVED_SEARCHES': 'company-saved-searches',
@@ -70,7 +71,7 @@ function getReportQueryString(currentState: any, next: any, exportReport: any, n
 }
 
 export const INIT_DATA = 'INIT_DATA';
-export function initData(data: any): any {
+export function initData(data: ICompanyReportsData): any {
     return function (dispatch: any) {
         dispatch(fetchProducts());
         dispatch({type: INIT_DATA, data});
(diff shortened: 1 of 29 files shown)


assets/company-reports/components/SavedMyTopicsAndCompanyTopics.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+interface IProps {
+    results: Array<IResult>;
+    print: boolean;
+}
+
+
+const SavedMyTopicsAndCompanyTopics = ({results, print}: IProps) => {
+    const list = results && results.map((item: any) =>
```
petrjasek: I think the type here should be `IResult` rather than `any` but it should be able to figure it out without the type hint

assets/utils.tsx
```diff
@@ -19,6 +19,7 @@ import {IArticle, IUser} from 'interfaces';
 // TODO: Improve how we load Moment locales, based on server config
 import 'moment/locale/fr-ca';
 import 'moment/locale/fi';
+import {type IClientConfig} from 'globals';
```
petrjasek: why is this imported?

assets/company-reports/actions.ts
```diff
@@ -1,6 +1,7 @@
 import {errorHandler, getItemFromArray, getDateInputDate, notify, gettext} from '../utils';
 import server from '../server';
 import {get, cloneDeep} from 'lodash';
+import {type ICompanyReportsData} from './types';
```
petrjasek, line 4: we usually don't use import type specifically, is there any benefit?

assets/globals.d.ts
```diff
@@ -62,52 +64,53 @@ interface IClientConfig {
...
+        translations: any;
+        profileData: any;
+        wireData: any;
+        homeData: any;
+        agendaData: any;
+        amNewsData: any;
+        viewData: any;
+        report: any;
+        factCheckData: any;
+        companyReportsData: ICompanyReportsData;
```
petrjasek: I think you can just declare this part of the data in the `company-reports/index` and keep this file unmodified, it should merge it with this declaration:
```ts
declare global {
    interface Window {
        companyReportsData: ...
    }
}
```

assets/company-reports/components/SavedMyTopicsAndCompanyTopics.tsx
eos87: yes, good catch, I forgot to update it :)

assets/globals.d.ts
eos87: and keep it with `any`?

assets/utils.tsx
eos87: because when I added this `companyReportsData: ICompanyReportsData;` to the `globals.d.ts` file, for some reason TS started complaining with some weird generic errors and I was only able to fix them by using specific imports and changing the declaration of the global Window object.

assets/globals.d.ts
eos87: I will give it a try.

petrjasek: no you can import there `ICompanyReportsData`

assets/company-reports/actions.ts
eos87: I don't think there isn't much benefit. Personally I prefer explicit than implicit but in this case it was because of the complain from TS I explained here https://github.com/superdesk/newsroom-core/pull/930#discussion_r1627162388.

assets/globals.d.ts
eos87: yeah, you're right. I misunderstood.

assets/utils.tsx
eos87: On a second thought, perhaps the benefit is that it provides more clarity. Also, there could be times when you would have the same name for a type more than once across the codebase, so an explicit import would be better than an implicit in those cases. What do you think?

assets/company-reports/actions.ts
eos87: After applying the suggested changes here https://github.com/superdesk/newsroom-core/pull/930#discussion_r1627132852 the import is still required, otherwise TS will complain about undefined type.

petrjasek (approved):

PR #948, opened 2024-06-11
[CPCN-814] Send `user id` to GA and add column to user management table

eos87:
### Purpose
Send `User Id` to Google Analytics to it can be used in `Number of logged-in sessions` report. Plus replace `Mobile` column with `User Id` in User Management table. 

### What has changed
- Adjusted `gtag` script usage so user id is sent to Google Analytics
- Added new column `User Id` to user management table
- Completed some types

### To test
- Checkout branch and run the project
- Navigate the app. Verify in the browser terminal that user id is sent in collect request from google analytics (see screenshot below)
- Optional: Verify in Google Analytics dashboard that user id is being tracked

### Checklist

### Screenshots
Google Analytics collect request:
image

User's management table:
image

Some references about the implementation from Google Analytics Documentation https://developers.google.com/analytics/devguides/collection/ga4/user-id?client_type=gtag#send_user_ids

Thanks for reviewing!

Resolves: [CPCN-814]

[CPCN-814]: https://sofab.atlassian.net/browse/CPCN-814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/templates/scripts.html
+++ b/newsroom/templates/scripts.html
@@ -46,17 +46,8 @@
 
     let configParams = {};
     {% if session.get('user') %}
-        gtag('set', {'user_id': '{{ hash(session['user']) | tojson }}' });
-
         configParams['user_id'] = '{{ session['user'] }}';
     {% endif %}
-
-    if ('{{ get_setting('google_analytics') }}'.startsWith('UA')) {
-        configParams['custom_map'] = {
-            dimension1: 'company',
-            dimension2: 'user',
-        };
-    }
 
     gtag('config', '{{ get_setting('google_analytics') | tojson }}', configParams);
 
--- a/newsroom/templates/scripts.html
+++ b/newsroom/templates/scripts.html
@@ -35,35 +35,36 @@
 <!-- Global site tag (gtag.js) - Google Analytics -->
 <script async src="https://www.googletagmanager.com/gtag/js?id={{ get_setting('google_analytics') | safe }}"></script>
 <script>
-  window.dataLayer = window.dataLayer || [];
-  function gtag(){dataLayer.push(arguments);}
-  gtag('js', new Date());
+    window.dataLayer = window.dataLayer || [];
+    function gtag(){dataLayer.push(arguments);}
+    gtag('js', new Date());
 
-  gtag('set', 'user_properties', {
-      company: (window.profileData || {}).companyName || 'none',
-      user: ((window.profileData || {}).user || {}).first_name || 'unknown',
-  });
+    gtag('set', 'user_properties', {
+        company: (window.profileData || {}).companyName || 'none',
+        user: ((window.profileData || {}).user || {}).first_name || 'unknown',
+    });
 
-  {% if session.get('user') %}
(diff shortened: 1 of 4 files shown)


newsroom/templates/scripts.html
```diff
@@ -35,35 +35,36 @@
...
-  }
-
-  {% with messages = get_flashed_messages(category_filter=['analytics']) %}
-    {% if messages %}
-        analytics.sendEvents({{ messages | tojson | safe }});
+        configParams['user_id'] = '{{ session['user'] }}';
     {% endif %}
-  {% endwith %}
+
+    if ('{{ get_setting('google_analytics') }}'.startsWith('UA')) {
```
petrjasek: you can drop this UA section, that won't be supported by Google after July 1st

newsroom/templates/scripts.html
```diff
@@ -35,35 +35,36 @@
...
+        company: (window.profileData || {}).companyName || 'none',
+        user: ((window.profileData || {}).user || {}).first_name || 'unknown',
+    });
 
-  {% if session.get('user') %}
-  gtag('set', {'user_id': '{{ hash(session['user']) | tojson }}' });
-  {% endif %}
+    let configParams = {};
+    {% if session.get('user') %}
+        gtag('set', {'user_id': '{{ hash(session['user']) | tojson }}' });
```
petrjasek: I think you have to remove that `hash` here otherwise it will send hash to analytics but that is not what you will see in the UI

eos87: I left it with the `hash` because it was already there. The one I added is [this one actually](https://github.com/superdesk/newsroom-core/pull/948/files#diff-2dfa5523a06deb5a874417e434f6bf221d060a1314ce91dc54c3a03a5b4e1837R51) and the one with the hash was [this one](https://github.com/superdesk/newsroom-core/pull/948/files#diff-2dfa5523a06deb5a874417e434f6bf221d060a1314ce91dc54c3a03a5b4e1837L47-L49) I just relocated it 

Do you think we should remove the hash? Wouldn't it break anything?

petrjasek: you can just set it here without hash and remove the other one, according to docs this is the way to do it
https://developers.google.com/analytics/devguides/collection/ga4/user-id?client_type=gtag#send_user_ids

eos87: yeah, but my point was that I don't know why the `gtag('set', {'user_id'` was there in the first place with a hashed id. But if you think it is safe to remove it, I'll do it right away.

newsroom/templates/scripts.html
eos87: done

newsroom/templates/scripts.html
eos87: done

petrjasek: it was there to obfuscate the real id but provide consistent value for a particular user, but now we want the actual ids

petrjasek (approved):

PR #949, opened 2024-06-12
update agenda datetime filters [CPCN-48]

devketanpro:
## Checklist

### DIFF
--- a/assets/utils.tsx
+++ b/assets/utils.tsx
@@ -700,6 +700,12 @@ export function getCreatedSearchParamLabel(created: any, dateFilters: any = null
             return {relative: gettext('This week')};
         } else if (created.from === 'now/M') {
             return {relative: gettext('This month')};
+        } else if (created.from === 'now-24h/h') {
+            return {relative: gettext('Last 24 hours')};
+        } else if (created.from === 'now-7d/d') {
+            return {relative: gettext('Last 7 days')};
+        } else if (created.from === 'now-30d/d') {
+            return {relative: gettext('Last 30 days')};
         } else {
             return {from: formatDate(created.from)};
         }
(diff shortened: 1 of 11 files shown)


assets/utils.tsx
```diff
@@ -700,6 +700,12 @@ export function getCreatedSearchParamLabel(created: any, dateFilters: any = null
             return {relative: gettext('This week')};
         } else if (created.from === 'now/M') {
             return {relative: gettext('This month')};
+        } else if (created.from === 'now-24h/h') {
+            return {relative: gettext('Last 24 hours')};
+        } else if (created.from === 'now-7d/d') {
+            return {relative: gettext('Last 7 days')};
+        } else if (created.from === 'now-30d/d') {
+            return {relative: gettext('Last 30 days')};
```
petrjasek: I wonder if we need those if the options will be configured with label already, maybe we do for backwards compatibility?

newsroom/utils.py
```diff
@@ -504,15 +504,21 @@ def today(time, offset):
...
         return today(time, offset)
     if date == "now/w":
-        _today = today(time, offset)
         monday = _today - timedelta(days=_today.weekday())
         return monday
     if date == "now/M":
         month = today(time, offset).replace(day=1)
         return month
+    if date == "now-24h/h":
+        return _today - timedelta(hours=24)
```
petrjasek, line 517: I think here it should also set `.replace(minutes=0, seconds=0)` but not sure how it's actually used

petrjasek (approved):

newsroom/utils.py
devketanpro: you're right about using `.replace(minute=0, second=0)` to ensure the datetime starts at the hour. However, since `_today` is already at `"00:00:00"`, it's unnecessary after subtracting `24` hours because the result will naturally be aligned to the hour.

PR #950, opened 2024-06-13
Implement case insensitive search

thecalcc:
CPCN-827
## Checklist


tomaskikutis (commented):
Filter before sorting so sorting function doesn't have to go through items that will be filtered out anyway.

tomaskikutis (approved):

PR #973, opened 2024-06-24
re-add loading indicator to agenda/wire

petrjasek:
NHUB-521

## Checklist


petrjasek:
hi @fritzSF can you check those classnames if that's still valid or we need to update it somehow

MarkLark86 (approved):

PR #975, opened 2024-06-25
Fix : Coverage and Event status display in English instead of French in the French email template [CPCN-838]

devketanpro:
## Checklist

### DIFF
--- a/newsroom/email.py
+++ b/newsroom/email.py
@@ -7,7 +7,7 @@ from typing_extensions import TypedDict
 
 from superdesk import get_resource_service
 from flask import current_app, render_template, url_for
-from flask_babel import gettext
+from flask_babel import gettext, force_locale
 from flask_mail import Attachment, Message
 from jinja2 import TemplateNotFound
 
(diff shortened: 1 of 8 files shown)


newsroom/email.py
```diff
@@ -266,7 +266,7 @@ def _send_localized_email(
     email_templates = get_resource_service("email_templates")
     html_template = get_language_template_name(template, language, "html")
     text_template = get_language_template_name(template, language, "txt")
-    with template_locale(language, timezone):
+    with template_locale(language, timezone), force_locale(language):
```
petrjasek: I'm wondering if it would actually make sense to call that `force_locale` inside `template_locale`, might be easier to use in general.
and pls add some test

newsroom/email.py
```diff
@@ -266,18 +266,19 @@ def _send_localized_email(
...
         template_kwargs.setdefault("recipient_language", language)
-        send_email(
-            to=to,
-            subject=subject,
-            text_body=render_template(text_template, **template_kwargs),
-            html_body=render_template(html_template, **template_kwargs),
-            sender_name=get_sender_name(language),
-            **email_kwargs,
-        )
+        with force_locale(language):
```
petrjasek: you can call this inside `template_locale` method, around that `yield` probably

devketanpro: yep

petrjasek (approved):

PR #976, opened 2024-06-26
Add a config option for top story sorting or default sorting

thecalcc:
NHUB-514
## Checklist


petrjasek:
I think the idea was to have the events with coverages on top being configurable, in theory it should be possible to use top story sorting without the coverages being on top

thecalcc:
> I think the idea was to have the events with coverages on top being configurable, in theory it should be possible to use top story sorting without the coverages being on top

Ok pushed some changes, making coverages sorting based on config and top story sorting is unconditional. In practice if we don't have top stories items would fall back to the default order, if we do top story ones will be on top.

newsroom/web/default_settings.py
```diff
@@ -396,7 +396,7 @@
             "button_label": lazy_gettext("Completed"),
         },
     },
-    "agenda_top_story_sorting": False,
+    "agenda_top_story_sorting_only": True,
```
petrjasek: would call it probably `agenda_sort_events_with_coverage_on_top` or something to be more descriptive

petrjasek (approved):

PR #977, opened 2024-06-27
Display user role in list and detail view based on new Config [CPCN-690]

devketanpro:
## Checklist


assets/tests.ts
```diff
@@ -57,6 +57,7 @@ window.newsroom = {
         display_all_versions_toggle: false,
         display_credits: false,
         display_news_only: true,
+        display_user_role: true,
```
petrjasek: we should probably call that `author_role` not `user_role` that seems like it's related to users management

petrjasek (approved):

PR #978, opened 2024-06-27
fix notifications popup not rendering

petrjasek:
NHUB-515

## Checklist

### DIFF
--- a/assets/components/NotificationList.tsx
+++ b/assets/components/NotificationList.tsx
@@ -75,13 +75,14 @@ class NotificationList extends React.Component<IProps, IState> {
     }
 
     toggleDisplay() {
-        this.setState({displayItems: !this.state.displayItems});
-        if (!this.state.displayItems) {
-            this.props.loadNotifications();
-            (document.getElementById('header-notification') as HTMLElement).classList.add('navbar-notifications--open');
-        } else {
-            (document.getElementById('header-notification') as HTMLElement).classList.remove('navbar-notifications--open');
-        }
+        this.setState({displayItems: !this.state.displayItems}, (() => {
+            if (this.state.displayItems) {
+                this.props.loadNotifications();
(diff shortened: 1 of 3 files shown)


petrjasek:
basically I've added missing `return` on:
https://github.com/superdesk/newsroom-core/pull/978/files#diff-99ecab827c4ac4bd8d141d0bc16515abdbe2068e412c56371cf13ab55693d4bdR164
and made sure via typescript it will return something

assets/components/NotificationList.tsx
```diff
@@ -112,94 +113,97 @@ class NotificationList extends React.Component<IProps, IState> {
...
-                                    <button
-                                        type="button"
-                                        className="nh-button nh-button--small nh-button--tertiary"
-                                        onClick={() => {
-                                            this.props.resumeNotifications();
-                                        }}
-                                    >
-                                        {gettext('Resume all notifications')}
-                                    </button>
+    renderPopup(notificationArePaused: boolean): Exclude<React.ReactNode, undefined> {
```
tomaskikutis: If you want to extract this from render method, can you make it a separtate component instead of adding a class method? Class methods are used if you want to access properties from the class itself(via `this`). If not, pure function is the way to go.

assets/components/NotificationList.tsx
```diff
@@ -113,98 +96,22 @@ class NotificationList extends React.Component<IProps, IState> {
                     <i className='icon--alert' onClick={this.toggleDisplay} />
                 </span>
 
-                {this.renderPopup(notificationArePaused)}
+                {this.state.displayItems === true && (
+                    <NotificationPopup
```
tomaskikutis, line 100: More often than not, when most or all props are passed to another component it means something is not right with the structure. Ideally we'd rather extract the wrapper or management of websocket subscriptions and keep notification popup code here - because it's the main thing of the component and rest are helpers.

Current change will do though, we'll improve further down the road.

tomaskikutis (approved):

PR #979, opened 2024-06-28
Hide expected wire item headline and announced version [STTNHUB-349]

devketanpro:
## Checklist


petrjasek:
we probably don't want to change defaults, but rather the config for stt only

PR #980, opened 2024-06-28
Newshub search by author name [TGA-102]

devketanpro:
## Checklist


petrjasek (dismissed):

petrjasek:
can you actually change it for tga only? we don't want to update core for it

devketanpro:
Alright, I'll change only for TGA

PR #985, opened 2024-07-01
update google analytics user properties

petrjasek:
- send user name as John D
- send user id as a custom property

CPCN-176

## Checklist


eos87 (approved):
Looks good!

PR #988, opened 2024-07-04
only show ongoing/future pause notifications info

petrjasek:
CPCN-844

## Checklist

### DIFF
--- a/assets/components/NotificationPopup.tsx
+++ b/assets/components/NotificationPopup.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import {formatDate, gettext, getTodayTimestamp, parseISODate} from 'utils';
+import {formatDate, gettext, notificationsArePaused, notificationsWillBePaused, parseISODate} from 'utils';
 import {IArticle, IUser} from 'interfaces';
 
 import NotificationListItem from './NotificationListItem';
@@ -19,11 +19,10 @@ export interface IProps {
 }
 
 export const NotificationPopup = (props: IProps) => {
-    const today = getTodayTimestamp();
     const pausedFrom = parseISODate(props.fullUser.notification_schedule?.pause_from);
     const pausedTo = parseISODate(props.fullUser.notification_schedule?.pause_to);
 
-    if (pausedFrom <= today && pausedTo >= today) {
+    if (notificationsArePaused(pausedFrom, pausedTo)) {
         return (
             <div className="notif__list dropdown-menu dropdown-menu-right show">
                 <div className='notif__list__header d-flex'>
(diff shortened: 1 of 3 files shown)


tomaskikutis (commented):
Can you instead adjust the variable `notificationsArePaused`?

Also don't use greater/less than operators on strings. That's the worst heresy I've seen in last few years :laughing: Despite all JS quirks it is still surprising it works at all.

tomaskikutis:
Also mark checklist items as done

petrjasek:
I can't adjust the variable because it should show the info also before the pause

assets/components/NotificationPopup.tsx
```diff
@@ -19,15 +19,11 @@ export interface IProps {
...
-    const notificationsArePaused = (
-        pausedFrom != '' && pausedFrom <= today &&
-        pausedTo != '' && pausedTo >= today
-    );
+    const today = getTodayTimestamp();
+    const pausedFrom = parseISODate(props.fullUser.notification_schedule?.pause_from);
+    const pausedTo = parseISODate(props.fullUser.notification_schedule?.pause_to);
 
-    if (notificationsArePaused) {
+    if (pausedFrom <= today && pausedTo >= today) {
```
tomaskikutis: I'd be easier to understand if you used variables. Maybe `notificationsArePaused` or `notificationsWillBePaused` or whatever you're checking for in that if condition.

assets/components/NotificationPopup.tsx
```diff
@@ -19,15 +19,11 @@ export interface IProps {
...
 export const NotificationPopup = (props: IProps) => {
-    const today = (new Date()).toISOString().substring(0, 10);
-    const pausedFrom = props.fullUser.notification_schedule?.pause_from ?? '';
-    const pausedTo = props.fullUser.notification_schedule?.pause_to ?? '';
-    const notificationsArePaused = (
-        pausedFrom != '' && pausedFrom <= today &&
-        pausedTo != '' && pausedTo >= today
-    );
+    const today = getTodayTimestamp();
+    const pausedFrom = parseISODate(props.fullUser.notification_schedule?.pause_from);
```
tomaskikutis, line 22: FYI you can parse by passing a string to constructor `new Date('2024-01-01')`

assets/utils.tsx
```diff
@@ -759,6 +759,18 @@ export function parseISODate(date?: string): number {
 }
 
 // Will return number of seconds since epoch for today at 00:00:00
-export function getTodayTimestamp(): number {
+function getTodayTimestamp(): number {
```
tomaskikutis, line 762: I'm thinking why are we working with UTC rather than local time? Even if date on server is stored in UTC, if we do `new Date(utcString)` or something similar and continue in local time.

tomaskikutis (approved):

assets/utils.tsx
petrjasek: it's just simpler to be able to compare those dates at 00:00,
`new Date("2024-07-10")` returns UTC 00:00 so atm 2:00 local time.
if you do `d = new Date(); d.setHours(0, 0, 0, 0);` that will make it 0:00 local time so UTC 2024-07-09T22:00:00,
but if you do `d  = newDate(); d.setUTCHours(0, 0, 0, 0);` that will be UTC 0:00 and 2:00 local time which I can compare with the stored date.

tomaskikutis: It depends if we want UI in server time or user time. I'd say we want user time in most cases. Do we pause notifications according to server time or user timezone?

petrjasek: we do it on the server based on the user timezone, but not the browser timezone rather the one selected in the dropdown

tomaskikutis (approved):

PR #992, opened 2024-07-09
Upgrade Flask to v3 & commands to Click library

eos87:
### Purpose
Migration to Flask v3 and commands to use `click` library. This PR does not go against `develop` branch but `async` instead so `develop` will continue working without disruption.

### What has changed
Most relevant ones:
- Commands have been migrated from `flask-script` to `click` + Flask command line management. Therefore the command runner is no longer `python manage.py <command>` but `flask newsroom <command>` name. At the same time, the commands name remain the same and the their parameters as well
- We're now targeting to minimum python v3.10, Mongo 6 and Elastic 7.17
- Using superdesk-core's async branch 
- Multiple dependencies have been upgraded because of their connection with Flask

This PR is connected with superdesk/superdesk-core#2632 although the one in superdesk is not ready to be merged. There are still multiple things to be checked and adjusted. As you may notice, we're depending on a "unstable" branch from superdesk that contains the most recent async changes and a partial upgrade to Flask v3. My idea is that we depend on that branch until Flask v3 migration is completed (at least stable enough) in superdesk 

Thanks for checking!

Resolves: [SDESK-7280] [SDESK-7281] 

[SDESK-7280]: https://sofab.atlassian.net/browse/SDESK-7280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDESK-7281]: https://sofab.atlassian.net/browse/SDESK-7281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.actions-docker-compose.yml
+++ b/.actions-docker-compose.yml
@@ -14,7 +14,7 @@ services:
             - /data/db
 
     elastic:
-        image: docker.elastic.co/elasticsearch/elasticsearch:7.17.21
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.17.22
         ports:
             - "9200:9200"
         environment:
--- a/.actions-docker-compose.yml
+++ b/.actions-docker-compose.yml
@@ -1,4 +1,5 @@
 version: "2.4"
+name: newsroom-e2e-tests
 services:
     redis:
         image: redis:alpine
@@ -6,14 +7,14 @@ services:
             - "6379:6379"
 
     mongo:
-        image: mongo:3.6
+        image: mongo:4
         ports:
             - "27017:27017"
         tmpfs:
             - /data/db
 
     elastic:
-        image: docker.elastic.co/elasticsearch/elasticsearch:7.10.1
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.17.21
         ports:
             - "9200:9200"
         environment:
--- a/.actions-docker-compose.yml
+++ b/.actions-docker-compose.yml
@@ -7,7 +7,7 @@ services:
             - "6379:6379"
 
     mongo:
-        image: mongo:4
+        image: mongo:6
         ports:
             - "27017:27017"
         tmpfs:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
             - newsroom
 
     elastic:
-        image: docker.elastic.co/elasticsearch/elasticsearch:7.10.1
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.17.22
         networks:
             - newsroom
         environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
             - newsroom
 
     mongo:
-        image: mongo:3.6
+        image: mongo:6
         networks:
             - newsroom
 
--- a/newsroom/commands/__init__.py
+++ b/newsroom/commands/__init__.py
@@ -1,5 +1,4 @@
-# disable "imported but unused" complains
-# flake8: noqa: F401
+# flake8 F401 "imported but unused" disabled in setup.cfg
 from .create_user import create_user
 from .elastic_rebuild import elastic_rebuild
 from .elastic_init import elastic_init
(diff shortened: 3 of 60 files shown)


MarkLark86:
@eos87 Can you remove Python version `3.8` from all Github actions, as we're targeting a minimum of Python v3.10

.actions-docker-compose.yml
```diff
@@ -1,19 +1,20 @@
...
+name: newsroom-e2e-tests
 services:
     redis:
         image: redis:alpine
         ports:
             - "6379:6379"
 
     mongo:
-        image: mongo:3.6
+        image: mongo:4
```
MarkLark86: Would be good to also update this to MongoDB 6 (see superdesk-core's [docker](https://github.com/superdesk/superdesk-core/blob/async/docker-compose.yml))

eos87: done

tests/core/test_oauth_clients.py
```diff
@@ -1,10 +1,12 @@
 from flask import json
+import pytest
 from newsroom.tests.users import test_login_succeeds_for_admin
 from superdesk import get_resource_service
 from newsroom.auth_server.auth import JWTAuth
 import base64
 
 
+@pytest.mark.skip(reason="Pending to figure out some issue with OAuth2")
```
eos87, line 9: I would like to have a separate ticket to take care of this part. Need to understand the whole logic to provide a proper fix.

MarkLark86: We have SDESK-7284 and NHUB-511 specifically to fix breaking changes. We can fix this as part of those tickets

newsroom/commands/__init__.py
```diff
@@ -1,24 +1,27 @@
-from .create_user import create_user  # noqa
-from .elastic_rebuild import elastic_rebuild  # noqa
-from .elastic_init import elastic_init  # noqa
-from .content_reset import content_reset  # noqa
-from .index_from_mongo import index_from_mongo, index_from_mongo_period  # noqa
-from .remove_expired import remove_expired  # noqa
-from .alerts import (  # noqa
+# disable "imported but unused" complains
+# flake8: noqa: F401
```
MarkLark86: From what I read flake8 doesn't support, and don't want to support, file level rule exclusions.
Does this work, or does this disable flake8 for this entire file for all rules?

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
             - newsroom
 
     mongo:
-        image: mongo:3.6
+        image: mongo:6
```
MarkLark86, line 9: Would be good to also update elastic to at least `7.17.22`, for this and e2e/docker-compose (that's what our test infra now uses). https://github.com/superdesk/superdesk-core/blob/async/docker-compose.yml#L9

requirements.txt
```diff
@@ -1,5 +1,5 @@
 WTForms[email]>=3.1,<3.2
-flask>=1.1,<3.1
+flask>=3.0
```
MarkLark86: We could probably remove some of the dependencies from this file (and leave it up to Superdesk-Core to define what versions are supported). Unless removing these has issues, I would recommend the following be removed:

* flask
* click
* flask-babel
* flask_pymongo
* eve-elastic
* MarkupSafe
* reportlab

MarkLark86 (approved):
Looks good, just a couple small things

newsroom/commands/__init__.py
eos87: You're right. I just double checked it and indeed it disabled flake8 entirely for the whole file. Instead, what I did was to add a `per-file-ignores` in the `setup.cfg` because it really bothered me to see the #noqa so many times.

requirements.txt
eos87: Done. It worked well (all tests passed) I just needed to do a minor adjustment to Babel init process.

PR #998, opened 2024-07-17
Create new async UIConfig resource & helper functions [NHUB-525]

devketanpro:
## Checklist


newsroom/ui_config_async.py
```diff
@@ -0,0 +1,41 @@
+from typing import Optional, Dict, Any
+from superdesk.core.module import Module
+from superdesk.core.mongo import MongoResourceConfig
+from superdesk.core.resources import ResourceModel, ResourceModelConfig
+from superdesk.core.resources.service import AsyncResourceService
+from content_api import MONGO_PREFIX
+
+
+class UiConfig(ResourceModel):
+    _id: str
```
MarkLark86: No need to provide `_id` attribute, as it is already provided by `ResourceModel` [here](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/model.py#L64)

newsroom/ui_config_async.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class UiConfigResourceService(AsyncResourceService[UiConfig]):
+    async def get_section_config(self, section_name: str) -> Dict[str, Any]:
+        """Get the section config"""
+        config = await self.find_by_id(section_name)
+
+        if not config:
+            return {}
+        return config.dict()
```
MarkLark86: If you include the args `by_alias=True, exclude_unset=True`, then aliased fields will be properly converted to their field name, and any fields that are not set will not be included.

newsroom/ui_config_async.py
```diff
@@ -0,0 +1,41 @@
...
+
+class UiConfig(ResourceModel):
+    _id: str
+    preview: Optional[Dict[str, Any]] = None
+    details: Optional[Dict[str, Any]] = None
+    list: Optional[Dict[str, Any]] = None
+    advanced_search_tabs: Optional[Dict[str, Any]] = None
+    multi_select_topics: Optional[bool] = None
+    search: Optional[bool] = None
+    enable_global_topics: Optional[bool] = None
```
MarkLark86: Please provide meaningful defaults, especially for booleans. For example, the old `UIConfigResource` sets `enable_global_topics` to `True` as default

MarkLark86:
@devketanpro If you pull the latest from `async` branch, and rebuild `dev-requirements` via `rm dev-requirements.txt && pip-compile dev-requirements.in`, this should fix the tests.

MarkLark86:
@devketanpro Can you place the following in the `requirements.txt` file and see if that fixes the tests not running:
```
# Fix issue with xhtml2pdf not working with python-bidi>=0.5
python-bidi<0.5
```

requirements.txt
```diff
@@ -11,6 +11,7 @@ xhtml2pdf>=0.2.4
 sentry-sdk[flask]>=1.5.7,<2.7
 python3-saml>=1.15,<1.17
 google-auth>=2.6,<2.30
+nest-asyncio>=1.6.0
```
MarkLark86: Please don't use `nest-asyncio`, it is modifying the behaviour of the asyncio library. If something does not work with this library, we should address the reason and not patch asyncio to make it work.

newsroom/settings.py
```diff
@@ -109,15 +113,33 @@ def validate_general_settings(values):
                 return gettext("{}: Email IDs not in proper format".format(field_txt))
 
 
+def sync_get_setting(setting_key=None):
+    loop = asyncio.get_event_loop()
```
MarkLark86: Rather than doing this, it would be good to await the appropriate data first before rendering the template.
Also, does Flask3/Jinja not support async inside templates?

devketanpro: yep jinja is not supporting async functions inside templates

devketanpro: Actually, I tried different approaches to use these async functions in the templates, but in the end, the issue was solved by this hack
The Flask operates synchronously, meaning it handles requests and responses in a blocking manner. This design doesn't support asynchronous operations directly. Ifwe try to use await in this context, it can block the event loop and cause runtime errors, like RuntimeError: This event loop is already running

newsroom/auth/views.py
```diff
@@ -367,7 +367,7 @@ def set_locale():
 
 @blueprint.route("/auth/impersonate", methods=["POST"])
 @admin_only
-def impersonate_user():
+async def impersonate_user():
```
MarkLark86: If a function does not use any async code, then it's not required to turn that function into async itself

newsroom/settings.py
```diff
@@ -48,12 +57,12 @@ def update_values():
...
 
-def get_initial_data(setting_key=None):
-    data = get_setting(setting_key=setting_key, include_audit=True)
+async def get_initial_data(setting_key=None):
+    data = await get_setting(setting_key=setting_key, include_audit=True)
     return data
 
 
-def get_setting(setting_key=None, include_audit=False):
+async def get_setting(setting_key=None, include_audit=False):
```
MarkLark86: Looks like `await` is not used anywhere in this function, therefor there is no need to make this function async. You would then be able to remove all of the `await get_setting` changes you have added

newsroom/ui_config_async.py
```diff
@@ -0,0 +1,40 @@
...
+from superdesk.core.resources.service import AsyncResourceService
+from content_api import MONGO_PREFIX
+
+
+class UiConfig(ResourceModel):
+    preview: Optional[Dict[str, Any]] = None
+    details: Optional[Dict[str, Any]] = None
+    list: Optional[Dict[str, Any]] = None
+    advanced_search_tabs: Optional[Dict[str, Any]] = None
+    multi_select_topics: Optional[bool] = False
```
MarkLark86: If you're providing a default other than `None`, then the `Optional` annotation is not required.

Annotated with `Optional[<type>]` indicates that this variable accepts a value of `None`, which isn't always the case

newsroom/users/users.py
```diff
@@ -21,6 +21,7 @@
 from newsroom.user_roles import UserRole
 from newsroom.signals import user_created, user_updated, user_deleted
 from newsroom.companies.utils import get_company_section_names, get_company_product_ids
+import asyncio
```
MarkLark86: Where is `asyncio` lib used in this module?

newsroom/users/users.py
```diff
@@ -361,7 +362,7 @@ def check_permissions(self, doc, updates=None):
             if doc.get("company") and doc["company"] == manager.get("company"):
                 allowed_updates = (
                     COMPANY_ADMIN_ALLOWED_UPDATES
-                    if not get_setting("allow_companies_to_manage_products")
+                    if not await get_setting("allow_companies_to_manage_products")
```
MarkLark86: Through changing `get_setting` back to sync, this entire module would no longer require the `async` changes

MarkLark86:
@devketanpro When you have finished making all these changes, please mark the PR as draft and I will work on getting the tests working (having issues myself with async tests)

newsroom/ui_config_async.py
```diff
@@ -0,0 +1,40 @@
...
+from content_api import MONGO_PREFIX
+
+
+class UiConfig(ResourceModel):
+    preview: Optional[Dict[str, Any]] = None
+    details: Optional[Dict[str, Any]] = None
+    list: Optional[Dict[str, Any]] = None
+    advanced_search_tabs: Optional[Dict[str, Any]] = None
+    multi_select_topics: bool = False
+    search: Optional[bool] = None
```
MarkLark86: Is `None` a valid value for the `search` field?

devketanpro: Yes, using `Optional[bool] = None` is valid for the `search` field

MarkLark86: I'm confused as to how. Do we store `None` somewhere for this field? It is a boolean after all, so a default value of `False` should suffice no?

newsroom/settings.py
```diff
@@ -23,10 +24,11 @@ def get_settings_collection():
 
 @blueprint.route("/settings/<app_id>")
 @account_manager_only
-def app(app_id):
+async def app(app_id):
     for app in flask.current_app.settings_apps:
         if app._id == app_id:
-            return flask.render_template("settings.html", setting_type=app_id, data=app.data())
+            data = await app.data() if iscoroutine(app.data()) else app.data()
```
MarkLark86: Here you're calling `app.data()` twice, once for assignment and the other for if condition.
It would be better to call it once, and check the response value in the condition

newsroom/ui_config_async.py
devketanpro: In some config, a search field is missing.

newsroom/agenda/views.py
```diff
@@ -95,11 +100,16 @@ def item(_id):
...
 
-    data = get_view_data()
+    data = await get_view_data()
     data["item"] = item
-    return flask.render_template("agenda_index.html", data=data, title=item.get("name", item.get("headline")))
+    return flask.render_template(
+        "agenda_index.html",
+        data=data,
+        title=item.get("name", item.get("headline"), user_profile_data=user_profile_data),
+    )
```
MarkLark86, line 113: There is an error here in pytests:  `dict.get() takes no keyword arguments`. I'm guessing this keyword argument was meant for the `render_template` call instead?

MarkLark86 (approved):

newsroom/agenda/views.py
devketanpro: Yes, you're right. Thanks for pointing that out!

PR #1002, opened 2024-07-23
[NHUB-526] api: Upgrade Company views to async

MarkLark86:
Depends on https://github.com/superdesk/superdesk-core/pull/2654


newsroom/utils.py
```diff
@@ -13,6 +13,7 @@
 from superdesk.json_utils import try_cast
 from superdesk.etree import parse_html
 from superdesk.text_utils import get_text
+from superdesk.core.web import Request
```
devketanpro: unused import

MarkLark86: Issue from merge parent branch. Fixed now

devketanpro (approved):

PR #1004, opened 2024-07-25
Fix profile notification

thecalcc:
CPCN-868

## Checklist


thecalcc:
I cannot remove the whitespace changes, as my linter is configured to lint on save.

tomaskikutis (approved):
That's fine since it's only a few lines.

PR #1008, opened 2024-07-29
Create new async OAuthClient resource, blueprint & helper functions [NHUB-528]

devketanpro:
## Checklist


newsroom/oauth_clients/clients_async.py
```diff
@@ -0,0 +1,47 @@
...
+from superdesk.core.resources import ResourceModel, ResourceConfig, MongoResourceConfig
+from superdesk.core.resources.service import AsyncResourceService
+from superdesk.core.web import EndpointGroup
+from pydantic import Field
+import logging
+from bson import ObjectId
+
+
+class ClientResource(ResourceModel):
+    id: Annotated[Union[str, ObjectId], Field(alias="_id")] = None
```
MarkLark86: `id` is not required, as it is already defined in `ResourceModel`

newsroom/oauth_clients/clients_async.py
```diff
@@ -0,0 +1,47 @@
...
+    name: str
```
MarkLark86: `name` must be unique, so you will need to add the unique [validator](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/validators.py#L173) ([example usage](https://github.com/superdesk/superdesk-core/blob/async/tests/core/modules/users/types.py#L39))

newsroom/oauth_clients/clients_async.py
```diff
@@ -0,0 +1,47 @@
...
+    password: str
+    last_active: Optional[str] = None
+    etag: Annotated[Optional[str], Field(alias="_etag")] = None
```
MarkLark86, line 20: We're not currently using `_etag` with new code. This will be done in a later ticket (and added to the `ResourceModel`)

newsroom/oauth_clients/clients_async.py
```diff
@@ -0,0 +1,47 @@
...
+from pydantic import Field
+import logging
+from bson import ObjectId
+
+
+class ClientResource(ResourceModel):
+    id: Annotated[Union[str, ObjectId], Field(alias="_id")] = None
+    name: str
+    password: str
+    last_active: Optional[str] = None
```
MarkLark86: `last_active` should be a `datetime` type

newsroom/oauth_clients/clients_async.py
```diff
@@ -0,0 +1,47 @@
...
+    etag: Annotated[Optional[str], Field(alias="_etag")] = None
+
+
+class ClientService(AsyncResourceService[ClientResource]):
+    """Service class for managing OAuthClient resources"""
+
+    resource_name = "oauth_clients"
+
+    async def get_all_client(self) -> List[Dict]:
```
MarkLark86: The naming of this function should be plural, as it potentially returns more than 1

newsroom/oauth_clients/clients_async.py
```diff
@@ -0,0 +1,53 @@
...
+from superdesk.core.resources.service import AsyncResourceService
+from superdesk.core.web import EndpointGroup
+
+
+from content_api import MONGO_PREFIX
+
+
+class ClientResource(ResourceModel):
+    name: Annotated[
+        Optional[str],
```
MarkLark86: `name` is a required field, therefor we should not use `Optional[str]` as it's type

newsroom/oauth_clients/clients_async.py
```diff
@@ -0,0 +1,53 @@
...
+
+from superdesk.core.resources import ResourceModel, ResourceConfig, MongoResourceConfig, validators
+from superdesk.core.resources.service import AsyncResourceService
+from superdesk.core.web import EndpointGroup
+
+
+from content_api import MONGO_PREFIX
+
+
+class ClientResource(ResourceModel):
```
MarkLark86: Because we're using `ObjectId` for the `_id` of this model, we should probably extend from `ResourceModelWithObjectId` instead of `ResourceModel`

**ResourceModelWithObjectId** specifically states that we're using ObjectId

devketanpro: Thanks for the update

MarkLark86 (approved):

PR #1011, opened 2024-08-01
[SDESK-7286] Use new Async GridFS MediaStorage in newsroom-core

eos87:
### Purpose 
To implement the new Async GridFS media storage class into newsroom-core

### What has changed
- The `upload.py` has been replace with a proper module according to the new Superdesk Async Framework
- Refactored the view responsible for rendering the objects taken from the media storage (`/assets/FILE-ID`)
- Simplified the code to try to make it easier to understand and maintain (split functions, added `utils.py` module)
- Adjusted the endpoints to use the new async storage and new endpoints approach
- Added multiple unit tests
- ‼️Added temporary method `_get_current_object` into `NewsroomWebApp` class to reduce the noise from broken tests because of the absence of such method. We should remove it and implemente proper fix (perhaps avoid using that method). This method is mostly used when broadcasting signals with `Blinker` 

### Checklist

Resolves: [SDESK-7286]

[SDESK-7286]: https://sofab.atlassian.net/browse/SDESK-7286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/factory/app.py
```diff
@@ -138,6 +141,7 @@ def setup_media_storage(self):
             self.media = AmazonMediaStorage(self)
```
MarkLark86, line 143: Until we add async S3 storage support, we can probably add `self.media_async = GridFSMediaStorageAsync()` here, otherwise if S3 is enabled `app.media_async` will be None

newsroom/assets/utils.py
```diff
@@ -0,0 +1,101 @@
...
+from superdesk.core import get_current_app
+from superdesk.upload import upload_url as _upload_url
+from superdesk.media.media_operations import guess_media_extension
+
+from .module import ASSETS_ENDPOINT_GROUP_NAME, ASSETS_RESOURCE
+
+CACHE_MAX_AGE = 3600 * 24 * 7  # 7 days
+
+
+def get_newsroom_app():
```
MarkLark86: This would be good to have in the new [newsroom.core](https://github.com/superdesk/newsroom-core/blob/async/newsroom/core/__init__.py) module, so that Newshub specific app features can be imported from there.

newsroom/news_api/default_settings.py
```diff
@@ -33,7 +33,6 @@
     "newsroom.products",
     "newsroom.news_api.api_audit",
     "newsroom.news_api.news.assets.assets",
-    "newsroom.upload",
```
MarkLark86, line 36: Would we need `MODULES = ["newsroom.assets.module"]` here so image downloads from the NewsAPI work?

newsroom/wire/views.py
```diff
@@ -56,7 +56,8 @@
 
 from .search import get_bookmarks_count
 from .items import get_items_for_dashboard
-from ..upload import ASSETS_RESOURCE, get_upload
+from newsroom.assets.module import ASSETS_RESOURCE
```
MarkLark86: If we include `__all__ = ["ASSETS_RESOURCE", "get_upload", ...]` etc into the `newsroom.assets.__init__.py` file, then we can provide an interface into the Assets module and expose only what other modules need.

Then we can do the following:
`from newsroom.asssets import ASSETS_RESOURCE, get_upload`

newsroom/cards/views.py
```diff
@@ -28,6 +29,10 @@ def get_settings_data():
     }
 
 
+def _get_file_from_request(key: str) -> Optional[Any]:
```
MarkLark86: This might be useful in the new [newsroom.flask](https://github.com/superdesk/newsroom-core/blob/async/newsroom/flask.py) module, so we can support flask features we use from the one file (until we add this feature into the new Endpoint Request/Response system)

MarkLark86 (commented):
Looks good, just a couple small improvements

eos87:
Hi @MarkLark86, I've introduced your suggested feedback. Thanks! I also mark as skipped a couple of tests as some of the views are yet to be moved to async. 
Please take another look when you have some time.

MarkLark86 (approved):

PR #1015, opened 2024-08-06
when using default filter keep the value empty

petrjasek:
and treat it like no filter is selected in the UI

CPCN-908

## Checklist


assets/agenda/components/AgendaApp.tsx
```diff
@@ -6,7 +6,7 @@ import {get, isEmpty} from 'lodash';
 
 import {ISearchSortValue} from 'interfaces';
 import {getItemFromArray, gettext} from 'utils';
```
devketanpro: remove the unused import

devketanpro (approved):

PR #1019, opened 2024-08-08
[SDESK-7287] Support async tasks in Celery

eos87:
### Purpose
Introduce the new class from `superdesk-core` to handle celery tasks with both async and sync code. This PR depends on https://github.com/superdesk/superdesk-core/pull/2658

### What has changed
- Created a new class `NewsroomContextTask` that inherits from superdesk's `HybridAppContextTask`. This has helped removing a lot of duplicated code
- Registered custom serializer for celery using superdesk's new `ContextAwareSerializerFactory` class.
- Removed `LockedError` exception class as it was basically useless (just to raise it and log a debug message)

### To test
You can, for instance, change the task `monitoring_immediate_alerts`, prefix it with `async` and add some asyncio.sleep(0.1) and a `print` statement inside. Then run the services (celery worker and beat) and observe that the task runs properly

Resolves: [SDESK-7287]

### Checklist

[SDESK-7287]: https://sofab.atlassian.net/browse/SDESK-7287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/push.py
+++ b/newsroom/push.py
@@ -20,7 +20,6 @@ from superdesk.utc import utcnow
 from superdesk.errors import SuperdeskApiError
 from planning.common import WORKFLOW_STATE
 from newsroom.celery_app import celery
-from newsroom.errors import LockedError
 from superdesk.lock import lock, unlock
 
 from newsroom.notifications import (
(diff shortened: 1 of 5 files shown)


newsroom/push.py
```diff
@@ -765,7 +764,9 @@ def set_item_reference(coverage):
 def locked(_id: str, service: str):
     lock_name = f"notify-{service}-{_id}"
     if not lock(lock_name, expire=300):
-        raise LockedError(lock_name)
+        logger.debug(f"Lock conflict on {lock_name}")
```
MarkLark86, line 767: I think raising an exception was intentional, maybe to get the exception/stack trace pushed to sentry for further investigation, @petrjasek?

MarkLark86: Scratch that, I just noticed the exception catch for this type

MarkLark86 (approved):

PR #1021, opened 2024-08-09
update agenda filter functionality to handle default filters and also…

devketanpro:
… add some more filters query params[CPCN-910]

https://github.com/superdesk/newsroom-app-cp/pull/186

## Checklist

### DIFF
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -45,7 +45,6 @@ import {searchParamsSelector, searchFilterSelector} from 'search/selectors';
 import {clearAgendaDropdownFilters} from '../local-store';
 import {getLocations, getMapSource} from '../maps/utils';
 import {ILocation} from 'interfaces/agenda';
-import {IDateFilter} from 'interfaces/common';
 
 const WATCH_URL = '/agenda_watch';
 const WATCH_COVERAGE_URL = '/agenda_coverage_watch';
@@ -323,14 +322,12 @@ function getAgendaSearchParamsFromState(state: IAgendaState): AgendaSearchParams
         ) ? agendaDate : createdFilter.from;
     }
 
-    const defaultFilter = state.dateFilters?.find((filter:any)=> filter.default);
-
     return {
         itemType: itemTypeFilter,
         searchParams: searchParams,
         featured: featuredFilter,
         fromDate: fromDate,
-        date_filter: state?.search?.createdFilter?.date_filter || defaultFilter?.filter,
+        date_filter: state?.search?.createdFilter?.date_filter,
         toDate: createdFilter.from?.startsWith('now/') ?
             createdFilter.from :
             createdFilter.to,
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -373,16 +373,16 @@ function setListGroupsAndLoadHiddenItems(items: Array<IAgendaItem>, next?: boole
                     maxDate = moment().startOf('day');
                 }else if (activeFilterQuery === 'now+7d/d') { 
                     minDate = moment().startOf('day');
-                    maxDate = moment().clone().add(7, 'days').endOf('day');
+                    maxDate = moment().add(7, 'days').endOf('day');
                 } else if (activeFilterQuery === 'now+30d/d') {
                     minDate = moment().startOf('day');
-                    maxDate = moment().clone().add(30, 'days').endOf('day');
+                    maxDate = moment().add(30, 'days').endOf('day');
                 } else if (activeFilterQuery === 'now+3M/d') {
                     minDate = moment().startOf('day');
-                    maxDate = moment().clone().add(90, 'days').endOf('day');
+                    maxDate = moment().add(90, 'days').endOf('day');
                 } else if (activeFilterQuery === 'now+12M/d') {
                     minDate = moment().startOf('day');
-                    maxDate = moment().clone().add(12, 'months').endOf('day');
+                    maxDate = moment().add(12, 'months').endOf('day');
                 } else {
                     minDate = moment().startOf('day');
                     maxDate = minDate.clone();
--- a/assets/agenda/actions.ts
+++ b/assets/agenda/actions.ts
@@ -45,6 +45,7 @@ import {searchParamsSelector, searchFilterSelector} from 'search/selectors';
 import {clearAgendaDropdownFilters} from '../local-store';
 import {getLocations, getMapSource} from '../maps/utils';
 import {ILocation} from 'interfaces/agenda';
+import {IDateFilter} from 'interfaces/common';
 
 const WATCH_URL = '/agenda_watch';
 const WATCH_COVERAGE_URL = '/agenda_coverage_watch';
@@ -234,6 +235,7 @@ function search(state: IAgendaState, fetchFrom: number): Promise<IRestApiRespons
         featured,
         fromDate,
         toDate,
+        date_filter
     } = getAgendaSearchParamsFromState(state);
 
     const params: any = {
@@ -245,6 +247,7 @@ function search(state: IAgendaState, fetchFrom: number): Promise<IRestApiRespons
         from: fetchFrom,
         date_from: fromDate,
         date_to: toDate,
+        date_filter: date_filter,
         timezone_offset: getTimezoneOffset(),
         featured: featured,
         itemType: itemType,
@@ -290,6 +293,7 @@ interface AgendaSearchParams {
     featured: boolean;
     fromDate?: string;
     toDate?: string;
+    date_filter?: string;
 }
 
 function getAgendaSearchParamsFromState(state: IAgendaState): AgendaSearchParams {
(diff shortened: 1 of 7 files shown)


newsroom/agenda/agenda.py
```diff
@@ -318,10 +313,19 @@ def _agenda_query():
...
     date_range = {}
     offset = int(args.get("timezone_offset", "0"))
+
+    agenda_filters = app.config.get("AGENDA_TIME_FILTERS")
+    default_filter = next((f for f in agenda_filters if f.get("default", False)), None)
+
     if args.get("date_from"):
-        date_range["gt"] = get_local_date(args["date_from"], "00:00:00", offset)
+        if default_filter and not args.get("date_to") and args.get("date_from") == get_utcnow().strftime("%Y-%m-%d"):
+            date_range["gt"] = get_local_date(default_filter["query"], "00:00:00", offset)
```
petrjasek: it seems like those filters are always limiting the upper bound not the lower one, so to make it work properly we should probably send that as a `date_to` or a new `date_filter` option and work with that

newsroom/utils.py
```diff
@@ -519,6 +519,13 @@ def format_date(date, time, offset):
         return _today - timedelta(days=7)
     if date == "now-30d/d":
         return _today - timedelta(days=30)
+    if date == "now30d/d":
```
petrjasek: I wonder where those values are coming from, seems like they are wrong filters

devketanpro: we defined here check this PR:- https://github.com/superdesk/newsroom-app-cp/pull/186
In the previous implementation, the date range was defined as `now+30d/d`. However, due to URL encoding, the `+` sign was replaced with a `space` in the payload, resulting in `now 30d/d.`

petrjasek: seems like a wrong PR link

devketanpro: https://github.com/superdesk/newsroom-app-cp/pull/186/files#diff-d8def684c9a77cfaaa19b4c1762a220d5978a030144169de2a502eedd110651bR419

petrjasek: I see, might be better to define some descriptive `filter` param then and resolve the query only on server, we don't need it in url

newsroom/agenda/agenda.py
petrjasek: actually thinking about it the filtering for those options should filter events which
end after the given filter, so if the filter is next 7 days the filter should be:
```
date_range["lt"] = default_filter["query"]
```
which will later use a filter like:
```
dates.end <= date_range["lt"]
```
and you can keep the `date_range["gt"]` for selected day, that should still work the same probably.

newsroom/agenda/agenda.py
```diff
@@ -315,13 +310,40 @@ def _agenda_query():
...
+    agenda_filter = get_date_filter_query(args.get("date_filter"))
+    date_from = args.get("date_from")
+
+    if agenda_filter:
+        if date_from:
+            start_date = get_local_date(date_from, "00:00:00", offset)
+            if agenda_filter.startswith("now-"):  # Handle past filters (e.g., last_7_days)
+                date_range["gt"] = get_local_date(agenda_filter, "00:00:00", offset)
+                date_range["lt"] = start_date
+            elif agenda_filter.startswith("now+"):  # Handle future filters (e.g., next_7_days)
```
petrjasek, line 341: I wonder how it handles the default config where it's just `now/d`, can you add some test where it would actually run those queries?

devketanpro: we always send default filter UI in the payload

petrjasek: we don't send it anymore, that was causing issues in the UI

petrjasek: but what I mean is rather here:
```py
    {"name": lazy_gettext("Today"), "query": "now/d", "filter": "today"},
    {"name": lazy_gettext("This Week"), "query": "now/w", "filter": "this_week"},
    {"name": lazy_gettext("This Month"), "query": "now/M", "filter": "this_month"},
```
because it seems like those queries are not covered, but we use them as defaults

devketanpro: >we don't send it anymore; that was causing issues in the UI

I have made the changes in this PR to send default `date_filter`

assets/interfaces/common.ts
```diff
@@ -86,4 +86,5 @@ export interface IDateFilter {
     default?: boolean,
     filter?: string;
     query?: string;
+    find?: any;
```
petrjasek: is this used somewhere?

newsroom/agenda/agenda.py
petrjasek: I see, but why? the server knows which one is the default

assets/interfaces/common.ts
devketanpro: yep here :- https://github.com/superdesk/newsroom-core/pull/1021/files#diff-48af75964c59c6cf135686fe342f79b0e2743556429529a579f8e2c42af4faf0R326

petrjasek: that looks like a wrong type then, there it should be an array of IDateFilter which already has find method

newsroom/agenda/agenda.py
devketanpro: added the case which handles these default filters, also removed functionality to send default `date_filter`

newsroom/agenda/agenda.py
```diff
@@ -313,6 +313,9 @@ def _agenda_query():
 def get_date_filter_query(filter_name: str):
     """Get the query for the given filter name."""
     time_filters = app.config.get("AGENDA_TIME_FILTERS", [])
+    default_filter = next((f for f in time_filters if f.get("default")), None)
+    if not filter_name:
+        return default_filter["query"]
```
petrjasek: this would be an error if there is no default filter configured

newsroom/agenda/agenda.py
```diff
@@ -315,13 +310,46 @@ def _agenda_query():
...
+            start_date = get_local_date(date_from, "00:00:00", offset)
+            if agenda_filter.startswith("now-"):  # Handle past filters (e.g., last_7_days)
+                date_range["gt"] = get_local_date(agenda_filter, "00:00:00", offset)
+                date_range["lt"] = start_date
+            elif agenda_filter.startswith("now+"):  # Handle future filters (e.g., next_7_days)
+                date_range["gt"] = start_date
+                date_range["lt"] = get_end_date(agenda_filter, start_date)
+            elif agenda_filter.startswith("now"):  # Handle current filters (e.g., this_week)
+                date_range["gt"] = get_local_date(agenda_filter, "00:00:00", offset)
+                date_range["lt"] = get_end_date(agenda_filter, get_local_date(agenda_filter, "23:59:59", offset))
```
petrjasek, line 346: I wouldn't use that `get_local_date` on those filters, that requires us to have all the options defined in core and if someone will want *next 10 months* it won't work without core changes, can we rather use those queries as they are without converting them to dates?

devketanpro: for all filters, as we already do it for wire section

devketanpro: so in this case, we have to update the UI grouping as well , because the list grouping is also using this behaviour currently

petrjasek: list grouping is working based on event dates, not filter query afaik

devketanpro: see this :- https://github.com/superdesk/newsroom-core/blob/develop/assets/agenda/actions.ts#L333-L392

petrjasek: I see, hm that's probably quite complex then, not sure how the client would know what to display then

devketanpro: that's exactly why I didn't implement the behavior you suggested

petrjasek: ok so lets keep it simple for now

petrjasek: so that client code needs to be updated too right? to support those `now+30d` etc

devketanpro: yep

assets/agenda/actions.ts
```diff
@@ -340,31 +343,46 @@ function setListGroupsAndLoadHiddenItems(items: Array<IAgendaItem>, next?: boole
...
+                    maxDate = moment().clone().add(7, 'days').endOf('day');
+                } else if (activeFilterQuery === 'now+30d/d') {
+                    minDate = moment().startOf('day');
+                    maxDate = moment().clone().add(30, 'days').endOf('day');
+                } else if (activeFilterQuery === 'now+3M/d') {
+                    minDate = moment().startOf('day');
+                    maxDate = moment().clone().add(90, 'days').endOf('day');
+                } else if (activeFilterQuery === 'now+12M/d') {
+                    minDate = moment().startOf('day');
+                    maxDate = moment().clone().add(12, 'months').endOf('day');
```
petrjasek: since `moment()` creates new instance there is no need to call `clone()` on it

petrjasek (approved):

PR #1028, opened 2024-08-13
Fix headline colors on read/unread

thecalcc:
STTNHUB-355

## Checklist


tomaskikutis:
what's the ticket here?

assets/agenda/components/AgendaListItem.tsx
```diff
@@ -132,7 +132,9 @@ class AgendaListItem extends React.Component<IProps> {
                 'wire-articles__item--not-covering': !hasCoverages(this.props.item),
                 'wire-articles__item--postponed': isPostponed(this.props.item),
                 'wire-articles__item--canceled': isCanceled(this.props.item),
-                'wire-articles__item--visited': this.props.isRead,
+
+                // Planning items don't have a version, while event items do, which is affecting the `isRead` prop.
```
tomaskikutis: This looks like a wrong place to do it. This line looks good
`'wire-articles__item--visited': this.props.isRead,`

if item is read - apply visited style. If `isRead` is not indicating correctly, it should be fixed in a parent component rather than adjusting for incorrectness here.

thecalcc: Yes, that is the second option. I went with this one, because I can't know all the item states, and didn't want to risk breaking anything.

assets/wire/utils.ts
```diff
@@ -8,19 +8,21 @@ export const DISPLAY_ABSTRACT = getConfig('display_abstract');
...
-
+const DEFAULT_NO_VERSION = 0;
 /**
  * Returns the item version as integer
  *
  * @param {Object} item
  * @returns {number}
  */
-export function getIntVersion(item: IArticle | IAgendaItem | undefined): number | undefined {
+export function getIntVersion(item: IArticle | IAgendaItem | undefined): number {
```
tomaskikutis: This is hacky as well. It was returning a number or undefined which makes sense. Now you're doing a unnecessary breaking change where it's not possible to distinguish value zero from no value.

A better approach would be to change it here
```
const version = getIntVersion(this.props.itemsById[itemId]);

isRead={version == null || version === this.props.readItems[itemId]}
```

petrjasek (approved):

PR #1030, opened 2024-08-14
Udate config

thecalcc:
STTNHUB-289

## Checklist


assets/search/components/SearchResultsBar/index.tsx
```diff
@@ -105,8 +105,9 @@ class SearchResultsBarComponent extends React.Component<IProps, IState> {
         super(props);
 
         this.topicNotNull = new URLSearchParams(window.location.search).get('topic') != null;
+
         this.state = {
-            isTagSectionShown: this.props.initiallyOpen || this.topicNotNull,
+            isTagSectionShown: !COLLAPSED_SEARCH_BY_DEFAULT ?? (this.props.initiallyOpen || this.topicNotNull),
```
petrjasek: I'm a bit lost with the logic here, so if that `COLLAPSED_SEARCH_BY_DEFAULT` is `false` this will be set to `true` ?

petrjasek: imo what we want is:
```ts
isTagSectionShown: COLLAPSED_SEARCH_BY_DEFAULT === true ? false : (this.props.initiallyOpen || this.topicNotNull),
```

petrjasek (approved):

PR #1035, opened 2024-08-14
update agenda search date filter elastic queries [CPCN-910] 

devketanpro:
## Checklist

### DIFF
--- a/newsroom/agenda/agenda.py
+++ b/newsroom/agenda/agenda.py
@@ -386,91 +386,131 @@ def _set_event_date_range(search):
 
     should = []
 
+    # Scenario 1: Default query with only date_from
     if date_from and not date_to:
-        # Filter from a particular date onwards
-        should = gen_date_range_filter("dates.end", "gte", search.args.get("date_from"), date_from)
+        should.extend(
+            [
+                # Events that start after or on date_from
+                {
+                    "bool": {
+                        "should": [
+                            {"range": {"dates.start": {"gte": date_from}}},
+                            {"range": {"dates.end": {"gte": date_from}}},
+                        ],
+                        "must_not": {"term": {"dates.all_day": True}},
+                    }
+                },
+                # All-day events starting or ending after date_from
+                {
+                    "bool": {
+                        "should": [
+                            {"range": {"dates.start": {"gte": date_from}}},
+                            {"range": {"dates.end": {"gte": date_from}}},
+                        ],
+                        "filter": {"term": {"dates.all_day": True}},
+                    }
+                },
+            ]
+        )
+
+    # Scenario 2 and 3: Query with specific filters
     elif not date_from and date_to:
-        # Filter up to a particular date
-        should = gen_date_range_filter("dates.end", "lte", search.args.get("date_to"), date_to)
+        # Apply date range filters for next_7_days, next_30_days, etc.
+        should.extend(
+            [
+                {
+                    "bool": {
+                        "should": [
+                            {"range": {"dates.start": {"gte": date_range["gt"], "lte": date_range["lt"]}}},
+                        ],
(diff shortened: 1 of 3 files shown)


tests/core/test_agenda.py
```diff
@@ -796,3 +796,123 @@ def test_agenda_filters_query(app):
...
+                "all_day": True,
+            },
+        },
+    ]
+
+    app.data.insert("agenda", agenda_items)
+
+    # Scenario 1: Default query with only date_from
+    data = get_json(client, "/agenda/search?date_from=2024-08-11")
+    assert set([item["_id"] for item in data["_items"]]) == {
```
petrjasek, line 884: would be good to have an item which is out of the default filter to actually see it's working

petrjasek: both before and after

newsroom/agenda/agenda.py
```diff
@@ -386,91 +386,131 @@ def _set_event_date_range(search):
...
     if date_from and not date_to:
-        # Filter from a particular date onwards
-        should = gen_date_range_filter("dates.end", "gte", search.args.get("date_from"), date_from)
+        should.extend(
+            [
+                # Events that start after or on date_from
+                {
+                    "bool": {
+                        "should": [
+                            {"range": {"dates.start": {"gte": date_from}}},
```
petrjasek, line 397: that part is not needed imo, we want to have events starting before or after, so no need to filter that, as long as those events end after the selected date it should be fine

newsroom/agenda/agenda.py
```diff
@@ -386,91 +386,131 @@ def _set_event_date_range(search):
...
+                    }
+                },
+            ]
+        )
+
+    # Scenario 2 and 3: Query with specific filters
     elif not date_from and date_to:
-        # Filter up to a particular date
-        should = gen_date_range_filter("dates.end", "lte", search.args.get("date_to"), date_to)
+        # Apply date range filters for next_7_days, next_30_days, etc.
```
petrjasek, line 418: it's kind of confusing that it's checking if `date_to` is set and then it's not used

newsroom/agenda/agenda.py
```diff
@@ -386,91 +386,131 @@ def _set_event_date_range(search):
...
-                "bool": {
-                    "filter": [
-                        {"range": {"dates.start": {"gte": date_from}}},
-                        {"range": {"dates.end": {"lte": date_to}}},
-                    ],
-                    "must_not": {"term": {"dates.all_day": True}},
+        # Filter based on both date_from and date_to
+        should.extend(
+            [
+                # Events that start and end within the range
```
petrjasek, line 444: what we actually want here is events that stared before the end and their end date is after the start,
so:
```py
{"must": [
    {"range": {"dates.start": {"lte": date_to}}},
    {"range": {"dates.end": {"gte": date_from}}},
]}
```
should cover all the cases, we just need to have also the all_day version

tests/core/test_agenda.py
```diff
@@ -796,3 +796,123 @@ def test_agenda_filters_query(app):
     date_range = get_date_filters(args)
     assert date_range["gt"] == datetime(2024, 8, 14, 0, 0, tzinfo=pytz.UTC)
     assert date_range["lt"] == datetime(2024, 8, 14, 23, 59, 59, tzinfo=pytz.UTC)
+
+
+def test_agenda_filters_items(client, app):
+    app.config.update(
+        {
+            "AGENDA_TIME_FILTERS": [
+                {"name": "Next 7 days", "filter": "next_7_days", "query": "now+7d/d"},
```
petrjasek, line 805: if I get it right the actual query is never used, is it?

petrjasek:
it's won't do for now

PR #1036, opened 2024-08-15
[NHUB-530] Implement async users management 

eos87:
### Purpose
Implement Users Management section in new Async Framework

### What's changed
- Create new module, resource model, service and helpers using async framework 
- Update references and adjusting parts of the app to use the new toys
- Fixed and revisited many tests that got broken because of the adjustments 

### What's pending
- Remove some references to `get_resource_service("users")` which were left there because of the time already used in this tasks
~- Fix a couple of tests that are failing also left out because of the time (behave and e2e)~

Thanks for checking!

## Checklist


newsroom/companies/__init__.py
```diff
@@ -1,39 +1,32 @@
+import superdesk
+
 from flask_babel import lazy_gettext
 
-import superdesk
+from newsroom.companies.companies_async.types import CompanyResource
```
MarkLark86: `CompanyResource` is exposed by the top level module, so you can import by using:

```python
from newsroom.companies import CompanyResource
```

This is a good example of a module having a defined interface of what it exposes. It might be good if we try and follow that with the modules, so we're not importing from too deep within the modules. `from newsroom.<x> import <y>`.

newsroom/companies/__init__.py
```diff
@@ -1,39 +1,32 @@
...
-
-    company = superdesk.get_resource_service("companies").find_one(req=None, _id=company_id)
 
-    rv = {
-        "monitoring_administrator": (company or {}).get("monitoring_administrator"),
-        "userSections": app.sections,
-    }
-    if company and company.get("sections"):
-        rv["userSections"] = [s for s in app.sections if company.get("sections").get(s["_id"])]
+    if not company or user.user_type == UserRole.ADMINISTRATOR.value:
```
MarkLark86: When checking against an enum, there is no longer the need to use `.value` from the enum attribute.

i.e.
```console
>>> from newsroom.user_roles import UserRole
>>> from pydantic import BaseModel
>>> 
>>> class User(BaseModel):
...     name: str
...     role: UserRole
... 
>>> user = User(name="Mark", role=UserRole.PUBLIC)
>>> user.role
<UserRole.PUBLIC: 'public'>
>>> user.role == UserRole.PUBLIC
True
>>> user.role == UserRole.INTERNAL
False
```

newsroom/companies/__init__.py
eos87: yeah, I think this was just auto import from the editor. I'll fix it

newsroom/users/service.py
```diff
@@ -0,0 +1,181 @@
...
+from newsroom.companies.utils import get_updated_products, get_updated_sections
+
+from .model import UserResourceModel
+from .users import COMPANY_ADMIN_ALLOWED_PRODUCT_UPDATES, COMPANY_ADMIN_ALLOWED_UPDATES, USER_PROFILE_UPDATES
+
+
+class UsersService(NewshubAsyncResourceService[UserResourceModel]):
+    resource_name = "users"
+
+    def get_model_instance_from_dict(self, data: Dict[str, Any]) -> ResourceModelType:
```
MarkLark86: Rather than supported invalid data, what if we have a data update script that fixes the data. That way we don't need this custom code to support invalid data in the DB

eos87: Yeah, that makes sense. Do you mean like those from this command `python manage.py data:generate_update`? If so, yes we should do it. Although I'm not sure if these commands are working at the moment. Perhaps we should have this in a separate small ticket (also to remove this workaround). Thoughts?

MarkLark86: Yes that's the one. Those commands should work, and seem to work on my local machine (although we're missing `manage.py` in newsroom-core repo, so I generally run it on a `newsroom-<name>-app` repo and copy the file over.

If you can, have a quick go at it, otherwise I will create a ticket to cover this (and test the commands).

newsroom/companies/views.py
```diff
@@ -176,15 +175,16 @@ async def edit_company(args: CompanyItemArgs, params: None, request: Request) ->
...
     service = CompanyService()
     original = await service.find_by_id(args.company_id)
 
     if not original:
         raise NotFound(gettext("Company not found"))
 
     try:
-        # TODO-ASYNC: Convert to users async service
-        get_resource_service("users").delete_action(lookup={"company": args.company_id})
+        await UsersService().delete_many(lookup={"company": str(args.company_id)})
```
MarkLark86: It's probably due to the `Store ObjectId in MongoDB from async resource services` bug, but should this be an ObjectId when looking up which users to delete and not a string?

newsroom/companies/views.py
```diff
@@ -199,14 +199,22 @@ async def delete_company(args: CompanyItemArgs, params: None, request: Request)
 @company_endpoints.endpoint("/companies/<string:company_id>/users", methods=["GET"])
 @login_required
 async def company_users(args: CompanyItemArgs, params: None, request: Request) -> Response:
-    # TODO-ASYNC: Convert to users async service
-    users = [get_public_user_data(user) for user in query_resource("users", lookup={"company": args.company_id})]
-    return Response(users, 200, ())
+    from newsroom.users.service import UsersService
+
+    users_list = []
+    company_users = await UsersService().search(lookup={"company": str(args.company_id)})
```
MarkLark86: Same goes here

newsroom/companies/views.py
```diff
@@ -222,9 +230,10 @@ async def approve_company(args: CompanyItemArgs, params: None, request: Request)
     await service.update(original.id, updates)
 
     # Activate the Users of this Company
-    # TODO-ASYNC: Convert to users async service
-    users_service = get_resource_service("users")
-    for user in users_service.get(req=None, lookup={"company": original.id, "is_approved": {"$ne": True}}):
+    users_service = UsersService()
+    company_users = await users_service.search(lookup={"company": str(original.id)})
```
MarkLark86: And here. Thinking because we have this in a few locations, it might be good to have a common/util function that provides the list of users for a company

newsroom/companies/views.py
eos87: yeah, I'll adjust it now that you prepared the fix.

newsroom/companies/views.py
eos87: yep, agree. I'll a separate method to avoid repeating.

eos87: done.

newsroom/users/service.py
eos87: Fixed in commits: https://github.com/superdesk/newsroom-core/pull/1036/commits/d437a72dc5ab0378ecb89db8e61a853c2b947dbb https://github.com/superdesk/newsroom-core/pull/1036/commits/c54ddd987aba4988563035cb8ba016218087a64b

You would need to run this below to make it work properly. 
```python manage.py data_upgrade -i 00016_20240820-184145_users```

**Review by MarkLark86 — APPROVED** [review]
Looks like tests are failing.

Behave tests looks like it's trying to add users from Superdesk, as Pydantic validation is complaining about `needs_activation` field which is from Superdesk and not Newshub.

Not sure what's causing the E2E tests to fail

File: newsroom/users/views.py [hunk 7]
```diff
@@ -56,196 +75,241 @@ def get_settings_data():
...
+        if company:
+            auth_provider = get_company_auth_provider(company.model_dump(by_alias=True))
 
-        if auth_provider.features["verify_email"]:
-            add_token_data(new_user)
+            if auth_provider.features["verify_email"]:
+                add_token_data(new_user)
 
-        ids = get_resource_service("users").post([new_user])
+        ids = await UsersService().create([new_user])
```
MarkLark86: Sorry, I missed this one. Would need to wrap `service.create` and `service.update` in a try...except to catch and return validation errors to the browser.

See an example [here](https://github.com/superdesk/newsroom-core/blob/async/newsroom/companies/views.py#L84)

eos87: done.

MarkLark86 (approved):

PR #1043, opened 2024-08-20
Add a tooltip on hover

thecalcc:
STTNHUB-289
## Checklist

### DIFF
--- a/assets/search/components/SearchResultsBar/index.tsx
+++ b/assets/search/components/SearchResultsBar/index.tsx
@@ -24,6 +24,7 @@ import {Dropdown} from './../../../components/Dropdown';
 
 import {SearchResultTagsList} from './SearchResultTagsList';
 import {IDateFilter} from 'interfaces/common';
+import {ToolTip} from 'ui/components/ToolTip';
 
 interface ISortOption {
     label: string;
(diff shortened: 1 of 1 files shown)


assets/search/components/SearchResultsBar/index.tsx
```diff
@@ -235,19 +236,26 @@ class SearchResultsBarComponent extends React.Component<IProps, IState> {
...
-                                    data-test-id="toggle-search-bar"
-                                    onClick={this.toggleTagSection}
-                                    className="icon-button icon-button--tertiary icon-button--bordered"
-                                >
-                                    <i className={classNames(
-                                        'icon--arrow-right',
-                                        {
-                                            'icon--collapsible-open': isTagSectionShown,
-                                            'icon--collapsible-closed': !isTagSectionShown,
+                                <ToolTip key={`${isTagSectionShown}--state`} placement="left">
```
petrjasek, line 239: I don't get the key here, is it supposed to be `true-state` / `false-state` ?

thecalcc: yeah, just doing it because we have a dynamic `title` which has to be refreshed based on the state. Use it only to re-render the button, thus its title tag

petrjasek: I see, can you add a comment there maybe?

petrjasek (approved):

PR #1047, opened 2024-08-21
fix close modal after delete and error message issue in Oauth client [NHUB-528]

devketanpro:
## Checklist

### DIFF
--- a/assets/oauth_clients/components/Clients.tsx
+++ b/assets/oauth_clients/components/Clients.tsx
@@ -58,6 +58,7 @@ class Clients extends React.Component<any, any> {
 
         if (confirm(gettext('Would you like to delete client: {{name}}', {name: this.props.clientToEdit.name}))) {
             this.props.deleteClient('clients');
+            this.props.cancelEdit();
         }
     }
 
--- a/assets/oauth_clients/components/Clients.tsx
+++ b/assets/oauth_clients/components/Clients.tsx
@@ -58,7 +58,6 @@ class Clients extends React.Component<any, any> {
 
         if (confirm(gettext('Would you like to delete client: {{name}}', {name: this.props.clientToEdit.name}))) {
             this.props.deleteClient('clients');
-            this.props.cancelEdit();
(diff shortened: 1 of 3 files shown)


assets/oauth_clients/components/Clients.tsx
```diff
@@ -58,6 +58,7 @@ class Clients extends React.Component<any, any> {
 
         if (confirm(gettext('Would you like to delete client: {{name}}', {name: this.props.clientToEdit.name}))) {
             this.props.deleteClient('clients');
+            this.props.cancelEdit();
```
MarkLark86: The `async` branch should be reserved for backend changes only. We're trying to limit the frontend changes to none at all if possible.

This also looks like a common issue in the `develop` branch, so this fix might be better off going there (so it's in v2.8 release as well)

assets/utils.tsx
```diff
@@ -513,6 +513,8 @@ export function errorHandler(error: {errorData: any} | Response, dispatch?: any,
     if ('errorData' in error) {
         if (setError) {
             dispatch(setError(error.errorData));
+            const errorMessage = Object.values(error.errorData).join(', ');
```
MarkLark86: Why are you displaying a notification for the validation error? This isn't the way validation currently works. The error should be displayed in the form, directly underneath the input that caused the error.

Have a look at the Company form, if you try and create 2 companies with the same name

MarkLark86:
@devketanpro The backend code seems fine

assets/utils.tsx
devketanpro: yep, I checked. It seems to me that this is also core work. not related to async

assets/oauth_clients/components/Clients.tsx
devketanpro: would be good, to create a separate ticket for these issues, As I checked, these issues is also reproducing on nra-develop

MarkLark86 (approved):

PR #1053, opened 2024-08-26
Code cleanup & map open by default

thecalcc:
CPCN-642
## Checklist

### DIFF
--- a/assets/agenda/components/AgendaPreviewImage.tsx
+++ b/assets/agenda/components/AgendaPreviewImage.tsx
@@ -4,7 +4,6 @@ import {shouldRenderLocation, getLocations} from 'maps/utils';
 import StaticMap from 'maps/components/static';
 import BannerDrop from 'components/BannerDrop';
 import {gettext} from '../../utils';
-import {get} from 'lodash';
 
 /**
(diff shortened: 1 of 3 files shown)


assets/agenda/components/AgendaPreviewImage.tsx
```diff
@@ -34,4 +34,4 @@ export default function AgendaPreviewImage({item, onClick}: any) {
 AgendaPreviewImage.propTypes = {
```
MarkLark86, line 39: Would be good to also replace `propTypes` with the ts interface

assets/maps/components/static.tsx
```diff
@@ -1,12 +1,14 @@
...
 
+interface IProps {
+    scale?: number;
+    locations?: Array<any>;
+}
 
-class StaticMap extends React.Component<any, any> {
-    static propTypes: any;
-    constructor(props: any) {
+export default class StaticMap extends React.Component<IProps, any> {
```
MarkLark86: `IState` should be easy to add for this one too

MarkLark86 (approved):
Looks good, just a couple interface definitions could be added

PR #1057, opened 2024-08-29
[async] Fix search in users management

eos87:
I will open a separate PR to solve the problem with resending invites to users.

This PR depends on https://github.com/superdesk/superdesk-core/pull/2677

Thanks for checking!

### DIFF
--- a/newsroom/users/views.py
+++ b/newsroom/users/views.py
@@ -151,7 +151,7 @@ async def search(args: None, params: SearchArgs, request: Request) -> Response:
         lookup = {"_id": {"$in": params.ids}}
 
     if params.sort:
-        sort = str(params.sort)
+        sort = params.sort
 
     where_param = params.where
 
--- a/newsroom/users/views.py
+++ b/newsroom/users/views.py
@@ -12,7 +12,8 @@ from werkzeug.exceptions import BadRequest, NotFound
 from superdesk import get_resource_service
 from superdesk.core.web import Request, Response
 from superdesk.core import get_current_app, get_app_config
-from superdesk.core.resources.service import MongoResourceCursorAsync
+from superdesk.core.resources.cursor import SearchRequest
+from superdesk.core.resources.fields import ObjectId as ObjectIdField
 
 from newsroom.user_roles import UserRole
 from newsroom.auth import get_user_by_email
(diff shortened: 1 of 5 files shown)


newsroom/users/views.py
```diff
@@ -156,7 +151,7 @@ async def search(args: None, params: SearchArgs, request: Request) -> Response:
         lookup = {"_id": {"$in": params.ids}}
 
     if params.sort:
-        sort = params.sort
+        sort = str(params.sort)
```
MarkLark86: Conversion to string not required, it's either None or string already

MarkLark86 (approved):

MarkLark86:
@eos87 Just one small removal of conversion that's not needed. Once the superdesk-core PR is merged, please make sure the tests in the PR all pass

MarkLark86:
@eos87 I also noticed that the `on_delete` method on the [async users service](https://github.com/superdesk/newsroom-core/blob/async/newsroom/users/service.py#L85) is not awaiting `check_permission` and `super().on_delete`. Could you please fix those as well

newsroom/users/views.py
eos87: indeed.

eos87:
Hey @MarkLark86! Thanks for the review. 

There have been a few changes since you reviewed it: 
- @devketanpro found an issue with the `create_user` command. The problem was that the password was being set as plain text because it was using the old service. See here https://github.com/superdesk/newsroom-core/pull/1057/commits/cd67b3cbd1cfd4e30baf2081316d9195edbca2ed and here https://github.com/superdesk/newsroom-core/pull/1057/commits/9476d2c17a0ef95a6d6692080406ecd2789f0365. Updating the command to use the new async service has been a pulling-my-hair-out kind of one 😅. I've implemented this helper `run_async_to_sync` which can also be used in the celery tasks to make sure the async tasks are run until completion no matter if the loop is running or not. I didn't cover the helper with tests because I just found out that Django community provides this nice utilities package https://github.com/django/asgiref which already has something similar (probably more robust) [asgiref.sync.async_to_sync](https://docs.djangoproject.com/en/5.1/topics/async/#asgiref.sync.async_to_sync). So we can either keep the one I introduced here and improve it or, try the one from `asgiref.sync.async_to_sync`. We can talk more about it on Slack.
- Fixed the issue with the resent_invite found by Alice. I thought I was going to provide a separate PR for it but, it is a tiny fix so no need for separate one https://github.com/superdesk/newsroom-core/pull/1057/commits/0af2ca1de4c859e93df90f310dc9ba85bdad042b.

I will merge this PR as I would like to move on and complete this ticket once and for all. We can circle back to the run_async_to_sync helper in a separate PR/ticket.

PR #1058, opened 2024-08-30
[NHUB-535] Convert settings module to async

MarkLark86:
Depends on https://github.com/superdesk/superdesk-core/pull/2679

I have kept the settings resource synchronous, as it's mostly used from a cache. Changing it to async would require changes in the search resource services, which are not async (and requires a bigger effort to do so).

I don't see any benefit in making this async when most of the time we will access it via the Flask/Quart global


eos87 (approved):
Looks good. Tests are failing but I think it will be solved when merging the other PR into superdesk-core.

PR #1059, opened 2024-09-02
Create new async Topics & Folder resources, blueprints & helper functions [NHUB-531]

devketanpro:
## Checklist


newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class TopicService(NewshubAsyncResourceService[TopicResourceModel]):
+    resource_name = "topics"
+
+    async def on_create(self, docs: List[TopicResourceModel]) -> None:
+        await super().on_create(docs)
+        for doc in docs:
+            doc.original_creator = get_user_id()
+            doc.version_creator = get_user_id()
```
eos87: I think this is already covered in `NewshubAsyncResourceService`

newsroom/topics/folders_async.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    Company Based FolderResource Model
+    """
+
+    company: Annotated[ObjectIdField, validate_data_relation_async("companies")]
+
+
+class CompanyFoldersResourceService(NewshubAsyncResourceService[UserFoldersResourceModel]):
+    resource_name = "company_topic_folders"
+    pass
```
eos87: nit:
```suggestion
```

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,243 @@
...
+    async def on_delete(self, doc: TopicResourceModel):
+        await super().on_delete(doc)
+        # remove topic from users personal dashboards
+        # TODO-ASYNC: use async users resource here
+        users = superdesk.get_resource_service("users").get(req=None, lookup={"dashboards.topic_ids": doc.id})
+        for user in users:
+            updates = {"dashboards": user["dashboards"].copy()}
+            for dashboard in updates["dashboards"]:
+                dashboard["topic_ids"] = [topic_id for topic_id in dashboard["topic_ids"] if topic_id != doc.id]
+            superdesk.get_resource_service("users").system_update(user["_id"], updates, user)
```
eos87: Given that Users is already async and Topics will soon be with this PR, I'd suggest you go ahead and update this piece as I think it's unlikely that we come back to the already migrated modules later on (unless there are issues of course). Thoughts?

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    mongo_cursor = await TopicService().search(lookup=lookup)
+    topics = await mongo_cursor.to_list_raw()
+
+    # filter out the topics those belong to inactive users
+    return [t for t in topics if users.get(str(t["user"]))]
+
+
+async def auto_enable_user_emails(updates, original, user):
```
eos87: Would be good to have the typings in this method.

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,243 @@
...
+    if not updates.get("subscribers"):
+        return
+
+    # If current user is already subscribed to this topic,
+    # then no need to enable their email notifications
+    data = original.dict(by_alias=True, exclude_unset=True) if isinstance(original, TopicResourceModel) else original
```
eos87: `dict` is marked as deprecated.

```suggestion
    data = original.model_dump(by_alias=True, exclude_unset=True) if isinstance(original, TopicResourceModel) else original
```

newsroom/topics/views.py
```diff
@@ -1,83 +1,108 @@
...
 @login_required
-async def get_topics(_id):
+async def get_topics(args: RouteArguments, params: None, request: Request):
     """Returns list of followed topics of given user"""
-    if session["user"] != str(_id):
+    if session["user"] != str(args.user_id):
         abort(403)
-    return jsonify({"_items": _get_user_topics(_id)}), 200
+    topics = await _get_user_topics(args.user_id)
+    return Response({"_items": topics}, 200, ())
```
eos87:
```suggestion
    return success_response({"_items": topics})
```

most of these 200 status code responses have empty headers. I created this small helper function in case you want to reuse it :) 

https://github.com/superdesk/newsroom-core/blob/6d96d26cec92684d7879f91fb6c1ebe83eb409ef/newsroom/utils.py#L150-L160

eos87 (commented):
I left a couple of comments, please check it. Also the e2e tests are failing.

newsroom/topics/topics_async.py
devketanpro: yep right, I'll update it

newsroom/topics/topics_async.py
eos87: FYI, We'll have a `to_dict` method available when this https://github.com/superdesk/superdesk-core/pull/2693 is merged (soon).

newsroom/topics/views.py
eos87: Also here, once https://github.com/superdesk/superdesk-core/pull/2693 is merged, you can disregard this suggestion and simply use `return Response({"_items": topics})` as `status_code` and `headers` will have default value.

newsroom/topics/folders_async.py
MarkLark86: @devketanpro @eos87 Actually the `resource_name` is not required as it is populated by the core framework when it is registered. I manually add it in the tests in superdesk-core, as there are some tests where it's not actually registered.

So you can remove the `resource_name` and just leave `pass`

newsroom/topics/folders_async.py
```diff
@@ -0,0 +1,103 @@
...
+                name="unique_topic_folder_name",
+                keys=[("company", 1), ("user", 1), ("section", 1), ("parent", 1), ("name", 1)],
+                unique=True,
+                collation={"locale": "en", "strength": 2},
+            )
+        ],
+    ),
+)
+
+topic_folders_endpoints = EndpointGroup("topic_folders", __name__)
```
MarkLark86: Looks like all 3 EndpointGroup's are not used, `topic_folders_endpoint`, `user_topic_folders_endpoint` and `company_topic_folders_endpoints`. These can be removed

newsroom/topics/folders_async.py
```diff
@@ -0,0 +1,103 @@
...
+
+    async def on_deleted(self, doc):
+        await self.delete({"parent": doc["_id"]})
+        await TopicService().delete({"folder": doc["_id"]})
+
+    async def on_user_deleted(self, sender, user, **kwargs):
+        await self.delete({"user": user["_id"]})
+
+
+folder_resource_config = ResourceConfig(
```
MarkLark86: This ResourceConfig is not used anywhere, and can be removed. Though the mongo indexes config would need to be defined somewhere (maybe the user_resource_config would suffice, no need to be in both user & company configs)

MarkLark86:
@devketanpro The changes for source/parent resources are in a [PR](https://github.com/superdesk/superdesk-core/pull/2694) and ready to use.

In the `newsroom.topics.folders_async` module, you can modify both `user_topic_folders_resource_config` and `company_topic_folder_resource_config` to include the following:

```python
from superdesk.core.resources import ..., RestEndpointConfig, RestParentLink

user_topic_folders_resource_config = ResourceConfig(
    ...,
    datasource_name="topic_folders",
    rest_endpoints=RestEndpointConfig(
        parent_links=[RestParentLink(resource_name="users", model_id_field="user")]
    ),
)

company_topic_folder_resource_config = ResourceConfig(
    ...,
    datasource_name="topic_folders",
    rest_endpoints=RestEndpointConfig(
        parent_links=[RestParentLink(resource_name="companies", model_id_field="company")]
    )
)
```

These changes will provide the `/api/companies/<company>/topic_folders` and `/api/users/<user>/topic_folders` endpoints that the [front-end uses](https://github.com/superdesk/newsroom-core/blob/async/assets/user-profile/actions.ts#L250). See the [docs](https://superdesk--2694.org.readthedocs.build/en/2694/core/web.html#resource-rest-endpoint-with-parent) for an example.

Let me know how that goes, and if any changes are required in my PR

MarkLark86:
@devketanpro The core [PR](https://github.com/superdesk/superdesk-core/pull/2694) has now been merged. Let me know how you go with using this new functionality as suggested

devketanpro:
Hi @MarkLark86,

I made changes based on your PR suggestions, but I'm still encountering an issue. I'm getting a 404 when trying to fetch these endpoints:
```
/api/users/<string:user_id>/topic_folders
/api/companies/<string:company_id>/topic_folders
/api/users/<string:user_id>/topic_folders/<string:folders_id>
/api/companies/<string:company_id>/topic_folders/<string:folders_id>
/api/users/<string:id>/topics/<string:topic_id>
```
Any guidance on what I might be missing?

MarkLark86:
> Hi @MarkLark86,
> 
> I made changes based on your PR suggestions, but I'm still encountering an issue. I'm getting a 404 when trying to fetch these endpoints:
> ```
> /api/users/<string:user_id>/topic_folders
> /api/companies/<string:company_id>/topic_folders
> /api/users/<string:user_id>/topic_folders/<string:folders_id>
> /api/companies/<string:company_id>/topic_folders/<string:folders_id>
> /api/users/<string:id>/topics/<string:topic_id>
> ```
> Any guidance on what I might be missing?
> 

Looks like you're missing the `url` param to the rest endpoints config. See example test module https://github.com/superdesk/superdesk-core/blob/async/tests%2Fcore%2Fmodules%2Ftopics.py#L41

newsroom/topics/__init__.py
```diff
@@ -1,45 +1,13 @@
-import superdesk
-from superdesk.flask import Blueprint
+from superdesk.core.module import Module
 
-from .topics import get_user_topics  # noqa
-from . import folders, topics
+from .topics_async import topic_resource_config, topic_endpoints, init, get_user_topics  # noqa
```
MarkLark86: If you add a `__all__` attribute here, outlining an interface into this module, then you won't need the `#noqa`. [Example](https://github.com/superdesk/newsroom-core/blob/async/newsroom/section_filters/__init__.py)

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+    user_id: Annotated[ObjectIdField, validate_data_relation_async("users")]
+    notification_type: NotificationType = Field(default=NotificationType.REAL_TIME)
+
+
+class TopicResourceModel(NewshubResourceModel):
+    label: str
+    query: Optional[str] = None
+    filter: Optional[Dict[str, Any]] = None
+    created_filter: Annotated[Optional[Dict[str, Any]], Field(alias="created")] = None
+    user: Annotated[Optional[ObjectIdField], validate_data_relation_async("users")]
```
MarkLark86: If a field is optional you must provide a default, otherwise it you'll have to manually provide `None` when programmatically constructing and instance of this model

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+@unique
+class TopicType(str, Enum):
+    WIRE = "wire"
+    AGENDA = "agenda"
+
+
+@dataclass
+class TopicSubscriber:
+    user_id: Annotated[ObjectIdField, validate_data_relation_async("users")]
+    notification_type: NotificationType = Field(default=NotificationType.REAL_TIME)
```
MarkLark86: No need to use `Field ` to provide a default value, just use ` = NotificationType.REAL_TIME`

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class TopicResourceModel(NewshubResourceModel):
+    label: str
+    query: Optional[str] = None
+    filter: Optional[Dict[str, Any]] = None
+    created_filter: Annotated[Optional[Dict[str, Any]], Field(alias="created")] = None
+    user: Annotated[Optional[ObjectIdField], validate_data_relation_async("users")]
+    company: Annotated[Optional[ObjectIdField], validate_data_relation_async("companies")]
+    is_global: Optional[bool] = False
```
MarkLark86: Marking this field as `Optional` means it also accepts a `None` value. Just provide the default without marking it optional. i.e. `is_global: bool = False`

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+    subscribers: Optional[List[TopicSubscriber]] = []
+    timezone_offset: Optional[int] = None
+    topic_type: TopicType
+    navigation: Optional[List[Annotated[ObjectIdField, validate_data_relation_async("navigations")]]] = None
+    folder: Annotated[Optional[ObjectIdField], validate_data_relation_async("topic_folders")] = None
+    advanced: Optional[Dict[str, Any]] = None
+
+
+class TopicService(NewshubAsyncResourceService[TopicResourceModel]):
+    resource_name = "topics"
```
MarkLark86: No need to provide this attribute, as it is set on registration: https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/model.py#L327

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+
+    async def on_create(self, docs: List[TopicResourceModel]) -> None:
+        await super().on_create(docs)
+        for doc in docs:
+            doc.version_creator = get_user_id()
```
MarkLark86: This is probably better off added to the [core nhub service](https://github.com/superdesk/newsroom-core/blob/async/newsroom/core/resources/service.py#L19)

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+            if doc.folder:
+                doc.folder = ObjectId(doc.folder)
```
MarkLark86: This should no longer be required, as [Pydantic field](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L144) should do this for us

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+
+    async def on_update(self, updates: Dict[str, Any], original: TopicResourceModel) -> None:
+        await super().on_update(updates, original)
+        set_version_creator(updates)
```
MarkLark86: This is already populated, see https://github.com/superdesk/newsroom-core/blob/async/newsroom/core/resources/service.py#L23

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+        await super().on_updated(updates, original)
+        current_user = get_user()
+
+        if current_user:
+            await auto_enable_user_emails(updates, original, current_user)
+
+    async def on_delete(self, doc: TopicResourceModel):
+        await super().on_delete(doc)
+        # remove topic from users personal dashboards
+        # TODO-ASYNC: use async users resource here
```
MarkLark86: No need for this comment now, right?

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+            await UsersService().update(user.id, updates)
+
+    async def on_user_deleted(self, sender, user, **kwargs):
+        # delete user private topics
+        await self.delete({"is_global": False, "user": user["_id"]})
+
+        # remove user topic subscriptions from existing topics
+
+        mongo_cursor = await self.search(lookup={"subscribers.user_id": user["_id"]})
+        topics = await mongo_cursor.to_list_raw()
```
MarkLark86: It's better to use `async for` loop instead of retrieving the entire list of docs

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+                subscribers=[s for s in topic["subscribers"] if s["user_id"] != user_object_id],
+            )
+
+            if topic.get("user") == user_object_id:
+                topic["user"] = None
+
+            self.update(topic["_id"], updates)
+
+        # remove user as a topic creator for the rest
+        user_topics = await self.find_one(lookup={"user": user["_id"]})
```
MarkLark86: Is `find_one` what should be used here, or `search` to find all topics?

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+        async for topic in user_topics:
+            await self.update(topic.id, {"user": None})
+
+
+async def get_user_topics(user_id):
```
MarkLark86: Please provide types for function signature and return values

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+    user = dict(await UsersService().find_by_id(user_id))
```
MarkLark86: No need to convert this to a dict, can just use the ResourceModel instance directly. For example, `"user": user.id` or `"company": user.company`

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+    data = await TopicService().find(
+        SearchRequest(
```
MarkLark86: There's now no need to provide `SearchRequest` instance, you can just pass in the where filter.
i.e.
```python
await TopicService().find({
    "$or": [
        {"user": user["id"]},
        {"$and": [{"company": user.get("company")}, {"is_global": True}]},
    ]
})
```

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,242 @@
...
+    if not user_newly_subscribed:
+        return
+
+    # The current user subscribed to this topic in this update
+    # Enable their email notifications now
+    await UsersService().update(user["_id"], updates={"receive_email": True})
+
+
+async def init(app: SuperdeskAsyncApp):
+    user_deleted.connect(await TopicService().on_user_deleted)  # type: ignore
```
MarkLark86: This won't work as intended. We might need to wait for https://sofab.atlassian.net/browse/SDESK-7376. For now I think we can comment this one out

newsroom/topics/views.py
```diff
@@ -1,83 +1,105 @@
...
+            "created_filter": topic.pop("created", {}),
+            "is_global": topic.get("is_global", False),
+        }
+        topic.update(data)
 
     for subscriber in topic.get("subscribers") or []:
         subscriber["user_id"] = ObjectId(subscriber["user_id"])
 
-    ids = get_resource_service("topics").post([topic])
+    data = TopicResourceModel.model_validate(topic)
```
MarkLark86: No need to convert the dict to a ResourceModel instance now, you can pass dicts to `create` methods now

newsroom/topics/views.py
```diff
@@ -176,49 +189,51 @@ def get_topic_url(topic):
...
+async def share(args: RouteArguments, params: None, request: Request):
     current_user = get_user(required=True)
     data = await get_json_or_400()
     assert data.get("users")
     assert data.get("items")
     topic = get_entity_or_404(data.get("items")["_id"], "topics")
     for user_id in data["users"]:
-        user = get_resource_service("users").find_one(req=None, _id=user_id)
+        user_data = await UsersService().find_by_id(user_id)
+        user = user_data.dict(by_alias=True, exclude_unset=True)
```
MarkLark86: Use the new `to_dict` method we have, won't need to pass in `by_alias` or `exclude_unset` anymore then.

newsroom/topics_folders/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+)
+
+
+module = Module(
+    name="newsroom.topics_folders",
+    resources=[company_topic_folder_resource_config, user_topic_folders_resource_config],
+)
+
+
+async def get_user_folders(user, section):
```
MarkLark86: Please provide typings to function signature and return values

newsroom/topics_folders/folders.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class FolderResourceModel(NewshubResourceModel):
+    name: str
+    parent: Annotated[Optional[ObjectIdField], validate_data_relation_async("topic_folders")] = None
+    section: SectionType
+
+
+class FolderResourceService(NewshubAsyncResourceService[FolderResourceModel]):
+    resource_name = "topic_folders"
```
MarkLark86: No need to define `resource_name` on async services, this will be populated on registration

newsroom/users/module.py
```diff
@@ -19,7 +19,13 @@
                 keys=[("email", 1)],
                 unique=True,
                 collation={"locale": "en", "strength": 2},
-            )
+            ),
+            MongoIndexOptions(
```
MarkLark86: Was this added here by mistake?

newsroom/topics/folders.py
```diff
@@ -4,8 +4,6 @@
 from newsroom.user_roles import UserRole
```
MarkLark86: Is this module needed anymore? Can it be deleted

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,243 @@
...
+        users = await UsersService().search(lookup={"dashboards.topic_ids": doc.id})
+        async for user in users:
+            updates = {"dashboards": user.dashboards.copy()}
+            for dashboard in updates["dashboards"]:
+                dashboard["topic_ids"] = [topic_id for topic_id in dashboard["topic_ids"] if topic_id != doc.id]
+            await UsersService().update(user.id, updates)
+
+    async def on_user_deleted(self, sender, user, **kwargs):
+        # delete user private topics
+        await self.delete({"is_global": False, "user": user["_id"]})
```
MarkLark86: This does not match the expected signature of the `delete` method, as it expects a ResourceModel instance.
What you're after here is the newer [delete_many](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/service.py#L412) method on the service

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,243 @@
...
+
+            self.update(topic.id, updates)
+
+        # remove user as a topic creator for the rest
+        user_topics = await self.search(lookup={"user": user["_id"]})
+        async for topic in user_topics:
+            await self.update(topic.id, {"user": None})
+
+
+async def get_user_topics(user_id: Union[ObjectId, str, None]):
```
MarkLark86: Missing return type for this one

newsroom/topics_folders/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+)
+
+
+module = Module(
+    name="newsroom.topics_folders",
+    resources=[company_topic_folder_resource_config, user_topic_folders_resource_config, topic_folders_resource_config],
+)
+
+
+async def get_user_folders(user: User, section: str):
```
MarkLark86: Missing return type for these 2 functions as well

newsroom/topics_folders/folders.py
```diff
@@ -0,0 +1,116 @@
...
+
+class FolderResourceModel(NewshubResourceModel):
+    name: str
+    parent: Annotated[Optional[ObjectIdField], validate_data_relation_async("topic_folders")] = None
+    section: SectionType
+
+
+class FolderResourceService(NewshubAsyncResourceService[FolderResourceModel]):
+    async def on_deleted(self, doc):
+        await self.delete({"parent": doc["_id"]})
```
MarkLark86: These 3 `delete` methods should use the newer [delete_many](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/service.py#L412) method on the newer service.

e2e/cypress/e2e/wire/wire_topic.cy.js
```diff
@@ -72,7 +72,7 @@ describe('Wire - Topic', function () {
     it('Rename folder in My Wire Topics', () => {
         addResources([
             {
-                resource: 'topic_folders',
+                resource: 'user_topic_folders',
```
MarkLark86, line 75: Why can't `topic_folders` resource be used?

newsroom/core/resources/service.py
```diff
@@ -33,3 +34,40 @@ async def on_deleted(self, doc: NewshubResourceModelType):
         if self.clear_item_cache_on_update:
             app = get_current_wsgi_app()
             app.cache.delete(str(doc.id))
+
+    async def update_many(self, lookup: Dict[str, Any], updates: Dict[str, Any]) -> List[str]:
```
MarkLark86: This would not be needed anymore. Also this would be better placed in the superdesk-core as it's not Newshub specific

MarkLark86: I will take care of this addition

newsroom/topics/topics.py
```diff
@@ -65,6 +65,7 @@ class TopicsResource(newsroom.Resource):
     datasource = {"source": "topics", "default_sort": [("label", 1)]}
     allowed_roles = [role for role in UserRole]
     allowed_item_roles = allowed_roles
+    internal_resource = True
```
MarkLark86, line 70: Is the old topics resource required anymore?

devketanpro: yes, untill we not convert reports to async

MarkLark86: Can you place a comment at the top of this line, with prefix TODO-ASYNC, keeping around until reports are updated to async

tests/core/test_topics.py
```diff
@@ -404,33 +414,39 @@ async def test_remove_user_topics_on_user_delete(client, app):
...
+    assert 3 == len(topics)
 
-    folders, _ = app.data.find("user_topic_folders", req=None, lookup=None)
-    assert 2 == folders.count()
+    cursor = await UserFoldersResourceService().search(lookup={})
+    folders = await cursor.to_list_raw()
+    assert 2 == len(folders)
 
-    await client.delete(f"/users/{PUBLIC_USER_ID}")
+    # # TODO:- Test cases based on signal
```
MarkLark86: Can you use prefix `TODO-ASYNC`, that will allow us to know what tasks are left in the code by searching for this todo

newsroom/topics/topics.py
devketanpro: sure

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,252 @@
...
+
+            # Then construct new array with either subscriber found or empty list
+            updates["subscribers"] = [subscriber] if subscriber is not None else []
+
+        if updates.get("folder"):
+            updates["folder"] = ObjectId(updates["folder"])
+
+    async def on_updated(self, updates: Dict[str, Any], original: TopicResourceModel) -> None:
+        await super().on_updated(updates, original)
+        current_user = get_user()
```
eos87: We have a new `get_user_or_abort` function that is already async. We should replace the usages on `get_user` as it is sync and using old framework.

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,252 @@
...
+            for dashboard in updates["dashboards"]:
+                dashboard_dict = dashboard.to_dict()
+                # Remove the deleted topic id from topic_ids
+                dashboard_dict["topic_ids"] = [
+                    topic_id for topic_id in dashboard_dict["topic_ids"] if topic_id != doc.id
+                ]
+                updated_dashboards.append(dashboard_dict)
+
+            updates["dashboards"] = updated_dashboards
+            await UsersService().system_update(user.id, updates=updates)
```
eos87, line 109: I have the feeling that this logic belongs to UsersService instead. Could be another service method to update user dashboards. What do you think?

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,252 @@
...
+
+    async def on_user_deleted(self, sender, user, **kwargs):
```
eos87: I'd suggest to rename this method to something more self-explanatory. It's easy to understand that it will be triggered when deleting a user but not so clear of what it is doing. Alternatively, you could add some docstrings briefly describing what it does so it is more clear. Thoughts?

Update: also, let's please add the types to the params.

newsroom/topics/topics_async.py
```diff
@@ -0,0 +1,252 @@
...
+    topics = await mongo_cursor.to_list_raw()
+
+    # filter out the topics those belong to inactive users
+    return [t for t in topics if users.get(str(t["user"]))]
+
+
+async def auto_enable_user_emails(
+    updates: Union[Topic, Dict[str, Any]],
+    original: Union[TopicResourceModel, Dict[str, Any]],
+    user: Optional[Union[User, Dict[str, Any]]],  # Allow user to be None
```
eos87:
```suggestion
    updates: Topic | dict[str, Any],
    original: TopicResourceModel | dict[str, Any],
    user: Optional[User | dict[str, Any]],  # Allow user to be None
```

I briefly talked with @MarkLark86 about the types we should use in a PR review. He mentioned that he's sticking to built-in types as much as possible, then resort back into `typings` and other libs if they'd be needed. It makes sense to me. If you agree too, perhaps we start moving in the direction. Thoughts?

newsroom/topics/views.py
```diff
@@ -1,83 +1,103 @@
...
 
 
-@blueprint.route("/topics/my_topics", methods=["GET"])
+@topic_endpoints.endpoint("/topics/my_topics", methods=["GET"])
 @login_required
-async def get_list_my_topics():
-    return jsonify(_get_user_topics(get_user_id())), 200
+async def get_list_my_topics(args: RouteArguments, params: None, request: Request):
+    topics = await _get_user_topics(get_user_id())
+    return Response(topics, 200, ())
```
eos87: nit:
```suggestion
    return Response(topics)
```

newsroom/topics/views.py
```diff
@@ -87,75 +107,66 @@ async def update_topic(topic_id):
...
+async def delete(args: RouteArguments, params: None, request: Request):
     """Deletes a followed topic by given id"""
+    service = TopicService()
     current_user = get_user(required=True)
-    original = get_resource_service("topics").find_one(req=None, _id=ObjectId(topic_id))
+    original = await service.find_by_id(args.topic_id)
 
-    if not can_edit_topic(original, current_user):
+    if not await can_edit_topic(original, current_user):
         abort(403)
```
eos87: we should use the abort method from the `request` view param to avoid having much dependencies on quart/flask
```suggestion
        await request.abort(403)
```

eos87 (commented):
Hi @devketanpro, I left some comments. Please check when you have some time.

newsroom/topics/topics_async.py
devketanpro: I'm not entirely sure about moving this logic yet. @MarkLark86 , what do you think? Would this fit better in `UsersService` or is it fine as it is?

newsroom/topics/topics_async.py
devketanpro: makes sense to me

newsroom/topics/topics_async.py
MarkLark86: It should not be the Topic's responsibility to update a User when a Topic is deleted, this responsibility lies on the UserService itself. Considering we don't have an async event system yet, for a `on_topic_deleted`, maybe this can live here in Topics for now (as it is in the `develop` branch).

We can revisit this later

newsroom/topics/views.py
```diff
@@ -58,8 +59,11 @@ async def post_topic(args: RouteArguments, params: None, request: Request):
             "user": user_dict.get("_id"),
             "company": user_dict.get("company"),
             "_id": ObjectId(),
-            "created_filter": topic.pop("created", None),
             "is_global": topic.get("is_global", False),
+            # `_created` needs to be set otherwise there is a clash given `TopicResourceModel` and
```
MarkLark86, line 54: When constructing a model from a dict, you use the aliased name for fields.
In this case we have the [created](https://github.com/devketanpro/newsroom-core/blob/NHUB-531/newsroom/topics/topics_async.py#L48) field, which is used to provide a filter for [creation date](https://github.com/devketanpro/newsroom-core/blob/NHUB-531/newsroom/topics/views.py#L191).

So here, `_created` and `created` have 2 different uses, and only `created` should need to be populated as `_created` is populated by the async Newshub resource service

eos87: Yeah, that's what I initially thought but when setting `created` attribute only (for the creation date filter) it throws a validation error for the `created` field that holds the date time instead. So there is some sort of clash between aliases and members here. 

image

MarkLark86 (approved):

PR #1077, opened 2024-09-06
[NHUB-534] Implement section filters async

eos87:
### Purpose
Implement the section filters management with new Async framework

### What has changed
- Added new model for section filters, service class and corresponding module
- Updated views to use new endpoints group + reference to new service class

This PR depends on https://github.com/superdesk/superdesk-core/pull/2688

Resolves: [NHUB-534]

[NHUB-534]: https://sofab.atlassian.net/browse/NHUB-534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/section_filters/module.py
```diff
@@ -0,0 +1,45 @@
...
+    default_sort=[("name", 1)],
+)
+
+
+def init_module(app: NewsroomWebApp):
+    app.wsgi.settings_app(
+        "section-filters",
+        lazy_gettext("Section Filters"),
+        weight=450,
+        data=get_settings_data,
```
devketanpro, line 36: I wonder `get_setting_data` is an async function, so why not await this?

devketanpro (approved):
looks good to me.

newsroom/section_filters/module.py
eos87: Because it is awaited when rendering the settings app. Both ways are possible but not sure which one we should stick with. See here in line `36`

https://github.com/superdesk/newsroom-core/blob/ac0d6b6053fb93dbfbb99ef35bd76c26e3a295b0/newsroom/settings/views.py#L33-L39

devketanpro: oh right

dev-requirements.txt
```diff
@@ -471,7 +471,7 @@ six==1.16.0
     #   python-dateutil
 sniffio==1.3.1
     # via asgi-tools
-superdesk-core @ git+https://github.com/superdesk/superdesk-core.git@async
+superdesk-core @ git+https://github.com/superdesk/superdesk-core.git@hg/NHUB-534-cacheable-service
```
eos87: Updated this just to make sure tests pass. I'll fix it back to async once this PR https://github.com/superdesk/superdesk-core/pull/2688 gets merged.

newsroom/section_filters/model.py
```diff
@@ -0,0 +1,12 @@
+from typing import Optional
+from newsroom.core.resources.model import NewshubResourceModel
+
+
+class SectionFilter(NewshubResourceModel):
+    name: str
+    description: Optional[str]
```
MarkLark86: If a field is marked as `Optional`, you must provide a default value otherwise usages of this class would also need to provide a value for this field.

newsroom/section_filters/service.py
```diff
@@ -0,0 +1,50 @@
...
+from newsroom.core.resources.service import NewshubAsyncResourceService
+
+from .model import SectionFilter
+
+
+class SectionFiltersService(NewshubAsyncResourceService[SectionFilter], AsyncCacheableService):
+    resource_name = "section_filters"
+    cache_lookup = {"is_enabled": True}
+
+    async def get_section_filters(self, filter_type) -> list:
```
MarkLark86: Missing types for `filter_type: str` argument and sub-type for returned list, `list[dict]`

newsroom/section_filters/service.py
```diff
@@ -0,0 +1,50 @@
...
+        """Get the list of section filter by filter type
+
+        :param filter_type: Type of filter
+        """
+        section_filters = await self.get_section_filters_dict()
+        return section_filters.get(filter_type) or []
+
+    async def get_section_filters_dict(self) -> dict[str, list]:
```
MarkLark86: Missing sub-type in return type for `list`

newsroom/section_filters/service.py
```diff
@@ -0,0 +1,50 @@
...
+        if not getattr(g, "section_filters", None):
+            filters: dict[str, list] = {}
+            for f in await self.get_cached():
+                if not filters.get(f.get("filter_type")):
+                    filters[f.get("filter_type")] = []
+                filters[f.get("filter_type")].append(f)
+            g.section_filters = filters
+        return g.section_filters
+
+    async def apply_section_filter(self, query, product_type, filters=None):
```
MarkLark86: Missing type definitions

newsroom/section_filters/views.py
```diff
@@ -1,86 +1,123 @@
...
-        "sections": get_current_app().as_any().sections,
+        "section_filters": all_filters,
+        "sections": get_current_wsgi_app().sections,
     }
+
     return data
 
 
-@blueprint.route("/section_filters", methods=["GET"])
+def validate_section_filter(section_filter: dict) -> Response | None:
```
MarkLark86: This type of validation should now be covered by the Pydantic model validation

newsroom/section_filters/views.py
```diff
@@ -95,15 +132,17 @@ async def edit(id):
     if validation:
         return validation
 
-    set_version_creator(updates)
-    get_resource_service("section_filters").patch(id=ObjectId(id), updates=updates)
-    return jsonify({"success": True}), 200
+    await SectionFiltersService().update(args.id, updates)
```
MarkLark86, line 114: We should probably catch Pydantic validation issues and return them to the client here

newsroom/section_filters/views.py
```diff
@@ -95,15 +132,17 @@ async def edit(id):
...
 
+    return success_response({"success": True})
 
-@blueprint.route("/section_filters/<id>", methods=["DELETE"])
+
+@section_filters_endpoints.endpoint("/section_filters/<id>", methods=["DELETE"])
```
MarkLark86: Missing type for `id` in the url string

newsroom/utils.py
```diff
@@ -149,27 +152,59 @@ async def get_json_or_400_async(req: Request):
...
+def success_response(data: Any, headers: Sequence = (), status_code: int = 200):
     """
     Shortcut to return a `superdesk.core.web.Response` with a
     status_code 200 and empty headers by default
     """
     if isinstance(data, BaseModel):
         data = data.model_dump(by_alias=True, exclude_unset=True)
 
-    return Response(data, 200, headers)
+    return Response(data, status_code, headers)
```
MarkLark86, line 160: I'm wondering if instead we make the `headers` and `status_code` attributes for the `Response` class in superdesk-core optional?

Then we can simply use `return Response({"success": True})`, and in [superdesk-core -> app](https://github.com/superdesk/superdesk-core/blob/async/superdesk/factory/app.py#L259) we can test if `isinstance(response.body, BaseModel)` and support that in core (as we'll need that for Superdesk as well).

This will improve the core framework, making it easier to work with for all apps that will use it

newsroom/utils.py
```diff
@@ -149,27 +152,59 @@ async def get_json_or_400_async(req: Request):
...
 
 
-def response_from_validation(error: ValidationError, code=400):
+def parse_validation_error(error: ValidationError) -> dict[str, str]:
```
MarkLark86, line 163: Same thing here, we can maybe check `isinstance(response.body, ValidationError)` in [superdesk-core -> app](https://github.com/superdesk/superdesk-core/blob/async/superdesk/factory/app.py#L259). This way we can do the following using this functionality from the core framework:

```python
try:
    ids = await service.create([model_data])
    return Response({"id": ids[0]}, 201)
except ValidationError as err:
    return Response(err, 400)
```

With this supported as a core feature, we won't need `create_or_abort` either, as a Newshub only feature

newsroom/section_filters/model.py
```diff
@@ -0,0 +1,12 @@
...
+    sd_product_id: Optional[str] = None
+    query: Optional[str] = None
+    is_enabled: bool = True
+    filter_type: Optional[str] = "wire"
```
MarkLark86: Marking `filter_type` and `search_type` as `Optional` means they accept a value of `None`. These should only contain a string, i.e. non-nullable. As we're already providing a default string value here, the type for these two can simply be `str` without the `Optional`

newsroom/utils.py
eos87: Yeah, I agree with that. I'll make the PR for sd core shortly.

newsroom/utils.py
eos87: but that would still mean having this `try/except` many times with exactly the same logic. My intention was to avoid having that as it feels kinda useless when it is just to check validation errors. Even if we keep your suggestion in a helper/util function, you would still need to check if the returned value contains errors and then return the response within the view function. The benefit of raising the exception is that it works similarly to `abort` from flask. It stops the execution of the view function right away (as you would normally do if there are validation errors) and the errors are in the body of the exception response so they can be used from the frontend as we're already doing so. 

So instead of having this (in many places):

```python
try:
    new_model = Navigation.model_validate(creation_data)
    ids = await NavigationsService().create([new_model])
    # do something else with new id 
    return ids[0]
except ValidationError as err:
    return Response(err, 400)
```

we would just 
```python
new_entry_id = await create_or_abort(NavigationsService, Navigation, creation_data)
# do something else with new id 
```    

What if instead we add these helpers in core framework? like 

`create_or_abort`
`update_or_abort`
`delete_or_404`

I'm sure the logic in these will be repeated again and again otherwise. Thoughts?

newsroom/section_filters/service.py
eos87: do you mean like this? 
```suggestion
    async def get_section_filters_dict(self) -> dict[str, list[dict]]:
```

eos87: done

newsroom/utils.py
eos87: https://github.com/superdesk/superdesk-core/pull/2693

I'll create a separate PR for newsroom-core to update the usage of `model_dump` to `to_dict` and Response with default values where possible.

newsroom/section_filters/views.py
eos87: that's right

newsroom/utils.py
MarkLark86: You make a good point. An improvement to your suggestion, what if we use Flask/Quart's register_error_handler functionality like we're already doing for [SuperdeskApiErrors](https://github.com/superdesk/newsroom-core/blob/async/newsroom/factory/app.py#L210).

This way we can customise the response for Newshub and Superdesk (as I believe they use different formats for the validation errors to the client), and not need any request specific helpers (create, update and delete validation errors would be caught by the same error handler).

eos87: yes, this is great idea. I'll try that and update the PRs.

eos87: done.

newsroom/section_filters/views.py
```diff
@@ -1,86 +1,108 @@
...
     }
+
     return data
 
 
-@blueprint.route("/section_filters", methods=["GET"])
+async def get_section_filter_or_abort(id: str) -> SectionFilter:
+    section_filter = await SectionFiltersService().find_by_id(id)
+    if section_filter is None:
+        abort(404)
```
MarkLark86: It would be good not use use Flask/Quart functionality directly, but rather through the newer `Request` class

eos87: do you mean like passing down the request instance coming into the view function?

eos87: done.

MarkLark86 (approved):

PR #1082, opened 2024-09-12
[NHUB-554] Fix async RecaptchaField for SignupForm

MarkLark86:
### Purpose
When we migrated from `flask_wtf` to `quart_wtf`, we lost the Recaptcha field. This is because this field requires to get data from the request body, which is asynchronous in Quart.

### What has changed
Copy the Recaptcha field from [flask_wtf.recaptcha](https://github.com/wtforms/flask-wtf/tree/main/src/flask_wtf/recaptcha) into our source code, and make it work for our use case.

Resolves: [NHUB-554](https://sofab.atlassian.net/browse/NHUB-554)

[NHUB-554]: https://sofab.atlassian.net/browse/NHUB-554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -42,7 +42,7 @@ jobs:
           CYPRESS_SCREENSHOTS_FOLDER: /tmp/cypress
       - name: Upload screenshots
         if: ${{ failure() }}
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         with:
           name: screenshots-e2e
           path: /tmp/cypress/**/*.png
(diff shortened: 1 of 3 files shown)


MarkLark86:
@eos87 I tested this locally with my own Google reCAPTCHA and works as intended. We can get QA to test it once it's on our own infrastructure

eos87 (approved):
WTF approved 😅

PR #1083, opened 2024-09-12
[NHUB-533] Implement GlobalTopics async

eos87:
### Purpose
Implement the GlobalTopics management with new Async framework

### What has changed
- Added new model, service class and module registration
- Updated views to use new endpoints group + reference to new service class

This PR depends on https://github.com/superdesk/superdesk-core/pull/2688

Resolves: [NHUB-533]

[NHUB-533]: https://sofab.atlassian.net/browse/NHUB-533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarkLark86:
@eos87 Some WebAPI behave tests are failing. One of those is due to using [get_resource_service(resource).delete_action()](https://github.com/superdesk/superdesk-core/blob/async/superdesk/tests/steps.py#L503) from superdesk-core. This could be fixed by attempting to get the resource from the new async framework first, i.e.

```python
from superdesk.core import get_current_async_app

async def run_delete_action_for_resource(resource: str) -> None:
    app = get_current_async_app()
    try:
        await app.resources.get_resource_service(resource).delete_many({})
    except KeyError:
        get_resource_service(resource).delete_action()
```

You will then need to do something similar with [get_resource_service(resource).post(items)](https://github.com/superdesk/superdesk-core/blob/async/superdesk/tests/steps.py#L510) to attempt to use the async resource, falling back to legacy sync resource.

e2e/cypress/e2e/wire/topics.cy.js
```diff
@@ -98,10 +98,15 @@ describe('Wire - Topics', function () {
         WirePage.filterPanel.getCurrentPanel('[data-test-id="edit-btn"]').click();
         profileTopics.createNewFolder('Weather');
         profileTopics.createNewFolder('Traffic');
-        profileTopics.dragTopicToFolder('Sofab Weather', 'Weather');
-        profileTopics
-            .getTopicCardAction('Sofab Weather', 'Remove from folder')
-            .should('exist');
+
+        // TODO-ASYNC: The dragTopicToFolder action triggers a PATCH action over a nested resource.
```
MarkLark86, line 102: This should be resolved once @devketanpro has finished with his [Topic Folder PR](https://github.com/superdesk/newsroom-core/pull/1059)

e2e/server/newsroom_e2e/module.py
```diff
@@ -45,7 +70,13 @@ async def populate_resources(request: Request) -> Response:
             resource = entry.get("resource")
             items = entry.get("items") or []
 
-            if entry.get("use_resource_service", True):
+            if entry.get("use_db_collection", False):
```
MarkLark86: This might not be ideal. The idea behind E2E is to reproduce as close as possible a production environment with testing user interactions. Here we aren't using the resource service, which might have side effects to creating resources. Which would mean we won't need custom serialisation here but use the one the rest of the system uses.

Instead might I suggest to continue to use the `use_resource_service` option, and attempt to get the resource service with `app.resources.get_resource_service(resource)` from the async framework, falling back to the legacy sync resources.

newsroom/navigations/module.py
```diff
@@ -0,0 +1,38 @@
...
+from superdesk.core.module import Module
+from superdesk.core.resources import ResourceConfig, MongoResourceConfig
+
+from newsroom import MONGO_PREFIX
+
+from .model import Navigation
+from .views import get_settings_data, navigations_endpoints
+from .service import NavigationsService
+
+if TYPE_CHECKING:
```
MarkLark86, line 12: What is the reason to only include this when `TYPE_CHECKING` is enable?

newsroom/navigations/model.py
```diff
@@ -0,0 +1,11 @@
+from typing import Optional
+from newsroom.core.resources.model import NewshubResourceModel
+
+
+class Navigation(NewshubResourceModel):
+    name: str
```
MarkLark86: This value should be unique, using [validate_iunique_value_async](https://github.com/superdesk/newsroom-core/blob/async/newsroom/users/model.py#L40C45-L40C73)

newsroom/products/views.py
```diff
@@ -67,6 +69,13 @@ def validate_product(product):
         return jsonify({"name": gettext("Name not found")}), 400
 
 
+async def find_nav_or_404(id: str) -> Navigation:
```
MarkLark86: [id](https://docs.python.org/3/library/functions.html#id) is a built-in python function, so there is a name clash here. Would be good to use a different name instead

tests/core/utils.py
```diff
@@ -1,7 +1,40 @@
...
+
+async def insert_into(collection_name: str, data: list[Any]):
+    """
+    Insert multiple documents into a specified MongoDB collection asynchronously.
+
+    Args:
+        collection_name (str): The name of the MongoDB collection.
+        data (list[Any]): A list of documents to insert into the collection.
+    """
+    db_collection = get_db_collection(collection_name)
```
MarkLark86: Same with this one, it would be better to use the `app.resources.get_resource_service`. Not only does this apply side effects, serialisation, but also makes sure the items are created in Elastic (if enabled for the resource)

newsroom/agenda/views.py
```diff
@@ -15,7 +15,7 @@
 from newsroom.products.products import get_products_by_company
 from newsroom.template_filters import is_admin_or_internal, is_admin
 from newsroom.topics import get_company_folders, get_user_folders, get_user_topics
-from newsroom.navigations.navigations import get_navigations
+from newsroom.navigations.utils import get_navigations
```
MarkLark86: Would it be possible to add the following to the `__all__` attribute in `newsroom.navigations.__init__.py`:
- get_navigations
- get_navigation_by_company
- get_navigations_as_list

This way we can define what the `newsroom.navigations` module exposes, and not need to import deeper into that module

`from newsroom.navigations.utils import get_navigations`
would then become
`from newsroom.navigations import get_navigations`

newsroom/cards/module.py
```diff
@@ -26,11 +27,14 @@
 async def get_settings_data():
     app = get_current_wsgi_app()
 
+    cursor = await NavigationsService().search(lookup={"is_enabled": True})
```
MarkLark86: In situations like this, where we have multiple `await` for separate DB requests, we can probably start using `asyncio.gather` to run both requests concurrently and await it's results.

In this case, it would be:
```python
from asyncio import gather

async def get_settings_data():
    ...
    cursors = await gather(
        NavigationsService().search({"is_enabled": True}),
        CardsResourceService().search({}),
    )
    [navigations, cards] = await gather(
        cursors[0].to_list_raw(),
        cursors[1].to_list_raw(),
    )

    return {
        ...,
        "cards": cards,
        "navigations": navigations,
    }
```

newsroom/navigations/views.py
```diff
@@ -1,77 +1,123 @@
...
-    nav_data = await _get_navigation_data(data)
-    product_ids = nav_data.pop("products", [])
+async def create(request: Request):
+    nav_data = await get_nav_data_from_request(request)
+    service = NavigationsService()
 
-    set_original_creator(nav_data)
-    ids = get_resource_service("navigations").post([nav_data])
+    creation_data = await prepare_navigation_data(nav_data)
+    creation_data["id"] = service.generate_id()
```
MarkLark86, line 77: We could probably have this in the base async service create method. If supplied item is a dict and `_id` not defined, assign using the `self.generate_id()` method.

This way we dont' have to worry about it with these create item routes

newsroom/navigations/module.py
eos87: a circular reference ~but I will double check this because I have changed many pieces that this might not be needed~.

newsroom/navigations/views.py
eos87: yeah, good idea. I'll do that in a separate PR to superdesk-core and then another one for newsroom core.

newsroom/cards/module.py
eos87: good idea. I've made the changes.

e2e/server/newsroom_e2e/module.py
eos87: Improvement done, please check when you have some time.

newsroom/cards/module.py
```diff
@@ -26,11 +27,16 @@
 async def get_settings_data():
     app = get_current_wsgi_app()
 
+    cards_task = CardsResourceService().find({})
+    navs_task = NavigationsService().search(lookup={"is_enabled": True})
+
+    cards, navigations = await gather((await cards_task).to_list_raw(), (await navs_task).to_list_raw())
```
MarkLark86, line 33: Using `await` is a synchronous action, meaning this will await `cards_task` first, then await `navs_task` sequentially. To gain the benefit of concurrently awaiting both requests for a cursor, we will have to use something like:
 ```python
# Await the request for a cursor
[cards_cursor, navs_cursor] = await gather(cards_task, navs_task)
# Await for the cursor data
[cards, navigations] = await gather(cards_cursor.to_list_raw(), navs_cursor.to_list_raw())
```

This is because the initial request gets sent to MongoDB, which returns a cursor to the request. This is an async I/O request.

Then `to_list_raw` then requests MongoDB to return items from the cursor. This is another async I/O request.

So because we have 2 separate I/O requests to MongoDB, we will need to gather twice.

File: newsroom/core/resources/validators.py [hunk 11]
```diff
@@ -62,3 +64,29 @@ def _validate_dashboard_type_exists(value: str | None = None) -> str | None:
...
+def validate_multi_field_iunique_value_async(resource_name: str, field_names: list[str]) -> AsyncValidator:
+    """Validate that the combination of fields is unique in the resource (case-insensitive)
+
+    :param resource_name: The name of the resource where the combination of fields must be unique
+    :param field_names: The names of the fields to ensure unique combination
+    """
+
+    async def _validate_iunique_value(item: ResourceModel, value: Any) -> None:
+        fields = {field: getattr(item, field) for field in field_names}
+        if any(value for value in fields.values()):
```
MarkLark86: What is the purpose of this if/return statement?

MarkLark86 (commented):
Looks good, just a couple small things then would be good to merge

newsroom/cards/module.py
MarkLark86: Sorry, I got my information wrong here. Your changes here should suffice for mongo resources.

**Mongo:**
mongo.find: sync
cursor: async

**Elastic**
elastic.search: async
cursor: sync

newsroom/core/resources/validators.py
eos87: my bad, I changed this last moment, it should have been 
```python
if any(value is None for value in fields.values()):
```
I'll fix it now.

MarkLark86 (approved):

PR #1089, opened 2024-09-20
[NHUB-543] Implement notifications async

eos87:
### Purpose
Implement notifications and notification queue app into new async framework.

### What has changed
- Added new models, service classes and module registration
- Updated code references to use these new modules across the app
- Removed all services and resource classes

Resolves: [NHUB-543]

[NHUB-543]: https://sofab.atlassian.net/browse/NHUB-543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/notifications/models.py
```diff
@@ -0,0 +1,34 @@
...
+from newsroom.users.module import users_resource_config
+
+
+user_validated_type = Annotated[ObjectId, validate_data_relation_async(users_resource_config.name)]
+
+
+class Notification(ResourceModel):
+    item: str | ObjectId
+    # TODO-async: replace the one above with the one below once `items` are migrated to async framework
+    # item: Annotated[Optional[ObjectId], validate_data_relation_async("items")] = None
```
MarkLark86: It should be possible to use this to reference both async and legacy resources.
https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/validators.py#L142-L150

newsroom/notifications/models.py
```diff
@@ -0,0 +1,34 @@
...
+
+    user: user_validated_type
+    resource: Optional[str] = ""
```
MarkLark86: Looks like notifications are created only from the `save_user_notifications` function, which based on it's usage the `resource` and `action` fields are always populated. I'm wondering if we should make these two fields mandatory. Thoughts?

newsroom/notifications/models.py
```diff
@@ -0,0 +1,34 @@
...
+    action: Optional[str] = ""
+    data: Optional[dict] = {}
+
+
+@dataclass
+class Topic:
```
MarkLark86: There is already a class named Topic, maybe a more meaningful name here would describe what it is used for and be unique.

newsroom/notifications/models.py
```diff
@@ -0,0 +1,34 @@
...
+    topic_id: ObjectId
```
MarkLark86: Should this validate the Topic with ID exists?

newsroom/notifications/models.py
```diff
@@ -0,0 +1,34 @@
...
+    last_item_arrived: datetime
+    section: str
+    items: list[Annotated[ObjectId, validate_data_relation_async("items")]] = Field(default_factory=list)
```
MarkLark86: Nice use of `default_factory` to properly provide default value

newsroom/notifications/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    Returns the stories that user has notifications for
+    :return: List of stories. None if there is not user session.
+    """
+    if not session.get("user"):
+        return None
+
+    cursor = await NotificationsService().search(lookup=user_notifications_lookup(session["user"]))
+
+    return {
+        "user": str(session["user"]) if session["user"] else None,
```
MarkLark86: We probably don't need a second if condition for `session["user"]` when we're already providing one at the top of this function (I fear I might have been the one that added this... :face_with_head_bandage: )

newsroom/notifications/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    try:
+        items.extend(superdesk.get_resource_service("wire_search").get_items(item_ids))
+    except (KeyError, TypeError):  # wire disabled
+        pass
+    try:
+        items.extend(superdesk.get_resource_service("agenda").get_items(item_ids))
+    except (KeyError, TypeError):  # agenda disabled
+        pass
+    try:
+        items.extend(superdesk.get_resource_service("topics").get_items(item_ids))
```
MarkLark86, line 74: Could this use the new async resource service to get the list of topics?

newsroom/push.py
```diff
@@ -993,9 +989,7 @@ async def send_topic_notification_emails(item, topics, topic_matches, users, com
             if not user.get("receive_email"):
                 continue
             elif subscriber.get("notification_type") == "scheduled":
-                superdesk.get_resource_service("notification_queue").add_item_to_queue(
-                    user["_id"], section, topic["_id"], item
-                )
+                await NotificationQueueService().add_item_to_queue(user["_id"], section, topic["_id"], item)
```
MarkLark86: As this line is in nested for loops (over topics and subscribers), it might be good to get the service instance before entering these loops so it's constructor isn't called more than once unnecessarily.

newsroom/users/views.py
```diff
@@ -313,11 +313,11 @@ async def edit(args: RouteArguments, params: None, request: Request):
                         updates.pop(field, None)
 
             await UsersService().update(args.user_id, updates)
-            return success_response({"success": True})
+            return Response({"success": True})
 
         return Response(form.errors, 400)
 
-    return success_response(user)
+    return Response(user.to_dict())
```
MarkLark86: You can now provide a ResourceModel instance directly to Response constructor.
https://github.com/superdesk/superdesk-core/blob/async/superdesk/factory/app.py#L265-L266

MarkLark86 (commented):
Looks good, just a couple small things

newsroom/notifications/models.py
eos87: I had it like that but I changed it because of some broken tests. Then I wondered the same. I can change the tests for sure but I was not sure if that should be the right behaviour. I'll adjust it.

newsroom/notifications/utils.py
eos87: I'm not sure about this. Isn't that what @devketanpro is working on right now?

MarkLark86: Sorry my bad. I was looking at my local code, which I had pulled Ketan latest from. Disregard this

MarkLark86 (approved):

PR #1094, opened 2024-09-25
improve(async): Remove use of model_dump and generate_id

MarkLark86:
### Purpose
Remove use of `model_dump` in favour of `to_dict`, and remove use of `generate_id` in favour of the model itself providing a default.

Depends on https://github.com/superdesk/superdesk-core/pull/2716


eos87 (approved):
Looks good. I assume tests are not passing because the other PR needs to be merged first.

PR #1099, opened 2024-09-27
Upgrade history resource to async [NHUB-539]

devketanpro:
### Checklist

Resolves: #[issue-number]


newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+    extra_data: Optional[Dict] = None
+
+
+class HistoryService(NewshubAsyncResourceService[HistoryResourceModel]):
+    async def create(
+        self, docs: List[Dict[str, Any]], action: str, user: UserResourceModel, section: str = "wire", **kwargs: Any
+    ):
+        now = utcnow()
+
+        async def transform(item: Dict[str, Any]) -> Dict[str, Any]:
```
MarkLark86: There's no need for this function to be async, as there's no async I/O happening in the body of it

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+            return {
+                "_id": ObjectId(),
```
MarkLark86: We should no longer need to provide an ID ourselves, as this is covered [in the Model now](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/model.py#L242)

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+    user: Optional[Annotated[ObjectIdField, validate_data_relation_async("users")]] = None
+    company: Optional[Annotated[ObjectIdField, validate_data_relation_async("companies")]] = None
+    item: Optional[str] = None
+    version: Optional[str] = None
+    section: Optional[str] = None
+    extra_data: Optional[Dict] = None
+
+
+class HistoryService(NewshubAsyncResourceService[HistoryResourceModel]):
+    async def create(
```
MarkLark86: The signature of this function does not match the parent class [signature](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/service.py#L261).

By the looks of it, this is only used by `HistoryService.create_history_record` method. You might be able to simply rename this method from `create` to `create_history_record` (and use `self.create` instead of `super().create` and delete the other `create_history_record`.

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+from newsroom.users.model import UserResourceModel
+
+from superdesk.core.module import Module
+from quart_babel import gettext
+from superdesk.utc import utcnow
+from superdesk.core.web import EndpointGroup, Response, Request
+from superdesk.flask import abort
+from superdesk.core import json
+from superdesk.core.resources.fields import ObjectId as ObjectIdField
+from eve.utils import ParsedRequest
```
MarkLark86: Using `eve` is deprecated, and our intent is to completely replace it

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+        self, items: List[Dict[str, Any]], action: str, user: UserResourceModel, section: str
+    ):
+        await self.create(items, action, user, section)
+
+    async def query_items(self, query: Dict[str, Any]):
+        if query["from"] >= 1000:
+            # https://www.elastic.co/guide/en/elasticsearch/guide/current/pagination.html#pagination
+            return abort(400)
+
+        req = ParsedRequest()
```
MarkLark86: This is from Eve. What you can do instead is using the `find` directly from the service. `return await self.find(query)`

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+        req.args = {"source": json.dumps(query)}
+        return await super().get(req, None)
+
+    async def fetch_history(self, query: Dict[str, Any], all: bool = False):
+        results = await self.query_items(query)
+        docs = results.docs
+        if all:
```
MarkLark86, line 93: When changing `query_items` to use `self.find`, the response will be a [ElasticsearchResourceCursorAsync](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/cursor.py#L66) cursor.

Please update this method accordingly

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+            while results.hits["hits"]["total"]["value"] > len(docs):
+                query["from"] = len(docs)
+                results = await self.query_items(query)
+                docs.extend(results.docs)
+
+        return {"_items": docs, "hits": results.hits}
+
+
+async def get_history_users(
+    item_ids: List[Union[ObjectId, str]],
```
MarkLark86: We can directly use Python type system here.

So:
```python
List[Union[ObjectId, str]]
```
Would become
```python
list[ObjectId | str]
```

Which is easier to read and doesn't require to import from the typing module.
Note: This only works in Python 3.10+

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+                    {"term": {"action": action}},
+                ]
+            }
+        },
+        "size": 25,
+        "from": 0,
+    }
+
+    # Get the results
+    histories = await HistoryService().fetch_history(source, all=True)
```
MarkLark86: This could probably instead use `await HistoryService().find(source)` instead

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+)
+
+
+@history_endpoint.endpoint("/history/new", methods=["POST"])
+async def create(args: None, params: RouteParams, request: Request) -> Response:
+    params_dict = await get_json_or_400()
+    user = await get_user_or_abort()
+
+    if not params_dict.get("item") or not params_dict.get("action") or not params_dict.get("section"):
+        return "", gettext("Activity History: Inavlid request")
```
MarkLark86: Incorrect response type, please return an instance of `Response` when you can

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class RouteParams(BaseModel):
+    item: Optional[Dict[str, Any]] = None
+    action: Optional[str] = None
+    section: Optional[str] = None
+
+
+class HistoryResourceModel(NewshubResourceModel):
+    action: Optional[str] = None
```
MarkLark86: `action`, `item` and `section` are marked as `not_analyzed` in the original schema.

For the new Elastic resources, we would use Keyword field instead of just string.

i.e.
```python
from superdesk.core.resources.fields import Keyword

class HistoryResourceModel(NewshubResourceModel):
    action: Keyword | None = None
    item: Keyword | None = None
    section: Keyword | None = None
```

Also please look at the current implementation, and determine if fields are optional or not.
For example, it looks like item action and section may be required fields

newsroom/history_async.py
```diff
@@ -0,0 +1,178 @@
...
+    # Filter out the users
+    user_ids = [str(uid) for uid in active_user_ids]
+    return [str(h.user) for h in histories_items if h.user in user_ids]
+
+
+history_resource_config = ResourceConfig(
+    name="history",
+    data_class=HistoryResourceModel,
+    service=HistoryService,
+    mongo=MongoResourceConfig(
```
MarkLark86, line 151: You need to [enable Elasticsearch](https://github.com/superdesk/superdesk-core/blob/async/tests/core/modules/users/resources.py#L36) for this resource

devketanpro: yep

newsroom/history_async.py
```diff
@@ -0,0 +1,184 @@
...
+                name="item",
+                keys=[("item", 1)],
+            ),
+            MongoIndexOptions(
+                name="company_user",
+                keys=[("company", 1), ("user", 1)],
+            ),
+        ],
+    ),
+    elastic=ElasticResourceConfig(),
```
MarkLark86: You'll have to set a prefix here from `newsroom.ELASTIC_PREFIX` value. That will fix the index issues

newsroom/history_async.py
```diff
@@ -0,0 +1,185 @@
...
+                "versioncreated": now,
+                "user": user["_id"],
+                "company": user.get("company"),
+                "item": str(item["_id"]),
+                "version": str(item.get("version", item.get("_current_version"))),
+                "section": section,
+            }
+
+        transformed_docs = [transform(doc) for doc in docs]
+        print(transformed_docs)
```
MarkLark86: Please remove this `print` statement

newsroom/history_async.py
```diff
@@ -0,0 +1,185 @@
...
+        cursor = await self.query_items(query)
+
+        # Fetch the documents from the cursor
+        docs = []
+        async for doc in cursor:
+            docs.append(doc)
+
+        if all:
+            # Handle pagination and retrieve additional results
+            while cursor.hits["hits"]["total"]["value"] > len(docs):
```
MarkLark86: The `ElasticsearchResourceCursorAsync` cursor has a `count` function you can use here instead

tests/core/test_download.py
```diff
@@ -155,15 +157,14 @@ async def test_download_single(client, app):
...
 async def test_wire_download(client, app):
     await setup_image(client, app)
-    for _format in wire_formats:
-        _file = await download_zip_file(client, _format["format"], "wire")
-        with zipfile.ZipFile(_file) as zf:
-            assert _format["filename"] in zf.namelist()
-            content = zf.open(_format["filename"]).read()
-            if _format.get("test_content"):
-                _format["test_content"](content)
+    _file = await download_zip_file(client, wire_formats[0]["format"], "wire")
```
MarkLark86, line 160: What is the reason to change these tests, and not test against `wire_formats` or `agenda_formats`?

newsroom/history_async.py
```diff
@@ -0,0 +1,185 @@
...
+    user: Optional[Annotated[ObjectIdField, validate_data_relation_async("users")]] = None
+    company: Optional[Annotated[ObjectIdField, validate_data_relation_async("companies")]] = None
+    item: Keyword
+    version: str
+    section: Keyword
+    extra_data: Optional[Dict] = None
+
+
+class HistoryService(NewshubAsyncResourceService[HistoryResourceModel]):
+    async def create_history_record(
```
MarkLark86, line 48: The `user` arg here is expecting an instance of `UserResourceModel`, yet the uses of this function pass in a dictionary. There is also a case that a mocked user is passed in here, only providing just `_id` and `company` attributes.

What if instead we modify this signature to accept `user_id` and `company_id` explicitly rather than a user instance/dict? This would allow to use this function without requiring the entire user instance

newsroom/history_async.py
devketanpro: my bad

tests/core/test_download.py
devketanpro: actually, we encountered a duplication error for the item because, in the case of wire_formats or agenda_formats, the same item is being sent multiple times, which results in the error.

MarkLark86: Then I would ask, what have we changed in async, because this passed in develop v2.x branches/versions

devketanpro: in async we configured mongo indexes for tests

MarkLark86: Fair enough

MarkLark86 (approved):

PR #1103, opened 2024-10-02
improve(cli): Support async commands and wrap in app context

MarkLark86:
### Purpose
Replace `register_async_command` with standard `command` when registering commands. This enables automatic wrapping of app context and running async in current thread.

### What has changed
1. Use `command` instead of `register_async_command`
2. Convert commands to async functions, and add await where necessary (probably needs further looking into)

Depends on https://github.com/superdesk/superdesk-core/pull/2723


eos87:
Closing this as these changes will be addressed in another PR that I'll be opening later on.

PR #1104, opened 2024-10-02
fix(es): Set query_string default in separate config setting

MarkLark86:
### Purpose
The `ELASTICSEARCH_SETTINGS` config is used for the settings when creating indexes. We're storing `query_string.analyze_wildcard` on this config for use when constructing query_strings from the client, which is an invalid index setting. This causes an exception to be raised when attempting to create a new index.

### What has changed
Use a `ELASTICSEARCH_QUERY_STRING_DEFAULT_PARAMS` config specifically for use when constructing query_strings from clients. This avoid the conflict of uses between the two configs.

### DIFF
--- a/newsroom/search/service.py
+++ b/newsroom/search/service.py
@@ -50,21 +50,26 @@ def strtobool(val):
 def query_string(
     query: str,
     default_operator: Literal["AND", "OR"] = "AND",
-    fields: List[str] = ["*"],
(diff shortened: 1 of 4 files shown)


eos87 (approved):
Looks good

eos87:
Fixed the issue. Merging this.

PR #1105, opened 2024-10-03
[NHUB-546] Upgrade commands to use async resources

eos87:
### Purpose
Update all the Newshub commands to use async resources as much as possible and leave them in a functional state.

### What has changed
- Updated all commands to use the async resources that are already available
- Adjust commands to use new `superdesk-core` commands cli api 
- Some of the commands have been reworked to make them actually run as they were broken

Resolves: [NHUB-546]

[NHUB-546]: https://sofab.atlassian.net/browse/NHUB-546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/core/resources/service.py
```diff
@@ -34,3 +34,10 @@ async def on_deleted(self, doc: NewshubResourceModelType):
         if self.clear_item_cache_on_update:
             app = get_current_wsgi_app()
             app.cache.delete(str(doc.id))
+
+    async def find_items_by_ids(self, ids: list[str]) -> list[NewshubResourceModelType]:
```
MarkLark86, line 38: This would be best off being in the core AsyncResourceService, as this is not Newshub specific

newsroom/email.py
```diff
@@ -226,19 +231,23 @@ def get_language_template_name(template_name: str, language: str, extension: str
 TemplateKwargs = Dict[str, Any]
 
 
-# TODO-ASYNC: change this to use newsroom.users.model.UserResourceModel instead
+# TODO-ASYNC: change this to use newsroom.users.model.UserResourceModel only
```
MarkLark86, line 235: In my current task I am moving ALL resource models to `newsroom.types` module, so we can safely import types without circular dependency hell.

newsroom/email.py
```diff
@@ -226,19 +231,23 @@ def get_language_template_name(template_name: str, language: str, extension: str
...
-    user: User,
+    user: Union[User, "UserResourceModel"],
     template: str,
     template_kwargs: Optional[TemplateKwargs] = None,
     ignore_preferences=False,  # ignore user email preferences
     **kwargs: EmailKwargs,
 ) -> None:
     """Send an email to Newsroom user, respecting user's email preferences."""
+    if isinstance(user, BaseModel):
+        user = user.to_dict()
```
MarkLark86, line 245: `to_dict` is one of our methods and not from Pydantic, right? If that is the case, would be better to test for ResourceModel instead

MarkLark86 (approved):
Looks good, just a couple small comments and then you can merge (mainly moving the find by ids to core)

newsroom/email.py
eos87: Indeed. I initially used BaseModel because of circular reference but then I did workaround it with TYPE_CHECKING flag. Fixed now.

newsroom/core/resources/service.py
eos87: I'll add the method in separate PR in SD core and here too.

PR #1106, opened 2024-10-04
avoid dynamic mapping in agenda

petrjasek:
we had issue in production where inside location there were all possible translations from openstreetmaps and eventually we hit 1000 fields limit

### Checklist

Resolves: #[issue-number]


MarkLark86 (approved):
Will this also be needed in Superdesk?

petrjasek:
yep will make a PR there as well

PR #1113, opened 2024-10-07
[NHUB-548] Refactor `push.py` for better maintainability

eos87:
### Purpose
This refactor mainly breaks down the large push module into smaller, more focused components for improved readability, and maintainability having the separation of concerns in mind. This PR does NOT contain much changes related to async in order to make the review process simpler as there is some considerable noise because of the modules moved around. I tried to add as many types as possible while trying to not spend much time on it. Some method might not be complete but I will try to complete those in the subsequent PR.

### What has changed
- Added new module based on Core Async Framework
- Relocated views into `views.py` module
- Extracted generic util functions into `utils.py` module
- Extracted celery tasks into `tasks.py` module
- Updated endpoints to use new `EndpointGroup` from core async
- Added `AgendaManager`: This class is now solely responsible for handling agenda-related logic. All agenda-specific methods are grouped together, making it easier to read and maintain.
- Added `NotificationManager`: This class is dedicated to managing user notifications, including sending item notifications and handling topic matches
- Added `Publisher`: This class is responsible for the logic related to publishing events, items, and planning. It handles setting dates, validating event pushes, and updating the relevant services
- Removed old Blueprint 
- Adjusted tests to work accordingly with new async endpoints

### Pending 
- To update code to use async resources as much as possible 
- Complete some more types of the refactored methods and add docstrings
- Potentially add some more tests for the refactored functions - **_if needed_**

Belongs to: [NHUB-548]

[NHUB-548]: https://sofab.atlassian.net/browse/NHUB-548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1114, opened 2024-10-08
[NHUB-548] Upgrade push endpoints to use async

eos87:
### What has changed
- Update the `push` endpoints to use async resources as much as possible.
- Migrate the last method from old `sync` users service to the new `async` one
- Removed some of the references to old `sync` users service in tests and a couple of more modules 

Resolves: [NHUB-548] 

Note: For future references, this is the second PR for [NHUB-548]. First one was https://github.com/superdesk/newsroom-core/pull/1113 was about refactoring push module.

[NHUB-548]: https://sofab.atlassian.net/browse/NHUB-548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/notifications/utils.py
+++ b/newsroom/notifications/utils.py
@@ -1,9 +1,9 @@
+import superdesk
 import datetime
 
 from bson import ObjectId
-import superdesk
-
 from typing import Any
+
 from superdesk.utc import utcnow
 from superdesk.flask import session
 from superdesk.core import get_app_config
--- a/newsroom/notifications/utils.py
+++ b/newsroom/notifications/utils.py
@@ -1,9 +1,9 @@
-import superdesk
 import datetime
 
 from bson import ObjectId
 from typing import Any
 
+import superdesk
 from superdesk.utc import utcnow
 from superdesk.flask import session
 from superdesk.core import get_app_config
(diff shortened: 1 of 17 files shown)


newsroom/notifications/utils.py
```diff
@@ -1,9 +1,9 @@
+import superdesk
```
MarkLark86: Seems like this change isn't needed. It's also a good idea to import python inbuilt libs, 3rd part libs before our own ones.

newsroom/users/service.py
```diff
@@ -189,3 +191,29 @@ async def system_update(self, item_id, updates):
...
+            now = datetime.now(timezone).date()
+            pause_from_date = date.fromisoformat(pause_from)
+            pause_to_date = date.fromisoformat(pause_to)
+
+            if pause_from_date <= now <= pause_to_date:
+                return True
+
+        return False
+
+    async def find_by_email(self, email: str) -> UserResourceModel | None:
```
MarkLark86, line 212: Good addition, I have also done this with my current task as well

MarkLark86 (approved):
Looks good, just 1 suggestion added

PR #1122, opened 2024-10-16
[NHUB-547] Upgrade NewsAPI Atom feed to async

eos87:
### Purpose
To update `news_api.atom` and `news_api.assets` to use async resources as much as possible. `news_api.item` is still pending and should be addressed in a future PR once `formatters` are migrated to async. 

### What has changed
- Added two simple modules one for atom another for assets
- Adjusted some references, removed old apps from settings 

Resolves: [NHUB-547]

[NHUB-547]: https://sofab.atlassian.net/browse/NHUB-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/news_api/news/assets/__init__.py
```diff
@@ -0,0 +1,12 @@
+from superdesk.core.module import SuperdeskAsyncApp, Module
+from .views import assets_endpoints
+
+
+def init_module(app: SuperdeskAsyncApp):
+    app.wsgi.register_endpoint(assets_endpoints)
```
MarkLark86: Endpoints can also be registered directly on the Module instance. Would that not work here?

newsroom/news_api/news/atom/__init__.py
```diff
@@ -0,0 +1,12 @@
+from superdesk.core.module import SuperdeskAsyncApp, Module
+from .views import atom_endpoints
+
+
+def init_module(app: SuperdeskAsyncApp):
+    app.wsgi.register_endpoint(atom_endpoints)
```
MarkLark86: Same here, regarding registering on Module instance instead

MarkLark86 (approved):
Looks good. Just a couple questions/improvements (as to not require init_app function)

newsroom/news_api/news/assets/__init__.py
eos87: yeah, this is a leftover from when I was testing the registering of the resources in mongo and elastic directly. I will adjust it, thanks for spotting it.

PR #1123, opened 2024-10-17
[NHUB-538] - Upgrade FeaturedAgenda resource and search to async

BrianMwangi21:
### Purpose
This PR upgrades the FeaturedAgenda resource and search to async

### What has changed
- Upgraded `FeaturedResource` to new async resource service
- Update push and view endpoints to use new async resource for featured agenda
- Extracted `featured` and `get_featured_stories` to new async resource service

Resolves: [NHUB-538](https://sofab.atlassian.net/browse/NHUB-538)

[NHUB-538]: https://sofab.atlassian.net/browse/NHUB-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/agenda/model.py
```diff
@@ -0,0 +1,12 @@
+from datetime import datetime
+from typing import List, Optional
+
+from newsroom.core.resources.model import NewshubResourceModel
+
+
+class FeaturedResourceModel(NewshubResourceModel):
+    _id: str
```
MarkLark86: ID field isn't required to be manually defined, as it's defined in the parent ResourceModel

newsroom/agenda/model.py
```diff
@@ -0,0 +1,12 @@
+from datetime import datetime
+from typing import List, Optional
```
MarkLark86: In Python 3.10+ there's no longer the need for importing these types. Instead you can use the following syntax:

items: list[str] | None = None.

newsroom/agenda/service.py
```diff
@@ -0,0 +1,123 @@
+from datetime import datetime
+from eve.utils import ParsedRequest
+
+from newsroom.auth.utils import get_user_from_request, get_company_from_request
+from newsroom.agenda.agenda import (
+    is_events_only_access,
+    _agenda_query,
```
MarkLark86: As these functions are no longer private to that module, it would be good to remove the `_` prefix from the name

newsroom/agenda/service.py
```diff
@@ -0,0 +1,123 @@
...
+        user = get_user_from_request(None)
+        company = get_company_from_request(None)
+        if is_events_only_access(user.to_dict(), company.to_dict()):
+            abort(403)
+
+        if not featured or not featured.get("items"):
+            return ListCursor([])
+
+        query = _agenda_query()
+        get_resource_service("section_filters").apply_section_filter(query, self.section)
```
MarkLark86: Section filters are now available via the new Async service. You can now import it's service and directly use it here instead (the get_resource_service is the old way, and returns the deprecated sync resource)

newsroom/agenda/service.py
```diff
@@ -0,0 +1,123 @@
...
+        source["from"] = req.args.get("from", 0, type=int)
+        if not source["from"]:
+            source["aggs"] = aggregations
+
+        if company and not is_admin(user) and company.get("events_only", False):
+            # no adhoc planning items and remove planning items and coverages fields
+            query["bool"]["filter"].append({"exists": {"field": "event"}})
+            _remove_fields(source, PLANNING_ITEMS_FIELDS)
+
+        internal_req = ParsedRequest()
```
MarkLark86: We're aiming to completely remove eve. Instead directly use the underlying search function with the supplied elastic query

newsroom/agenda/views.py
```diff
@@ -161,7 +165,7 @@ async def get_view_data() -> Dict:
         "locators": get_vocabulary("locators"),
         "ui_config": await ui_config_service.get_section_config("agenda"),
         "groups": get_groups(get_app_config("AGENDA_GROUPS", []), company_dict),
-        "has_agenda_featured_items": get_resource_service("agenda_featured").find_one(req=None) is not None,
+        "has_agenda_featured_items": await FeaturedService().find_one(req=None) is not None,
```
MarkLark86: We now have a get all function on Async services, would be good to use that instead

newsroom/push/views.py
```diff
@@ -65,7 +66,7 @@ async def handle_publish_text_item(_, item):
 
 async def handle_publish_planning_featured(_, item):
     assert item.get("_id"), {"_id": 1}
-    service = get_resource_service("agenda_featured")
+    service = FeaturedService()
```
MarkLark86, line 69: Async resources have a find_by_id for use cases like this, so you don't have to construct a query to get by id

MarkLark86 (commented):
Looks mostly fine, just some fine tuning and improvements recommended

MarkLark86 (approved):

eos87:
Conflicts have been solved. I will merge this soon.

eos87:
@BrianMwangi21 @MarkLark86 I'll revert this PR in order to get `nra-async` instance back to fully work. At the moment `/agenda` and `/agenda/search` are throwing 500 error. I tried to look at it but there are multiple issues which would take much longer. 
- One of the issues I found is that `FeaturedResourceModel` should inherit from `ResourceModel` instead of `NewshubResourceModel` as feature items have string based ids and not ObjectId. 
- Another one is because of the return here https://github.com/superdesk/newsroom-core/pull/1123/files#diff-38690b7ac310f9b0408b2e5113521bd0ccfa99f4df433bc5c297c10f4a7ed0e3R128 does not seem to be valid as it returns a ListCursor. 
- Finally, and one that I could not figure out yet, is about the format of the provided date for featured items. Look at the format coming from the frontend here https://github.com/BrianMwangi21/newsroom-core/blob/NHUB-538/assets/agenda/actions.ts#L311 (`DD/MM/YYYY HH:mm`) and then in the backend https://github.com/BrianMwangi21/newsroom-core/blob/NHUB-538/newsroom/utils.py#L459 (`"%Y-%m-%dT%H:%M:%S"`). This throws an error because the formats don't match.

PR #1127, opened 2024-10-18
[NHUB-567] Migrate CompanyAdmin Blueprint to an EndpointGroup

eos87:
### What has changed
- Added new async framework module 
- Adjusted views functions according to new async framework

Resolves: [NHUB-567]

[NHUB-567]: https://sofab.atlassian.net/browse/NHUB-567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/company_admin/views.py
+++ b/newsroom/company_admin/views.py
@@ -1,29 +1,33 @@
 from typing import List
 from bson import ObjectId
 
-from werkzeug.exceptions import NotFound
 from quart_babel import gettext
 
-from superdesk.core import get_current_app, get_app_config
-from superdesk.flask import render_template, jsonify
+from superdesk.core.web import EndpointGroup
+from superdesk.core.types import Response, Request
+from superdesk.core import get_app_config
+from superdesk.flask import render_template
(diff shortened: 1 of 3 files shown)


newsroom/company_admin/views.py
```diff
@@ -51,28 +57,28 @@ def get_view_data():
...
     user = get_user_from_request(None)
     company = get_company_from_request(None)
 
-    if not company:
-        return NotFound(gettext("Company not found"))
+    assert company is not None
 
-    data = await get_json_or_400()
+    if not company:
+        await request.abort(404, gettext("Company not found"))
```
MarkLark86, line 66: Would this code block never be reached due to the above `assert` statement?

MarkLark86 (approved):

newsroom/company_admin/views.py
eos87: yeah, indeed. I moved it down as it is needed to avoid mypy complains.

PR #1131, opened 2024-10-21
fix(config): Duplicate key error with History resource

MarkLark86:
### Purpose
History Mongo index were configured to be unique, and we had 2 indexes when we can just use the 1.

### What has changed
* Remove the "item" index
* Place "item" first in the list of keys for "company_user" index
* Make "company_user" index not unique


eos87 (approved):
Looks good. Once tests pass I'll merge it to get async instance up to date

PR #1136, opened 2024-10-23
[NHUB-538] - Upgrade FeaturedAgenda resource and search to async

BrianMwangi21:
### Purpose
This PR upgrades the FeaturedAgenda resource and search to async

### What has changed
- Upgraded `FeaturedResource` to new async resource service
- Update push and view endpoints to use new async resource for featured agenda
- Extracted `featured` and `get_featured_stories` to new async resource service

Resolves: [NHUB-538](https://sofab.atlassian.net/browse/NHUB-538)

[NHUB-538]: https://sofab.atlassian.net/browse/NHUB-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/agenda/service.py
```diff
@@ -0,0 +1,118 @@
...
+        for item in docs:
+            date = datetime.strptime(item._id, "%Y%m%d")
+            item.display_from = local_to_utc(item.tz, date.replace(hour=0, minute=0, second=0))
+            item.display_to = local_to_utc(item.tz, date.replace(hour=23, minute=59, second=59))
+        await super().on_create(docs)
+
+    async def find_one_for_date(self, for_date: datetime) -> FeaturedResourceModel | None:
+        return await self.find_one(req=None, display_from={"$lte": for_date}, display_to={"$gte": for_date})
+
+    async def get_featured_stories(self, req, lookup):
```
eos87: can you add the type for the return value here?

newsroom/agenda/service.py
```diff
@@ -0,0 +1,118 @@
...
+        offset = int(req.args.get("timezone_offset", "0"))
+        local_date = get_local_date(
+            for_date.strftime("%Y-%m-%d"),
+            datetime.strftime(for_date, "%H:%M:%S"),
+            offset,
+        )
+        featured_doc = await self.find_one_for_date(local_date)
+        return await self.featured(req, lookup, featured_doc)
+
+    async def featured(self, req, lookup, featured):
```
eos87: Same here

newsroom/agenda/service.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        from newsroom.section_filters.service import SectionFiltersService
+
+        user = get_user_from_request(None)
+        company = get_company_from_request(None)
+        if is_events_only_access(user.to_dict(), company.to_dict()):
+            abort(403)
+
+        if not featured or not featured.get("items"):
+            return ListCursor([])
```
eos87: do we need to return a ListCursor? can we simply return a list?

newsroom/agenda/service.py
```diff
@@ -0,0 +1,118 @@
...
+
+        docs = []
+        agenda_ids = set()
+        for _id in featured["items"]:
+            if docs_by_id.get(_id) and docs_by_id.get(_id).get("_id") not in agenda_ids:
+                docs.append(docs_by_id.get(_id))
+                agenda_ids.add(docs_by_id.get(_id).get("_id"))
+
+        cursor.docs = docs
+        return cursor
```
eos87: If you loop through the items above (in the for loop) you wouldn't need to modify the cursor.docs and return the list of items instead.

newsroom/agenda/views.py
```diff
@@ -145,6 +151,11 @@ async def get_view_data() -> Dict:
     check_user_has_products(user, products)
     ui_config_service = UiConfigResourceService()
 
+    async def has_agenda_featured_items():
+        async for _ in FeaturedService().get_all():
+            return True
+        return False
```
eos87: I think it might be more efficient to use a count_documents instead. Maybe like this below:

```python
has_featured_items = (await FeaturedService().mongo_async.count_documents({}) > 0)
```

@MarkLark86 what do you think?

eos87 (commented):
I left a couple of minor comments. I haven't tested it locally but will do it tomorrow.

newsroom/agenda/model.py
```diff
@@ -0,0 +1,15 @@
+from datetime import datetime
+
+from superdesk.core.resources import ResourceModel, validators
+from typing_extensions import Annotated
+
+
+class FeaturedResourceModel(ResourceModel):
+    _id: Annotated[
```
MarkLark86: We have the [id](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/model.py#L69) field defined on the parent ResourceModel class, use that field here instead of `_id`. To make sure the field is a string, you can redefine it here (without the validation, as MongoDB should handle that for us already).
```python
id: Annotated[str, Field(alias="_id")]
```

On a side note, fields starting with an `_` are reserved for private fields. To get around this, we use Pydantic's alias feature, as shown above.

newsroom/agenda/model.py
```diff
@@ -0,0 +1,15 @@
+from datetime import datetime
```
MarkLark86: We have now moved all resource model definitions into the [newsroom.types](https://github.com/superdesk/newsroom-core/tree/async/newsroom/types) module, please move this file there under the appropriate file name

newsroom/agenda/__init__.py
```diff
@@ -5,7 +5,6 @@
 from newsroom.utils import url_for_agenda
 from .agenda import AgendaResource, AgendaService, aggregations, PRIVATE_FIELDS
 from newsroom.search.config import init_nested_aggregation
-from .featured import FeaturedResource, FeaturedService
```
MarkLark86, line 8: Is `newsroom.agenda.featured` module use anywhere in the code now, if not it would be good to remove that file

newsroom/agenda/views.py
MarkLark86: I've created a [PR](https://github.com/superdesk/superdesk-core/pull/2739) to add a count function on async resource services, that works for both MongoDB and Elasticsearch.
@eos87 Could you please review that, then @BrianMwangi21 can use it here instead

newsroom/agenda/module.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+featured_resource_config = ResourceConfig(
+    name="agenda_featured",
+    data_class=FeaturedResourceModel,
+    service=FeaturedService,
+    mongo=MongoResourceConfig(prefix=MONGO_PREFIX),
+)
+
+module = Module(name="newsroom.agenda_featured", resources=[featured_resource_config])
```
MarkLark86: As this module will be for normal Agenda items (Events & Planning items) and FeaturedAgenda items, I think the name of the module should just be `newsroom.agenda`

newsroom/agenda/service.py
```diff
@@ -0,0 +1,118 @@
...
+from superdesk.flask import abort
+from superdesk.core.resources import AsyncResourceService
+from superdesk.utc import local_to_utc
+from superdesk.utils import ListCursor
+
+
+class FeaturedService(AsyncResourceService[FeaturedResourceModel]):
+    resource_name = "agenda_featured"
+
+    async def on_create(self, docs):
```
MarkLark86: It would be good to add types here, `docs: list[FeaturedResourceModel]) -> None:`, so that the mypy tool can test types here

newsroom/agenda/service.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        Add UTC from/to datetimes on save.
+        Problem is 31.8. in Sydney is from 30.8. 14:00 UTC to 31.8. 13:59 UTC.
+        And because we query later using UTC, we store those UTC datetimes as
+        display_from and display_to.
+        """
+        for item in docs:
+            date = datetime.strptime(item._id, "%Y%m%d")
```
MarkLark86: When removing the `_id` field you defined on the resource, you would then use the standard `item.id` field instead

newsroom/agenda/service.py
```diff
@@ -0,0 +1,118 @@
...
+            item.display_from = local_to_utc(item.tz, date.replace(hour=0, minute=0, second=0))
+            item.display_to = local_to_utc(item.tz, date.replace(hour=23, minute=59, second=59))
+        await super().on_create(docs)
+
+    async def find_one_for_date(self, for_date: datetime) -> FeaturedResourceModel | None:
+        return await self.find_one(req=None, display_from={"$lte": for_date}, display_to={"$gte": for_date})
```
MarkLark86: In the new AsyncResourceService, `req` is no longer required (nor in the function signature). Instead just provide the keyword arguments.
```python
return await self.find_one(display_from={"$lte": for_date}, display_to={"$gte": for_date})
```

newsroom/agenda/service.py
MarkLark86: Instead of `req, lookup` (which is the old way of doing things), you can directly provide the arguments you require here, and then use `date_from` and `timezone_offset` directly. Such as:
```python
async def get_featured_stories(self, date_from: str, timezone_offset: int = 0) -> Any:
```

I wouldn't worry too much about the return type, this will change anyhow when the Agenda resource itself is upgraded, so leaving it as `Any` should suffice for now

newsroom/agenda/service.py
MarkLark86: Same here, figure out what the improved function signature should be, providing only the arguments you require to get the job done.

MarkLark86: Actually there is an issue here, this function requires the `req` to be of Eve ParsedRequest type, and not our own Request instance. This might need a re-think (or wait until the Agenda resource is updated...)

newsroom/agenda/service.py
```diff
@@ -0,0 +1,118 @@
...
+            name="featured",
+        )
+        if req.args.get("q"):
+            query["bool"]["filter"].append(self.query_string(req.args["q"]))
+            planning_items_query["nested"]["query"]["bool"]["filter"].append(planning_items_query_string(req.args["q"]))
+
+        query["bool"]["filter"].append(planning_items_query)
+
+        source = {"query": query}
+        self.set_post_filter(source, req)
```
MarkLark86: Similar issue here, req is expected to be a ParsedRequest instance, and we're calling a function that doesn't exist on this class. Again, might need to wait for the Agenda resource to be upgraded to async search

MarkLark86 (commented):
@BrianMwangi21 @eos87 I'm sorry to say, but I think this task now depends on the Agenda asycn upgrade task. As the `async def featured` function expects a request instance from Eve and not our new Request instance. :-(

@BrianMwangi21 Can you please make the recommended changes in other parts of this PR, and then we can mark this PR as draft while we wait for the Agenda task to be completed (otherwise I fear doubling of effort here, changing it now just to change it again next week).

BrianMwangi21:
> @BrianMwangi21 @eos87 I'm sorry to say, but I think this task now depends on the Agenda asycn upgrade task. As the `async def featured` function expects a request instance from Eve and not our new Request instance. :-(
> 
> @BrianMwangi21 Can you please make the recommended changes in other parts of this PR, and then we can mark this PR as draft while we wait for the Agenda task to be completed (otherwise I fear doubling of effort here, changing it now just to change it again next week).

That's very ok @MarkLark86. Understood.

newsroom/agenda/views.py
eos87: Approved and merged already. @BrianMwangi21 you can pull the latest from superdesk-core and use the new method added by Mark. Thanks guys!

newsroom/agenda/service.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+        if query_string:
+            query["bool"]["filter"].append(self.query_string(query_string))
+            planning_items_query["nested"]["query"]["bool"]["filter"].append(planning_items_query_string(query_string))
+
+        query["bool"]["filter"].append(planning_items_query)
+
+        source = {"query": query, "size": len(featured_doc["items"]), "from": from_offset}
+        # self.set_post_filter(source, req)  # TODO: Confirm usage of function with previous ParsedReq
```
BrianMwangi21: @MarkLark86 could you reconfirm about this. The function is in the agenda file and uses ParsedReq. To wait for the Agenda class to be async or create a similar function for this service ?

MarkLark86: For now you could just use:
```python
# Top of the file import

async def featured(
    ....
    get_resource_service("agenda").set_post_filter(source, req)
```

newsroom/agenda/service.py
```diff
@@ -0,0 +1,120 @@
...
+        if not from_offset:
+            source["aggs"] = aggregations
+
+        if company and not is_admin(user) and company.get("events_only", False):
```
MarkLark86: `company` here is an instance of `CompanyResource`, and therefor you can just use `company.events_only`

newsroom/agenda/service.py
```diff
@@ -0,0 +1,135 @@
+from datetime import datetime
+from eve.utils import ParsedRequest
+from typing import Optional, Any, List
```
eos87:
```
from typing import Optional, Any, List
```

we're trying to stick as much as possible to python built-ins, so we should avoid using Optional and List from typing and instead using `| None` and `list`

newsroom/agenda/service.py
```diff
@@ -0,0 +1,135 @@
...
+from superdesk import get_resource_service
+from superdesk.flask import abort
+from superdesk.core.resources import AsyncResourceService
+from superdesk.utc import local_to_utc
+
+
+class FeaturedService(AsyncResourceService[FeaturedResourceModel]):
+    resource_name = "agenda_featured"
+
+    async def on_create(self, docs: List[FeaturedResourceModel]) -> None:
```
eos87: here for instance
```suggestion
    async def on_create(self, docs: list[FeaturedResourceModel]) -> None:
```

newsroom/agenda/service.py
```diff
@@ -0,0 +1,135 @@
...
+
+    async def find_one_for_date(self, for_date: datetime) -> FeaturedResourceModel | None:
+        return await self.find_one(display_from={"$lte": for_date}, display_to={"$gte": for_date})
+
+    async def get_featured_stories(
+        self,
+        date_from: str,
+        timezone_offset: int = 0,
+        query_string: Optional[str] = None,
+        filter_string: Optional[str] = None,
```
eos87: here too and so on
```suggestion
        query_string: str | None = None,
        filter_string: str | None = None,
```

newsroom/agenda/service.py
```diff
@@ -0,0 +1,135 @@
...
+        docs = []
+        agenda_ids = set()
+        for _id in featured_doc["items"]:
+            if docs_by_id.get(_id) and docs_by_id.get(_id).get("_id") not in agenda_ids:  # type: ignore
+                docs.append(docs_by_id.get(_id))
+                agenda_ids.add(docs_by_id.get(_id).get("_id"))  # type: ignore
+
+        # TODO-ASYNC: Temporary implementation to ensure the response includes _meta details
+        # for total count as the front end relies on data._meta.total
+        return ({"_items": docs, "_meta": {"total": len(docs)}},)
```
eos87, line 136: @MarkLark86 I talked with @BrianMwangi21 about this and in order to not spend much more time in this task, I suggested this approach. Given the whole agenda task is not yet to be completed, we can tackle this one in the next one and move on to QA with it.

eos87 (approved):
Looks good. Just left some minor comments

newsroom/agenda/service.py
MarkLark86: Yep, looks good

MarkLark86 (approved):

PR #1138, opened 2024-10-24
[NHUB-565] Upgrade Products resource and endpoints to async

eos87:
### Purpose
Migrate products into new async framework

### What has changed
- Added new resource model and service for products
- Implemented views using new async framework endpoints 
- Added a custom validator to parse and validate proper IDs values 
- Removed references to old resource and service as much as possible

### Pending
- Remove references to old products's model and service from search services 
- Move a couple of helper functions from `products.py` into new `utils.py` and make sure they're all async

> [!IMPORTANT]
> We have a flaky test (`test_delete_all_notifications`) that I will try to fix in a separate PR, probably early next week. 

Thanks for checking!

Resolves: [NHUB-565]

[NHUB-565]: https://sofab.atlassian.net/browse/NHUB-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/companies/companies_async/service.py
+++ b/newsroom/companies/companies_async/service.py
@@ -5,6 +5,7 @@ from newsroom.types import CompanyResource
 from newsroom.core.resources import NewshubAsyncResourceService
 from newsroom.signals import company_create
 from newsroom.auth.utils import is_from_request, get_current_request, get_company_from_request
+from newsroom.types.company import CompanyProduct
 
 from ..utils import get_company_section_names, get_company_product_ids, get_users_by_company
 
@@ -18,13 +19,24 @@ class CompanyService(NewshubAsyncResourceService[CompanyResource]):
         for company in docs:
             company_create.send(self, company=company.to_dict())
 
+    def _get_products(self, updates: Dict[str, Any], original: CompanyResource):
+        """
+        Loop over products and checks if any of them is instance of CompanyProduct to
+        convert them into dict
+        """
+        for product in updates.get("products", original.products) or []:
+            if isinstance(product, CompanyProduct):
+                yield product.to_dict()
+            else:
+                yield product
+
     async def on_update(self, updates: Dict[str, Any], original: CompanyResource) -> None:
         await super().on_update(updates, original)
         if "sections" in updates or "products" in updates:
             sections = updates.get("sections", original.sections) or {}
             updates["products"] = [
                 product
-                for product in updates.get("products", original.products) or []
+                for product in self._get_products(updates, original)
                 if product.get("section") and sections.get(product["section"]) is True
             ]
 
--- a/newsroom/companies/companies_async/service.py
+++ b/newsroom/companies/companies_async/service.py
@@ -1,4 +1,4 @@
-from typing import List, Dict, Any
+from typing import Iterable, List, Dict, Any
 from copy import deepcopy
 
 from newsroom.types import CompanyResource
@@ -19,7 +19,7 @@ class CompanyService(NewshubAsyncResourceService[CompanyResource]):
         for company in docs:
             company_create.send(self, company=company.to_dict())
 
-    def _get_products(self, updates: Dict[str, Any], original: CompanyResource):
+    def _get_products(self, updates: Dict[str, Any], original: CompanyResource) -> Iterable[dict[str, Any]]:
         """
         Loop over products and checks if any of them is instance of CompanyProduct to
         convert them into dict
--- a/newsroom/core/resources/service.py
+++ b/newsroom/core/resources/service.py
@@ -49,3 +49,9 @@ class NewshubAsyncResourceService(AsyncResourceService[Generic[NewshubResourceMo
         """
         cursor = await self.search({"_id": {"$in": ids}})
         return await cursor.to_list()
+
+    async def get_all_raw_as_list(self):
+        """
+        Returns the list of all the entries raw as list
+        """
+        return [entry async for entry in self.get_all_raw()]
--- a/newsroom/core/resources/service.py
+++ b/newsroom/core/resources/service.py
@@ -50,7 +50,7 @@ class NewshubAsyncResourceService(AsyncResourceService[Generic[NewshubResourceMo
         cursor = await self.search({"_id": {"$in": ids}})
         return await cursor.to_list()
 
-    async def get_all_raw_as_list(self):
+    async def get_all_raw_as_list(self) -> list[dict[str, Any]]:
         """
         Returns the list of all the entries raw as list
         """
--- a/newsroom/core/resources/validators.py
+++ b/newsroom/core/resources/validators.py
@@ -1,7 +1,8 @@
 import re
 import ipaddress
 from typing import Any, Optional
-from pydantic import AfterValidator
+from bson import ObjectId, errors
+from pydantic import AfterValidator, BeforeValidator
 from pydantic_core import PydanticCustomError
 from quart_babel import gettext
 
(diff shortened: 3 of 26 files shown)


newsroom/companies/companies_async/service.py
```diff
@@ -18,13 +19,24 @@ async def on_create(self, docs: List[CompanyResource]) -> None:
         for company in docs:
             company_create.send(self, company=company.to_dict())
 
+    def _get_products(self, updates: Dict[str, Any], original: CompanyResource):
```
devketanpro: would be good to add a return type to the function signature

newsroom/core/resources/service.py
```diff
@@ -49,3 +49,9 @@ async def find_items_by_ids(self, ids: list[str]) -> list[NewshubResourceModelTy
         """
         cursor = await self.search({"_id": {"$in": ids}})
         return await cursor.to_list()
+
+    async def get_all_raw_as_list(self):
```
devketanpro: also here missing return type

newsroom/core/resources/validators.py
```diff
@@ -1,7 +1,8 @@
 import re
 import ipaddress
 from typing import Any, Optional
-from pydantic import AfterValidator
+from bson import ObjectId, errors
```
devketanpro: would we good to import only `ObjectId` and `InvalidId` from `bson` to avoid unnecessary imports and improve readability.

thoughts ?

newsroom/products/views.py
```diff
@@ -109,64 +99,114 @@ async def edit(id):
...
+        A tuple with the list of products of the company and a boolean to indicate if an update is
+        required or not.
+    """
+    company_products = [p for p in company.products if str(p._id) != str(product_id)]
+    is_update_required = len(company_products) != len(company.products)
+
+    return cast(list[ProductRef], company_products), is_update_required
+
+
+def update_company_sections(company: CompanyResource, company_products: list[ProductRef]):
```
devketanpro: missing return types here

newsroom/products/views.py
```diff
@@ -44,61 +39,56 @@ def get_product_ref(product: Product, seats=0) -> ProductRef:
     }
 
 
-@blueprint.route("/products", methods=["GET"])
-@admin_only
+class SearchParams(BaseModel):
+    q: str | None = None
```
devketanpro, line 43: It would be good to consider updating the variable name `q` to something more descriptive, such as `query`. thoughts?

devketanpro:
Hey @eos87, looks good overall! Just a few minor improvement suggestions in the comments

newsroom/products/views.py
eos87: I think this parameter is coming as `q` from the frontend so I wouldn't change it for now to reduce the code changes to the frontend.

devketanpro (approved):

eos87:
@MarkLark86 I'll merge this in order to try to get it tested. We can address any comment that you would have in a separate PR.

newsroom/core/resources/validators.py
```diff
@@ -116,3 +118,39 @@ async def _validate_iunique_value(item: ResourceModel, value: Any) -> None:
             )
 
     return AsyncValidator(_validate_iunique_value)
+
+
+def validate_valid_objectid(field_name: str) -> BeforeValidator:
```
MarkLark86, line 123: @eos87 This should not be needed as we already have [superdesk.core.resources.fields.ObjectId](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L144) that'll do this validation for us

newsroom/navigations/service.py
```diff
@@ -9,12 +7,12 @@ class NavigationsService(NewshubAsyncResourceService[NavigationModel]):
...
         """Remove references in products to this navigation entry once it is deleted"""
+        from newsroom.products import ProductsService
 
         await super().on_delete(doc)
 
         navigation = doc.id
-        products = get_resource_service("products").find(where={"navigations": str(navigation)})
-
-        for product in products:
+        products_cursor = await ProductsService().search({"navigations": navigation})
```
MarkLark86, line 15: @eos87 `ProductService` is being constructed once here and once for each product affected in the for loop. It would be good to construct it once, store in a local variable, and re-use that variable wherever you need to use the service

newsroom/navigations/utils.py
```diff
@@ -17,13 +17,15 @@ async def get_navigations(user: UserData | None, company: Company | None, produc
     """
     Returns list of navigations for given user and company
     """
+    from newsroom.products import get_products_by_company
```
MarkLark86, line 20: @eos87 Is this required here inside the function? As we're already importing from `newsroom.products` at the top of this function, that tells me there shouldn't be a cyclic import issue

newsroom/news_api/news/atom/views.py
```diff
@@ -64,7 +64,7 @@ def _format_update_date(date):
     # TODO-ASYNC: replace once the service is ready
     response = await get_internal("news/search")
     company = get_company_or_none_from_request(None)
-    products = get_products_by_company(company.to_dict(context={"use_objectid": True}) if company else None)
+    products = await get_products_by_company(company.to_dict(context={"use_objectid": True}) if company else None)
```
MarkLark86, line 67: I'm now wondering if we should default `use_objectid` to True, so we will always get ObjectId instances where required. I will look into that

newsroom/products/service.py
```diff
@@ -0,0 +1,57 @@
...
+        return res
+
+    async def on_deleted(self, doc: ProductResourceModel):
+        from newsroom.users import UsersService
+        from newsroom.companies.companies_async import CompanyService
+
+        lookup = {"products._id": doc.id}
+
+        for service in [UsersService(), CompanyService()]:
+            search_cursor = await service.search(lookup)
```
MarkLark86, line 54: Just curious, why use dictionary over Pydantic models?
```python
async for item in await service.search(lookup):
    updates = {"products": [p for p in item.products if p._id != doc.id]}
    await service.system_update(item.id, updates)
```

Pydantic allows us to use model fields without needing to worry if they're a string or ObjectId, as they're converted to the correct format for us.

newsroom/types/company.py
```diff
@@ -16,6 +17,9 @@ class CompanyProduct:
     section: ProductType
     seats: int = 0
 
+    def to_dict(self):
```
MarkLark86, line 20: We should look at adding this into superdesk.core.resources.dataclass somehow

newsroom/types/products.py
```diff
@@ -7,3 +14,33 @@ class ProductType(str, Enum):
...
+    product_type: ProductType = ProductType.WIRE
+    navigations: list[
+        Annotated[
+            ObjectId,
+            validators.validate_data_relation_async("navigations"),
+            validate_valid_objectid("navigations"),
+        ]
+    ] = Field(default_factory=list)
+
+    # obsolete
```
MarkLark86, line 38: @eos87 Originally product permissions where on the Product itself. This was changed so these permissions are now stored on Company and Use resources

MarkLark86 (commented):
@eos87 Looks good, just a couple small things I noticed

newsroom/core/resources/validators.py
eos87: Good one, I'll adjust it.

newsroom/navigations/utils.py
eos87: Actually, there is a cyclic import. If you see it again, the top import is `newsroom.products.products` and the one that has the cyclic import is `newsroom.products` (meaning `__init__.py`). Maybe I can double check what is producing the issue but I didn't want to waste too much time on that at the given time as there were some trickier things to deal with.

newsroom/products/service.py
eos87: I did have it like that but there was a problem, that I was not able to track, when the `p._id` was not converted into a pydantic model and was a dictionary. Once again, time was pressing, so I went with dicts as it allowed the tests to pass and the feature worked as expected. I will double check it today though.

newsroom/navigations/utils.py
MarkLark86: Yeah fair enough, all good then

newsroom/products/service.py
MarkLark86: All good, no need to spend more time on it, as it works

PR #1139, opened 2024-10-25
Migrate design Blueprint to an EndpointGroup [NHUB-568]

devketanpro:
### Purpose
Migrate design Blueprint to an async EndpointGroup

Resolves: #[NHUB-568](https://sofab.atlassian.net/browse/NHUB-568)

[NHUB-568]: https://sofab.atlassian.net/browse/NHUB-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/design.py
+++ b/newsroom/design.py
@@ -1,14 +1,25 @@
+from typing import Optional
+from pydantic import BaseModel
 from datetime import datetime, timedelta
-from superdesk.flask import Blueprint, url_for, render_template
 
-blueprint = Blueprint("design", __name__)
+from superdesk.flask import url_for, render_template
+from superdesk.core.web import EndpointGroup
+from superdesk.core.types import Request
+from superdesk.core.module import Module
+
+blueprint = EndpointGroup("design", __name__)
 
 
-@blueprint.route("/design/detail")
-async def detail():
+class RouteArguments(BaseModel):
+    page: Optional[str] = None
+
+
+@blueprint.endpoint("/design/detail", auth=False)
+async def detail(args: None, params: None, req: Request):
     item = {
         "headline": "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
         "slugline": "Slugline",
+        "type": "text",
         "byline": "Author byline",
         "versioncreated": datetime.utcnow(),
         "description_html": "<p>IN this exclusive interview, Aussie rock legend Jimmy Barnes reveals how he almost paid the ultimate price for his years of hard living.</p>",  # noqa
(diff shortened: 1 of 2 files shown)


newsroom/design.py
```diff
@@ -39,11 +50,14 @@ async def detail():
     return await render_template("wire_item.html", item=item, previous_versions=previous_versions)
 
 
-@blueprint.route("/design/")
-async def index():
+@blueprint.endpoint("/design/", auth=False)
+async def index(args: None, params: None, req: Request):
```
eos87: nit: you probably don't need all these arguments
```suggestion
async def index():
```

newsroom/design.py
```diff
@@ -39,11 +50,14 @@ async def detail():
...
     return await render_template("design_index.html")
 
 
-@blueprint.route("/design/<page>")
-async def page(page):
-    return await render_template("design_%s.html" % page)
+@blueprint.endpoint("/design/<string:page>", auth=False)
+async def page(args: RouteArguments, params: None, req: Request):
+    return await render_template("design_%s.html" % args.page)
```
eos87: nit:
```suggestion
    return await render_template(f"design_{args.page}.html")
```

Also, we might need to check if the template exists first, otherwise it will be a 500 error. But we can do it in a separate PR.

newsroom/design.py
```diff
@@ -39,11 +50,14 @@ async def detail():
...
+@blueprint.endpoint("/design/", auth=False)
+async def index(args: None, params: None, req: Request):
     return await render_template("design_index.html")
 
 
-@blueprint.route("/design/<page>")
-async def page(page):
-    return await render_template("design_%s.html" % page)
+@blueprint.endpoint("/design/<string:page>", auth=False)
+async def page(args: RouteArguments, params: None, req: Request):
```
eos87, line 58: nit:
```suggestion
async def page(args: RouteArguments):
```

newsroom/design.py
```diff
@@ -1,14 +1,25 @@
+from typing import Optional
```
eos87: We can use `| None` instead
```suggestion
```

newsroom/design.py
```diff
@@ -1,14 +1,25 @@
...
+from superdesk.core.types import Request
+from superdesk.core.module import Module
 
+blueprint = EndpointGroup("design", __name__)
 
-@blueprint.route("/design/detail")
-async def detail():
+
+class RouteArguments(BaseModel):
+    page: Optional[str] = None
```
eos87: Following above suggestion
```suggestion
    page: str | None = None
```

eos87 (approved):
Looks good. I just left a few silly comments.

newsroom/design.py
devketanpro: I think here we need all.

PR #1143, opened 2024-10-28
[NHUB-532] Upgrade Wire to async resource

MarkLark86:
### Purpose
Upgrade Wire resource and search to async.

### What has changed

Resolves: #NHUB-532
Note: Uses Superdesk branch NHUB-540 for now, will change once the Superdesk PR is merged.
Depends on https://github.com/superdesk/superdesk-core/pull/2740


dev-requirements.txt
eos87, line 1: The requirements are pointing to your local environment 😄

newsroom/notifications/module.py
```diff
@@ -25,6 +25,7 @@
                 name="user_created",
                 keys=[("user", 1), ("_created", -1)],
                 collation={"locale": "en", "strength": 2},
+                unique=False,
```
eos87, line 28: why is this change to the index here?

newsroom/notifications/services.py
```diff
@@ -14,36 +16,38 @@ async def create_or_update(self, entries: list[dict[str, Any]]) -> list[str]:
...
+                if "data" in doc:
+                    updates["data"] = doc["data"] or {}
                 await self.update(
                     notification_id,
-                    updates={
-                        "_created": now,
-                        "action": doc.get("action") or original.action,
-                        "data": doc.get("data") or original.data,
-                    },
+                    updates=updates,
```
eos87: nit: I think the second param is the updates dict so to simplify and make the `self.update` one-liner
```suggestion
                    updates,
```

dev-requirements.txt
MarkLark86: This was just until the supedesk-core PR was merged, so the tests on this PR can run. I have reverted this change now

newsroom/notifications/module.py
MarkLark86: In the new code, we're defaulting to `unique:True`, and after auditing the problems we've had with indexes and build issues and the code, I made these changes to make sure those that are meant to be unique are unique

newsroom/push/publishing.py
```diff
@@ -101,14 +107,31 @@ async def publish_item(self, doc: Item, original: Item):
...
-            service.patch(_id, updates={"associations": None})
+        doc_id = await self.publish_doc_to_content_api(doc)
         if "evolvedfrom" in doc and parent_item:
-            service.system_update(parent_item["_id"], {"nextversion": _id}, parent_item)
-        return _id
+            await wire_search.service.system_update(parent_item.id, {"nextversion": doc_id})
+        return doc_id
+
+    async def publish_doc_to_content_api(self, item_dict: dict[str, Any]) -> str:
+        # Copied from content_api.publish.utils.publish_doc_to_content_api
```
eos87, line 116:
```suggestion
        # Copied from content_api.publish.utils.publish_doc_to_content_api
```
is there a way to reuse it instead or there is an intention to remove this repeated code?

newsroom/push/utils.py
```diff
@@ -31,14 +31,16 @@ def set_dates(doc: dict[str, Any]):
...
     ancestors = (doc.get("ancestors") or []) + [doc["guid"]]
 
     for i in range(50):
-        updates = {"ancestors": ancestors + (next_item.get("ancestors") or []), "original_id": doc["original_id"]}
-        service.system_update(next_item["_id"], updates, next_item)
-        next_item = service.find_one(req=None, evolvedfrom=next_item["_id"])
+        updates = {"ancestors": ancestors + (next_item.ancestors or []), "original_id": doc["original_id"]}
+        # updates = {"ancestors": ancestors + (next_item.get("ancestors") or []), "original_id": doc["original_id"]}
+        await wire_search.service.system_update(next_item.id, updates)
+        # service.system_update(next_item.id, updates, next_item)
```
eos87: we can remove these comments I think
```suggestion
        updates = {"ancestors": ancestors + (next_item.ancestors or []), "original_id": doc["original_id"]}
        await wire_search.service.system_update(next_item.id, updates)
```

newsroom/search/filters.py
```diff
@@ -0,0 +1,440 @@
...
+    if request.is_admin:
+        if len(request.args.navigation_ids):
+            request.products = get_products_by_navigation(
+                request.args.navigation_ids, product_type=request.section.value
+            )
+        elif len(request.args.product_ids):
+            # TODO-ASYNC: Convert to Async service when it's available
+            request.products = list(
+                products_service.get_from_mongo(req=None, lookup={"_id": {"$in": request.args.product_ids}})
+            )
```
eos87, line 90: The PR for products async has been merged, although we can replace this in a separate PR if you like.

newsroom/search/filters.py
```diff
@@ -0,0 +1,440 @@
...
+
+def apply_products_filter(request: NewshubSearchRequest) -> None:
+    if request.is_admin and not len(request.args.navigation_ids) and not request.args.product_ids:
+        # admin will see everything by default
+        return
+    elif request.section is None:
+        # This should not happen, as it's prefilled by search service
+        return
+
+    # TODO-ASYNC: Convert to async Product model when available
```
eos87, line 276: same with products here

newsroom/wire/filters.py
```diff
@@ -0,0 +1,180 @@
...
+    ignoreLatest: bool = False
+
+    #: If ``True``, will exclude embargoed items from this search
+    exclude_embargoed: bool = False
+
+    #: If ``True``, will only include embargoed items in this search
+    embargoed_only: bool = False
+
+    #: If ``True``, will apply the configured NewsOnly filter for this search
+    newsOnly: bool = False
```
eos87: just out of curiosity, why is the camelCase in this one?

newsroom/wire/service.py
```diff
@@ -0,0 +1,545 @@
...
+
+    async def get_current_user_bookmarks_count(self, section: SectionEnum | None = None) -> int:
+        user = get_user_or_none_from_request(None)
+        if not user:
+            return 0
+
+        cursor = await self.search(
+            NewshubSearchRequest(
+                section=section or self.section,
+                args=WireSearchRequestArgs(bookmarks=[user.id], page_size=0),
```
eos87, line 89: will this count work as expected if the page_size is equal to 0?

eos87 (approved):
Great job with the search service, it looks much better now! 

I left a few minor comments/questions. Two last suggestions (don't have to be in this PR of course): 
- we should add some docstrings into the some of the most important methods
- please update the dev-requirements as there are still some references to your local environment

MarkLark86:
> Great job with the search service, it looks much better now!
> 
> I left a few minor comments/questions. Two last suggestions (don't have to be in this PR of course):
> 
> * we should add some docstrings into the some of the most important methods
> * please update the dev-requirements as there are still some references to your local environment

Will add docstrings to the search class methods.
As for the dev-requirements, see [pip-tools issue #2131](https://github.com/jazzband/pip-tools/issues/2131). I've fixed this by find/replace. We won't have this issue once pip-tools fixes it.

newsroom/wire/filters.py
MarkLark86: So we don't have to change the front-end code. I have changed this now, so we can support camelCase from front-end and use snake_case on the back-end

newsroom/search/filters.py
MarkLark86: I'll do these changes with the Agenda task I'm currently working on

newsroom/wire/service.py
MarkLark86: Yes, because it uses the hits metadata returned by elastic to [get the count](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/cursor.py#L105).
image

newsroom/push/publishing.py
MarkLark86: There are subtle changes with this one vs the one in the ContentAPI, such as using the WireItem model instead of ContentAPIItem model (with changes to the fields & schema). Kept it here because it's a small function, and barely ever changed in both Newshub and ContentAPI for many years. Last change in ContentAPI publishing, in regards to Newshub usage, was 4+ years ago.

PR #1151, opened 2024-11-06
Send copy of the email sent to the user who submitted the form [CPCN-979]

devketanpro:
### Purpose
Send copy of the email sent to the user who submitted the Coverage inquiry form

### What has changed
add user email to the recipients

### Steps to test

In NewsPro Calendar section, each event or planning item has “Coverage inquiries” form that a user can fill out and when they submit it, an email is sent to whichever email addresses we’ve specified in General Setting. from the PR logged user also get same mail.

### Checklist

Resolves: #[CPCN-979](https://sofab.atlassian.net/browse/CPCN-979)

[CPCN-979]: https://sofab.atlassian.net/browse/CPCN-979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


petrjasek (dismissed):

newsroom/agenda/email.py
```diff
@@ -83,6 +83,9 @@ def send_coverage_request_email(user, message, item):
     name = "{} {}".format(user.get("first_name"), user.get("last_name"))
     email = user.get("email")
 
+    # send coverage request email copy to current User.
+    recipients.append(email)
```
petrjasek: current user should be rather in cc and not actual recipient

devketanpro: alright, and one more question— is this change CP-specific? If so, we should consider adding a config option to make it configurable.

petrjasek: yep that would make sense :+1:

newsroom/agenda/email.py
```diff
@@ -84,7 +85,7 @@ def send_coverage_request_email(user, message, item):
     email = user.get("email")
 
     # send coverage request email copy to current User.
-    recipients.append(email)
+    cc = [email] if app.config.get("INCLUDE_CURRENT_USER_IN_CC") else []
```
petrjasek: would add something to the config name to make it more obvious it's related to coverage request, eg.:
`COVERAGE_REQUEST_EMAIL_CC`

newsroom/email.py
```diff
@@ -81,13 +81,16 @@ def handle_long_lines_html(html):
 
 
 @celery.task(soft_time_limit=120)
-def _send_email(to, subject, text_body, html_body=None, sender=None, sender_name=None, attachments_info=None):
+def _send_email(to, subject, text_body, cc=None, html_body=None, sender=None, sender_name=None, attachments_info=None):
```
petrjasek: not sure if that would break something, would be better to add new param at the end

petrjasek (approved):

PR #1154, opened 2024-11-07
Send copy of the email sent to the user who submitted the form [CPCN-979] (#1151)

devketanpro:


petrjasek:
it would be better to just cherry-pick that fix and push to the branch

devketanpro:
I attempted to push the commit directly to the branch but encountered permission restrictions, so I created a PR instead. Could you please help by pushing it?

petrjasek:
done

PR #1157, opened 2024-11-11
[NHUB-583] fix(celery): Await when sending task to celery

MarkLark86:
### Purpose
Sending tasks to celery was not working, as the task sending itself was not awaited (therefor never executed in event loop).

### What has changed
* Add await when sending task to celery
* Enable tests for unit tests that require celery, that we previously disabled (before async celery was supported)
* Fix unit tests

### DIFF
--- a/newsroom/reports/content_activity.py
+++ b/newsroom/reports/content_activity.py
@@ -1,4 +1,4 @@
-from typing import Any, cast
+from typing import Any, cast, TypedDict
 from copy import deepcopy
 
 from quart_babel import gettext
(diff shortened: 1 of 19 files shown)


newsroom/reports/content_activity.py
```diff
@@ -84,16 +85,16 @@ async def get_items(args):
...
         if not len(items):
             break
 
         source["from"] += CHUNK_SIZE
-        yield items
+        yield [item.to_dict() for item in items]
 
 
-def get_aggregations(args, ids):
+async def get_aggregations(args, ids):
```
devketanpro: missing types here

devketanpro (approved):

PR #1160, opened 2024-11-13
Move the searched fields toggle inline with the advanced search pills

dzonidoo:
NHUB-580

### Purpose

To save space, the toggles should be moved inline with the advanced search result pills. Functionally, they belong together anyway.

### What has changed

Deleted one row in filter tag section
### Steps to test

Activate Advanced Search

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/assets/search/components/SearchResultsBar/SearchResultsAdvancedSearchRow.tsx
+++ b/assets/search/components/SearchResultsBar/SearchResultsAdvancedSearchRow.tsx
@@ -150,42 +150,48 @@ export function SearchResultsAdvancedSearchRow({
         body_html: gettext('Body'),
     };
 
+    const fieldSearchTags = (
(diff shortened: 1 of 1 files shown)


assets/search/components/SearchResultsBar/SearchResultsAdvancedSearchRow.tsx
```diff
@@ -150,42 +150,40 @@ export function SearchResultsAdvancedSearchRow({
         body_html: gettext('Body'),
     };
 
+    const fieldSearchTags = (
+        <div className="toggle-button__group toggle-button__group--spaced toggle-button__group--compact" data-test-id='search-results--advanced-fields'>
+            <span className="search-result__tags-list-row-helper-text me-1">inside</span>
```
tomaskikutis: Does this "inside" string supposed to be here? If so translate it with gettext

tomaskikutis (approved):

PR #1163, opened 2024-11-15
Newshub: Push changes to support 1 Planning to many Events [STT-51]

devketanpro:
### Checklist

Resolves: #[STT-51]

[STT-51]: https://sofab.atlassian.net/browse/STT-51?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarkLark86:
@devketanpro There's some behave tests failing with:
```if agenda.get("item_type") == "event" and planning["_id"] not in agenda["planning_ids"]```
```KeyError: '_id'```

File: newsroom/push/publishing.py [hunk 1]
```diff
@@ -271,6 +276,9 @@ async def publish_planning_into_event(self, planning: dict[str, Any]) -> Optiona
 
         agenda = deepcopy(orig_agenda)
 
+        if agenda.get("item_type") == "event" and planning["_id"] not in agenda["planning_ids"]:
+            agenda["planning_ids"].append(planning["_id"])
+
         if (
```
MarkLark86: We do not want to process any of these other conditions for events with a secondary link. These are side affects we want to avoid. If the link is secondary, then we only want to add to the `planning_ids` and `coverages`, and nothing else. In this case may I suggest a different function altogether if we're processing a secondary link

newsroom/push/views.py
```diff
@@ -47,6 +47,10 @@ async def handle_publish_planning(item):
     plan_id = await publisher.publish_planning_item(item, orig.to_dict() if orig else {})
     event_id = await publisher.publish_planning_into_event(item)
 
+    if item.get("related_events"):
+        for events in item["related_events"]:
+            event_id = await publisher.publish_planning_into_event(item, events["_id"])
```
MarkLark86: Only if the link is not primary (as that would have already been handled by the `publish_planning_into_event` thats executed straight after `publish_planning_item` call

newsroom/types/agenda.py
```diff
@@ -247,6 +247,8 @@ class AgendaItem(ResourceModel, ModelWithVersions):
     planning_items: Annotated[
         list[AgendaPlanningItem], fields.nested_list(include_in_parent=True), Field(default_factory=list)
     ]
+    planning_ids: Annotated[list[fields.Keyword], Field(default_factory=list)]
+    event_ids: Annotated[list[fields.Keyword], Field(default_factory=list)]
```
MarkLark86, line 251: Where are the `event_ids` populated?

devketanpro: in the ticket does not specify the functionality or behavior of the `event_ids` field—such as where or how it should be populated. It only mentions that we need to add two fields to the schema

newsroom/push/views.py
devketanpro: got it

newsroom/types/agenda.py
MarkLark86: That's my bad, I apologise. Here you would need to store all IDs from the related events onto that event id field

newsroom/push/publishing.py
```diff
@@ -298,3 +300,27 @@ async def publish_planning_into_event(self, planning: dict[str, Any]) -> Optiona
...
+
+    async def publish_planning_related_events(self, planning: dict[str, Any], related_event: dict[str, Any]):
+        if not related_event.get("link_type") == "secondary":
+            return None
+
+        service = AgendaItemService()
+        event_id = related_event.get("_id")
+
+        event = await service.find_by_id(event_id)
+        agenda = event.to_dict()
```
MarkLark86: Instead of using the entire item dict for updates, you can instead construct a list of updates with only the required fields, starting with an empty dict. For example:

```python
from newsroom.types import AgendaItemType

event = await service.find_by_id(event_id)
updates: dict[str, Any] = {}

if event.item_type == AgendaItemType.EVENT and planning and planning.get("guid") not in event.planning_ids:
    updates["planning_ids"] = event.planning_ids + [planning["guid"]]

...

await service.update(event_id, updates)
```

Notice also how above I am using the `event` as a ResourceModel instance and not a dictionary. It's better to use instances of the model instead of a dictionary.

newsroom/push/publishing.py
```diff
@@ -298,3 +300,27 @@ async def publish_planning_into_event(self, planning: dict[str, Any]) -> Optiona
...
+
+        if agenda.get("item_type") == "event" and planning and planning.get("guid") not in agenda["planning_ids"]:
+            agenda["planning_ids"].append(planning["guid"])
+
+        if event_id not in agenda["event_ids"]:
```
MarkLark86: This actually doesn't belong here on the Event, as the Event already knows it's own ID. Instead you can store the list of `event_ids` in a function on the Agenda [convert_planning_to_agenda_dict](https://github.com/superdesk/newsroom-core/blob/async/newsroom/agenda/agenda_service.py#L196).
You would add there:

```python
plan["event_ids"] = [link["_id"] for link in (planning_item.get("related_events") or [])]
```

This way the Planning item stores the list of Event IDs that it is linked to

newsroom/push/publishing.py
```diff
@@ -298,3 +300,27 @@ async def publish_planning_into_event(self, planning: dict[str, Any]) -> Optiona
...
+            agenda["event_ids"].append(event_id)
+
+        planning_item = await service.find_by_id(planning["guid"])
+        if planning_item:
+            agenda["coverages"] = planning_item.coverages
```
MarkLark86: The Event might already have coverages from a previous push, from a different Planning item. This would then wipe those others off the Event.

Instead you will need to add the coverages from this Planning item to the Event, making sure not to include duplicates (based on the coverage_id field).

newsroom/push/views.py
```diff
@@ -47,6 +47,10 @@ async def handle_publish_planning(item):
     plan_id = await publisher.publish_planning_item(item, orig.to_dict() if orig else {})
     event_id = await publisher.publish_planning_into_event(item)
 
+    if item.get("related_events"):
+        for events in item["related_events"]:
+            event_id = await publisher.publish_planning_related_events(item, events)
```
MarkLark86: We won't desire to change the `event_id` used here. The `event_id` stored above is the one we require, as that will include the primary linked Event, which is what is used when sending notifications

tests/core/test_push.py
```diff
@@ -1141,3 +1141,147 @@ async def test_global_topic_after_deleting_user(client, app):
         item["guid"], [topic], list(users.values()), companies
     )
     assert matching == {topic_id}
+
+
+# STT-51
+async def test_planning_to_many_events(client, app):
```
MarkLark86, line 1147: Can we also have tests to cover this scenario:
* Push 1 Event
* Push 2 Planning items (that have a link to the same event)

Test:
* The list of Event coverages contains all coverages from both Planning 1 & 2, and does not contain any duplicates

newsroom/push/publishing.py
devketanpro: so in this case we have to add `event_ids` field in the [planning schema](https://github.com/superdesk/newsroom-core/blob/async/newsroom/types/agenda.py#L146) right ?

MarkLark86: No. AgendaItem is used for both Event & Planning items.

devketanpro: but if we populate `events_ids` here [convert_planning_to_agenda_dict](https://github.com/superdesk/newsroom-core/blob/async/newsroom/agenda/agenda_service.py#L196). we got pydantic validation error

MarkLark86: Ah sorry, it should be in that agenda variable instead, my bad

newsroom/push/publishing.py
```diff
@@ -298,3 +299,27 @@ async def publish_planning_into_event(self, planning: dict[str, Any]) -> Optiona
...
+            return None
+
+        service = AgendaItemService()
+        event_id = related_event.get("_id")
+
+        event = await service.find_by_id(event_id)
+        updates: dict[str, Any] = {}
+
+        if event.item_type == AgendaItemType.EVENT and planning and planning.get("guid") not in event.planning_ids:
+            updates["planning_ids"] = event.planning_ids + [planning["guid"]]
```
MarkLark86, line 317: This line will also need to be added in the `publish_planning_into_event` function, so an Event with primary link also get's it's `planning_ids` populated

MarkLark86:
I have a PR in place that will fix the test failures https://github.com/superdesk/newsroom-core/pull/1169
Once it's merged, I will update the upstream feature branch this PR is based on

MarkLark86 (approved):

PR #1164, opened 2024-11-15
[NHUB-572] Upgrade NewsAPI Search and Feed services to async

eos87:
### Purpose
Upgrade NewsAPI Search and Feed to new async framework

### What has changed
- Implemented new Token base authentication for NewsAPI
- Fixed recursivity issue with AuthorizationError exception 
- Added additional error handlers and log useful information for unhandled exceptions 
- Implemented new async search service for NewsAPI
- Implemented NewsAPI Feed base on new service 
- Removed old sync code 

> [!IMPORTANT]  
> I've skipped one behave's test becase I didn't have enough time to fix it. It's not a big problem. I think it is just something connected with the format of the date we're using. We can either have a separate small ticket for that or I could open a second PR for this ticket.

Thanks for checking!

Resolves: [NHUB-572]

[NHUB-572]: https://sofab.atlassian.net/browse/NHUB-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/news_api/api_tokens/__init__.py
```diff
@@ -23,6 +23,7 @@
 from . import views  # noqa
 
 
+# TODO-ASYNC: Remove altogether once all the modules from news_api are async
 class CompanyTokenAuth(TokenAuth):
```
eos87, line 27: We should have a separate ticket to double check if this is still needed. It probably isn't but better to check if carefully and removed it :)

MarkLark86: I'll be going over such things with my cleanup tasks that I have for Newshub (there's a lot that will need to be cleaned up)

newsroom/news_api/api_tokens/auth.py
```diff
@@ -0,0 +1,139 @@
...
+                    updates["rate_limit_requests"] = token.get("rate_limit_requests", 0) + 1
+
+        rate_limit_period = get_app_config("RATE_LIMIT_PERIOD")
+        if rate_limit_period and new_period:
+            updates["rate_limit_expiry"] = now + timedelta(seconds=rate_limit_period)
+
+        if updates:
+            tokens_service.patch(token.get("token"), updates)
+
+            # TODO-ASYNC: check if we should use request.storage instead
```
MarkLark86: Yes, internally they point to the same thing. `request.storage.request` is a wrapper around `flask.g` and `request.storage.session` is a wrapper around `flask.session`

So you should be able to replace any code that uses the above with the new request storage equivalent.

newsroom/news_api/news/feed/service.py
```diff
@@ -1,137 +1,96 @@
...
             exclude_ids = []
 
             for item in desc_items:
-                if item.get("versioncreated").strftime("%Y-%m-%dT%H:%M:%S") != last_datetime:
+                if item.get("versioncreated") != last_datetime:
                     break
 
                 exclude_ids.append(item.get("_id"))
 
-            args["exclude_ids"] = ",".join(exclude_ids)
```
MarkLark86, line 118: `exclude_ids` list is being constructed but not used in the `next_page` HATEOAS link. Any reason for this?

newsroom/news_api/news/search_service.py
```diff
@@ -1,669 +1,123 @@
...
 
-        # Can't get the exclude projection to work do pop the exclude fields here
-        exclude_fields = (
-            self.mandatory_exclude_fields.union(set(orig_request_params.get("exclude_fields").split(",")))
-            if orig_request_params.get("exclude_fields")
-            else self.mandatory_exclude_fields
-        )
+class NewsApiSearchServiceAsync(BaseNewshubSearchService[NewsApiSearchRequestArgs, WireItem]):
+    search_args_class = NewsApiSearchRequestArgs
+    filters = cast(list[SearchFilterFunction], default_search_filters)
```
MarkLark86: If you defined the type of `default_search_filters` above, then this `cast` here would not be required (and allows mypy to pick up type errors for these as well)

newsroom/news_api/news/types.py
```diff
@@ -0,0 +1,138 @@
...
+    "versioncreated",
+    "firstcreated",
+    "headline",
+    "byline",
+    "slugline",
+}
+
+
+class NewsApiSearchRequestArgs(BaseSearchRequestArgs):
+    # TODO-ASYNC: check if we need to raise exception on unknown fields
```
MarkLark86: This might be ideal, if a client requests a field that is not supported by the API, then we should raise an error

newsroom/news_api/news/feed/service.py
eos87: what do you mean is not used? It is attached to the URL with the rest of parameters.

MarkLark86 (approved):

PR #1167, opened 2024-11-18
[STT-52] Layout and style improvements as part of the tiket

fritzSF:
### Checklist

Resolves: #[issue-number]


dzonidoo (approved):

assets/styles/index.scss
```diff
@@ -2526,25 +2532,38 @@ article.list {
     .agenda-planning__preview-metadata {
         display: flex;
         flex-direction: column;
-        margin-block-end: 20px !important;
+        //margin-block-end: var(--space--2) !important;
```
MarkLark86: If a style is no longer needed, would be good to remove it

MarkLark86 (approved):
@fritzSF Looks good, just one not required commented line in a stylesheet

PR #1168, opened 2024-11-20
[NHUB-575] fix: Filters users by ProductID not working

MarkLark86:
### Purpose
In the CompanyAdmin page, filtering users by the available Products is not working, it always shows all users.

### What has changed
* The front-end is sending the param as "products._id" not "products_id", so add support for both

Resolves: [NHUB-575](https://sofab.atlassian.net/browse/NHUB-575)

[NHUB-575]: https://sofab.atlassian.net/browse/NHUB-575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/users/views.py
+++ b/newsroom/users/views.py
@@ -2,8 +2,8 @@ import re
 import json
 
 from copy import deepcopy
-from typing import Any, Dict, Optional
-from pydantic import BaseModel, field_validator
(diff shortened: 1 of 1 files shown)


eos87 (approved):
👏

PR #1169, opened 2024-11-21
Improve error exceptions

MarkLark86:
### Purpose
When an uncaught exception was raised, another exception `TypeError: get_headers() takes from 1 to 2 positional arguments but 3 were given` was raised.

### What has changed
* Add a catch all other exception handler, so we can control more the error handling across all error types (this is a big hammer to fix above exception, but I could not find the root cause).
* Add a 500.html template and use that to show InternalServerError to the user
* Add `AsyncioIntegration` to Sentry
* Also fixed an issue in the Monitoring app raising an exception when getting the bookmark count, causing the page to crash
* Log the validation error, so we can debug where it happens (we should really separate out UserValidation and SystemValidation errors. i.e. is validation from the request data or data constructed from code)
* Use strict model fields in most of Newshub resources (as we've proved it works so far in Newshub)
* Raise pydantic.ValidationError for Topic Folder name (we're using the MongoDB index to raise the exception for us, so catch it and re-raise it)

Resolves: [NHUB-572](https://sofab.atlassian.net/browse/NHUB-572)

[NHUB-572]: https://sofab.atlassian.net/browse/NHUB-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/core/resources/__init__.py
+++ b/newsroom/core/resources/__init__.py
@@ -1,6 +1,6 @@
 from .model import NewshubResourceModel, NewshubResourceDict
 from .service import NewshubAsyncResourceService
-from .validators import validate_ip_address, validate_auth_provider
+from .validators import validate_ip_address, validate_auth_provider, raise_custom_validation_error
 
 __all__ = [
     "NewshubResourceModel",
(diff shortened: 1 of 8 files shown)


eos87 (approved):
Looks good to me!

PR #1179, opened 2024-11-27
Fix iCalendar DTEND Handling for Multi-Day All-Day Events[CPCN-840]

devketanpro:
### Purpose
This PR fixes the handling of multi-day all-day events in the iCalendar format. The issue occurs when the `DTEND` property does not correctly reflect the last day of the event, causing calendar applications to exclude the final day. This change ensures compliance with the iCalendar standard, where the `DTEND` field is **exclusive**, and the last day of an **all-day** event is properly included

### What has changed
The `DTEND` field for multi-day all-day events is updated to include the day after the last event date (i.e., adding one day to `DTEND`).

### Steps to test

1. Create a multi-day all-day event (e.g., from November 18–20, 2024).
2. Generate the iCalendar output and check that DTEND is set to November 21 (one day after the last day).
3. Confirm that the event shows correctly in calendar applications like Google Calendar or Outlook

Resolves: #[CPCN-840]

[CPCN-840]: https://sofab.atlassian.net/browse/CPCN-840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/core/test_ical_formatter.py
+++ b/tests/core/test_ical_formatter.py
@@ -107,6 +107,5 @@ def test_onclusive_multiday():
     }
     formatter = iCalFormatter()
     output = formatter.format_item(event, item_type="agenda").decode("utf-8")
-    print(output)
     assert "DTSTART;VALUE=DATE:20241118" in output
     assert "DTEND;VALUE=DATE:20241121" in output
--- a/tests/core/test_ical_formatter.py
+++ b/tests/core/test_ical_formatter.py
@@ -78,7 +78,7 @@ def test_onclusive_all_day():
     formatter = iCalFormatter()
     output = formatter.format_item(event, item_type="agenda").decode("utf-8")
     assert "DTSTART;VALUE=DATE:20240101" in output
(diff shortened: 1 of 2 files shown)


tests/core/test_ical_formatter.py
```diff
@@ -93,4 +93,20 @@ def test_onclusive_no_end_time():
...
+        "name": "test",
+        "dates": {
+            "start": "2024-11-18T00:00:00",
+            "end": "2024-11-20T00:00:00",
+            "all_day": True,
+        },
+    }
+    formatter = iCalFormatter()
+    output = formatter.format_item(event, item_type="agenda").decode("utf-8")
+    print(output)
```
petrjasek: extra output

petrjasek (approved):

tests/core/test_ical_formatter.py
devketanpro: my bad

PR #1182, opened 2024-11-29
improve: Use new async signal system

MarkLark86:
### Purpose
Use the new async signal system from core, and fix usage for user delete and topic resource

### What has changed
* Convert the existing signals to `superdesk.core.AsyncSignal`
* Update where signals are sent to
* Re-connect signals that were commented out due to compatibility with old system
* Uncomment tests that were disabled due to old signals not used

Depends on: https://github.com/superdesk/superdesk-core/pull/2760


eos87 (approved):
🚀

PR #1183, opened 2024-11-29
Fix: Exclude restricted fields when copying Agenda items [STTNHUB-381]

devketanpro:
### Checklist

Resolves: #[STTNHUB-381]

[STTNHUB-381]: https://sofab.atlassian.net/browse/STTNHUB-381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/wire/views.py
+++ b/newsroom/wire/views.py
@@ -56,6 +56,8 @@ from newsroom.public.views import (
 from .search import get_bookmarks_count
 from .items import get_items_for_dashboard
 from ..upload import ASSETS_RESOURCE, get_upload
+from newsroom.companies.utils import restrict_coverage_info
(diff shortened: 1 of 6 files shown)


newsroom/wire/views.py
```diff
@@ -458,24 +460,33 @@ def bookmark():
 def copy(_id):
     item_type = get_type()
     item = get_entity_or_404(_id, item_type)
+    user = get_user()
```
MarkLark86: this should be `get_user(True)` as we require to get a user here

MarkLark86 (commented):
Some tests would be good please

MarkLark86 (approved):

PR #1187, opened 2024-12-02
Order the versions of completed coverages in the NH Agenda by version created / descending order [NHUB-588]

devketanpro:
Resolves: #[NHUB-588]

[NHUB-588]: https://sofab.atlassian.net/browse/NHUB-588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/wire/search.py
+++ b/newsroom/wire/search.py
@@ -128,7 +128,9 @@ class WireSearchService(BaseSearchService):
         except Forbidden:
             return 0
 
-    def get_items(self, item_ids, size=None, aggregations=None, apply_permissions=False):
+    def get_items(
+        self, item_ids, size=None, aggregations=None, apply_permissions=False, sort_by=None, sort_order="desc"
+    ):
         search = SearchQuery()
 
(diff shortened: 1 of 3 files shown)


petrjasek:
is it possible to sort it via elastic?

newsroom/wire/search.py
```diff
@@ -128,7 +128,9 @@ def get_bookmarks_count(self, user_id):
         except Forbidden:
             return 0
 
-    def get_items(self, item_ids, size=None, aggregations=None, apply_permissions=False):
+    def get_items(
+        self, item_ids, size=None, aggregations=None, apply_permissions=False, sort_by=None, sort_order="desc"
```
petrjasek: I would only use single `sort` param that would be used as is, this would be limited to single sort field

petrjasek (approved):

PR #1196, opened 2024-12-10
send emails via `newsroom.push` celery queue

petrjasek:
atm emails might get blocked by other tasks

CPCN-983

### What has changed
use `newsroom.push` queue for sending emails instead of the default `newsroom`

### Checklist

Resolves: CPCN-983

### DIFF
--- a/newsroom/web/default_settings.py
+++ b/newsroom/web/default_settings.py
@@ -433,6 +433,10 @@ CELERY_TASK_ROUTES = {
         "queue": celery_queue("newsroom.push"),
         "routing_key": "newsroom.push",
     },
+    "newsroom.email.*": {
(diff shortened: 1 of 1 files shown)


MarkLark86 (commented):
What if we have a new worker for just email queue, email tasks are important and we can control the number of workers specifically for emailing then

petrjasek:
I was rather thinking to keep this push queue as a high priority one and the rest would go to the default one, having a third one to scale is imo too much and would require some deployment changes

MarkLark86:
> I was rather thinking to keep this push queue as a high priority one and the rest would go to the default one, having a third one to scale is imo too much and would require some deployment changes

Fair enough, sounds good

MarkLark86 (approved):

PR #1197, opened 2024-12-11
Don't show empty metadata sections and create an empty state for metadata in general

dzonidoo:
NHUB-610

### Purpose

For the ‘Category’ section we show the label even if there are no values available. This is not consistent wit other metadata fields. The sections should be hidded if there is no data available

### What has changed

Rendering of metadata fields

### Steps to test

1. Open an article panel
2. Scroll to the metadata section

### Checklist

Resolves: #[issue-number]


tomaskikutis (approved):
All good, except when using conditionals or ternary expressions the condition should be a boolean e.g.
`result = <boolean> ? a : b` and not `result = <number> ? a : b`

dzonidoo:
yes ure right... ill change taht one and then merge

PR #1199, opened 2024-12-12
[NHUB-592] fix: User profile not showing if user has no product permissions

MarkLark86:
### Purpose
If the user does not have any product permissions, the user profile react app was not being rendered (which includes the logout menu). This was caused by the auth error template not rendering the user-profile react app not providing user session details.

### What has changed
* Move user_profile_data to a template global function, and remove from all views (let the template call it instead)
* Add `user_profile_js` file to the auth error template

Resolves: [NHUB-592](https://sofab.atlassian.net/browse/NHUB-592)

[NHUB-592]: https://sofab.atlassian.net/browse/NHUB-592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/agenda/views.py
+++ b/newsroom/agenda/views.py
@@ -19,7 +19,6 @@ from newsroom.auth.utils import (
 )
 from newsroom.ui_config_async import UiConfigResourceService
 from newsroom.formatters import get_formatters_id_and_names
-from newsroom.users import get_user_profile_data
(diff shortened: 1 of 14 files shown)


eos87 (approved):
Much cleaner!

PR #1212, opened 2024-12-17
Update Agenda preview to display multiple Events [STT-53]

devketanpro:
### Purpose
Update Ui views to display, secondary link events, and planning.

### Steps to test

[UI Views]
image
image

### Checklist

- [X] This pull request is adding missing TypeScript types to modified code segments where it's easy to do so with confidence

Resolves: #[STT-53]

[STT-53]: https://sofab.atlassian.net/browse/STT-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/agenda/components/AgendaPreviewEvent.tsx
```diff
@@ -24,110 +23,112 @@ class AgendaPreviewEventComponent extends React.Component<any, any> {
...
+                    ) : (
+                        this.props.events.map((event: any) => this.renderEvent(event))
+                    )}
                 </div>
             </div>
         );
     }
 }
 
-AgendaPreviewEventComponent.propTypes = {
```
MarkLark86, line 119: If you're going to remove `propTypes` for a React component, please replace with Typescript interfaces

assets/agenda/components/AgendaPreviewEvent.tsx
```diff
@@ -24,110 +23,112 @@ class AgendaPreviewEventComponent extends React.Component<any, any> {
...
+        this.reloadEvent = this.reloadEvent.bind(this);
     }
 
     componentDidMount() {
         this.reloadEvent();
     }
 
     componentDidUpdate(prevProps: any) {
-        if (get(prevProps.item, 'event_id') !== get(this.props.item, 'event_id')) {
+        if (get(prevProps.item, 'event_ids') !== get(this.props.item, 'event_ids')) {
```
MarkLark86: `lodash.get` should be replaced with optional chaining

assets/agenda/components/AgendaPreviewEvent.tsx
```diff
@@ -24,110 +23,112 @@ class AgendaPreviewEventComponent extends React.Component<any, any> {
...
-            this.props
-                .fetchEvent(this.props.item.event_id)
-                .finally(() => {
-                    this.setState({loading: false});
-                });
-        });
+        const {item, fetchEvent} = this.props;
+        const eventIds = item.event_ids || [];
+
+        this.setState({loading: true});
```
MarkLark86, line 62: Rather than sending a request for each Event, you could use the search endpoint to search by ids.
i.e.

```ts
// file: assets/agenda/actions.ts
export function fetchItemsById(ids: Array<string>): Promise<IRestApiResponse<IAgendaItem>> {
    return server.get(`/agenda/search?ids=${ids.join(',')}`);
}

export function fetchItemsByIdToRedux(ids: Array<string>) {
    return (dispatch: any) => {
        fetchItemsById(ids).then((response) => {
            dispatch(recieveItems(response));
        });
    };
}
```

And then use it here:
```ts
if (item.event_ids == null || item.event_ids.length == 0) {
    return;
}

this.setState({loading: true});
this.props
    .fetchItemsByIdToRedux(item.event_ids)
    .finally(() => {
        this.setState({loading: false});
    });
```

assets/agenda/components/AgendaPreviewEvent.tsx
```diff
@@ -24,110 +23,112 @@ class AgendaPreviewEventComponent extends React.Component<any, any> {
...
 }
 
-AgendaPreviewEventComponent.propTypes = {
-    item: PropTypes.object,
-    event: PropTypes.object,
-    fetchEvent: PropTypes.func,
+const mapStateToProps = (state: any, ownProps: any) => {
+    const eventIds = ownProps.item.event_ids || [];
+    return {
+        events: eventIds.map((eventId: string) => state.itemsById[eventId]),
```
MarkLark86: Rather than constructing this here, which creates a new value every time, you could add `itemsById` to the `mapStateToProps` and just use it in the render function.

You will also have to check for null values, as `itemsById[eventId]` is not guarantee'd to be non-null (if the Event was not loaded for whatever reason).

assets/agenda/components/AgendaPreviewPlanning.tsx
```diff
@@ -14,93 +15,203 @@ interface IOwnProps {
...
+        }
+    }
+
+    fetchSecondaryPlanningItems() {
+        const {item, planningId} = this.props;
+        const planningIds = item.planning_ids?.filter((id:string) => id !== planningId) || [];
+
+        this.setState({loading: true});
+        try {
+            Promise.all(planningIds.map((id: string) => fetchItem(id)));
```
MarkLark86: Same here, try and use 1 request to the server to retrieve all required items.

assets/agenda/components/AgendaPreviewPlanning.tsx
```diff
@@ -14,93 +15,203 @@ interface IOwnProps {
...
+                                />
+                            ))}
+                        </div>
+                    </div>
+                )}
+                {/* Secondary Planning Items */}
+                <div className="agenda-planning__container info-box">
+                    <div className="info-box__content">
+                        <span className="info-box__label">
+                            {gettext('Secondary Planning Items')}
```
MarkLark86: I'm not sure if `Secondary Planning Items` would make sense to the average user. Maybe something like `Related Planning Items`?

assets/agenda/components/AgendaPreviewPlanning.tsx
```diff
@@ -14,93 +15,203 @@ interface IOwnProps {
...
+    }
 }
 
-const mapStateToProps = (state: IAgendaState): IReduxStateProps => ({
-    wireItems: state.agenda.agendaWireItems || [],
-});
+const mapStateToProps = (state: any, ownProps: any) => {
+    const planningIds = ownProps.item.planning_ids || [];
+    return {
+        planningItems: planningIds.map((eventId: string) => state.itemsById[eventId]),
```
MarkLark86: Again here, try not to construct new values in `mapStateToProps`, as React will see this as a new value every time. Either construct these in the render function, or create a redux selector that memoizes the results based on provided args

assets/agenda/components/AgendaPreviewEvent.tsx
devketanpro: For this endpoint, we need to update the backend code as well. afaik, we can't handle the `_ids` param in this endpoint.

MarkLark86: It _should_ work, if not it needs to be fixed.
https://github.com/superdesk/newsroom-core/blob/async/newsroom/search/types.py#L196
https://github.com/superdesk/newsroom-core/blob/async/newsroom/search/filters.py#L158

devketanpro: it's, working, Thanks

assets/agenda/actions.ts
```diff
@@ -399,6 +399,18 @@ export function fetchItem(id: any) {
     };
 }
 
+export function fetchItemsById(ids: Array<string>): Promise<IRestApiResponse<IAgendaItem>> {
+    return server.get(`/agenda/search?ids=${ids.join(', ')}`);
+}
+
+export function fetchItemsByIdToRedux(ids: string[]): (dispatch: any) => Promise<void> {
```
MarkLark86: The 1st function is using `Array<string>` while the 2nd function is using `string[]`, though they're exactly the same. I believe by convention we're using `Array<type>` syntax. It would be good to stick to the one syntax as opposed to mixing different styles.

assets/agenda/components/AgendaPreviewEvent.tsx
```diff
@@ -16,118 +14,148 @@ import AgendaPreviewAttachments from './AgendaPreviewAttachments';
 import AgendaTags from './AgendaTags';
 import AgendaEdNote from './AgendaEdNote';
 import AgendaInternalNote from './AgendaInternalNote';
+import {IAgendaItem} from 'interfaces';
 
-class AgendaPreviewEventComponent extends React.Component<any, any> {
-    static propTypes: any;
-    constructor(props: any) {
+interface AgendaPreviewEventProps {
+    item: {
```
MarkLark86: I believe `item` should be `IAgendaItem`, or is this not the case here?

assets/agenda/components/AgendaPreviewEvent.tsx
```diff
@@ -16,118 +14,148 @@ import AgendaPreviewAttachments from './AgendaPreviewAttachments';
...
+        event_ids: Array<IAgendaItem['_id']>;
+    };
+    itemsById: Record<string, IAgendaItem>;
+    eventIds: string[];
```
MarkLark86: Same here, please use `Array<string>`

MarkLark86 (commented):
Looks good now, just a couple small style improvements

MarkLark86 (approved):

PR #1214, opened 2024-12-18
fix: Cannot download items from Agenda in json format [STTNHUB-362]

devketanpro:
### What has changed
 handle both JSON and Blob responses for downloads

Resolves: #[STTNHUB-362]

[STTNHUB-362]: https://sofab.atlassian.net/browse/STTNHUB-362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/server.ts
+++ b/assets/server.ts
@@ -7,13 +7,15 @@ function options(custom: any = {}) {
     return Object.assign({}, defaultOptions, custom);
 }
 
-function checkStatus(response: Response, rawResponse = false): Promise<any> {
+function checkStatus(response: Response, options?: { parseJson?: boolean }): Promise<any> {
+    const shouldParseJson = options?.parseJson ?? true;
+
     if (response.status === 204) {
         return Promise.resolve({});
     }
 
     if (response.status >= 200 && response.status < 300) {
-        if (rawResponse) {
+        if (!shouldParseJson) {
             return Promise.resolve(response);
         }
 
@@ -80,12 +82,17 @@ class Server {
      * @param {Object} data
      * @return {Promise}
      */
-    post(url: any, data: any, etag?: string, rawResponse = false) {
+    post(
+        url: any, 
+        data: any, 
+        etag?: string, 
+        config: { parseJson?: boolean } = {parseJson: true}
+    ) {
         return fetch(url, options({
             method: 'POST',
             headers: getHeaders(etag),
             body: data ? JSON.stringify(data) : null,
-        })).then((response) => checkStatus(response, rawResponse));
+        })).then((response) => checkStatus(response, config));
     }
 
     /**
--- a/assets/server.ts
+++ b/assets/server.ts
@@ -19,12 +19,8 @@ function checkStatus(response: Response, requestOptions: RequestOptions = {parse
     }
 
     if (response.status >= 200 && response.status < 300) {
-        if (!parseJson) {
-            return Promise.resolve(response);
-        }
-
         const contentType = response.headers.get('Content-Type');
-        if (contentType && contentType.includes('application/json')) {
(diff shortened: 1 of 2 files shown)


assets/wire/actions.ts
```diff
@@ -448,8 +448,17 @@ export function submitDownloadItems(items: any, params: any) {
         else{
             try {
                 const response = await server.post(url, payload);
-                const blob = await response.blob();
-                initiateDownload(blob);
+                if (typeof response === 'object' && !response.blob) {
```
petrjasek: I think we should do it based on `Content-Type` header

devketanpro: [checkStatus](https://github.com/superdesk/newsroom-core/blob/develop/assets/server.ts#L13-L17) function in the [post](https://github.com/superdesk/newsroom-core/blob/develop/assets/server.ts#L78) method already handles the `Content-Type` header, parsing JSON responses into objects. In `submitDownloadItems`, we handle this by checking if the response is a raw `Response` or already parsed `JSON`, so an additional Content-Type check isn't needed here. thoughts ?

petrjasek: would rather go for explicit here and introduce a param to `server.post` that would avoid calling that `json` but rather return the response as it is, and then we can just return the binary in all cases (when the response is ok and 200)

assets/server.ts
```diff
@@ -70,12 +80,12 @@ class Server {
      * @param {Object} data
      * @return {Promise}
      */
-    post(url: any, data: any, etag?: string) {
+    post(url: any, data: any, etag?: string, rawResponse = false) {
```
petrjasek: would rather put it in an options object:
```ts
post(url: any, data: any, etag?: string, {rawResponse}: {rawResponse: bool})
```
 so you see what is going on when looking at the usage:
```ts
await server.post(url, payload, undefined, {rawResponse: true});
```
and I would consider calling it `parseJson: false` or `processResponse: false`
so it's more obvious it's trying to avoid something

assets/server.ts
```diff
@@ -80,12 +82,17 @@ class Server {
      * @param {Object} data
      * @return {Promise}
      */
-    post(url: any, data: any, etag?: string, rawResponse = false) {
+    post(
+        url: any, 
+        data: any, 
+        etag?: string, 
+        config: { parseJson?: boolean } = {parseJson: true}
```
petrjasek: I would call it `options` in general, we use `config` for the client/server config

petrjasek: and seems like you could define an interface and use it for both post and checkStatus

devketanpro: as we already use `options` here as a function https://github.com/superdesk/newsroom-core/pull/1214/files/5171c7af00ee55da154aedd46bd58b0b3a996361..a24ad9e7e4d1cc02e91057fb1877d056e761a086#diff-2fb41b398d7dae3e0e795294a461513a587ea451bd7f36ab6887f81978848e10R91.
that's why I use diff name

petrjasek: I see, ok np

assets/server.ts
```diff
@@ -3,22 +3,38 @@ const defaultOptions = {
...
-    } else if (response.status >= 200 && response.status < 300) {
+    }
+
+    if (response.status >= 200 && response.status < 300) {
+        if (!parseJson) {
+            return Promise.resolve(response);
+        }
+
         const contentType = response.headers.get('Content-Type');
         if (contentType && contentType.includes('application/json')) {
```
petrjasek: would be enough to do `if (parseJson === true && ...)` instead of adding additional if condition

petrjasek (approved):

PR #1215, opened 2024-12-18
drop support for python 3.8

petrjasek:
and update github actions setup

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
               - "6379:6379"
 
       mongo:
-          image: mongo:4
+          image: mongo:6
           ports:
               - "27017:27017"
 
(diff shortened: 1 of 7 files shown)


.github/workflows/tests.yml
```diff
@@ -10,7 +10,25 @@ jobs:
...
+        python-version: ['3.10']
+
+    services:
+      redis:
+          image: redis:alpine
+          ports:
+              - "6379:6379"
+
+      mongo:
+          image: mongo:4
```
MarkLark86: I believe Mongo v4 is already EOL. Do we support Mongo v6 in `develop` (I know we do in async branch, just not sure about the PyMongo version we're using here)

petrjasek: seems working with 6

MarkLark86 (approved):

PR #1232, opened 2025-02-05
Fix: planning items with no related events the ‘related events' section in preview displays a constant loading circle [STT-137]

devketanpro:
### Checklist

Resolves: #[STT-137] [STT-138]

[STT-137]: https://sofab.atlassian.net/browse/STT-137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[STT-138]: https://sofab.atlassian.net/browse/STT-138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/agenda/components/AgendaPreviewEvent.tsx
```diff
@@ -129,6 +129,10 @@ class AgendaPreviewEventComponent extends React.Component<AgendaPreviewEventProp
     render() {
         const {itemsById, eventIds} = this.props;
 
+        if (!eventIds || eventIds.length === 0) {
```
MarkLark86: **1. Use the same logic to determine list of Event IDs**
* `eventIds == null || eventIds.length == 0`
* `!eventIds || eventIds.length === 0`
* And:
https://github.com/superdesk/newsroom-core/blob/2eb06ef775125ef36fe3ddc57c16fdf262b7ac09/assets/agenda/components/AgendaPreviewEvent.tsx#L146-L149

It would be good if we can have a util function that does this, so the logic is all the same (making sure all items are in the list).

**2. Explicitly tell the user there are "No Related Events"**
If there are no related Events, it might be good to indicate that to the end user. Maybe placing a label "No Related Events" or something like that.

MarkLark86 (approved):

PR #1235, opened 2025-02-10
Create new Management Async API in core [CPCN-1021]

devketanpro:
### Checklist

Resolves: #[issue-number]


features/mgmt_api/environment.py
```diff
@@ -0,0 +1,68 @@
...
+
+def before_scenario(context, scenario):
+    if "skip" in scenario.tags:
+        scenario.skip("Marked with @skip")
+        return
+
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
```
MarkLark86: Creating a loop here should not be needed. Check how it is currently working with the [NewsAPI tests](https://github.com/superdesk/newsroom-core/blob/async/newsroom/tests/news_api/environment.py#L38).

You will also need to log the exception and re-raise it, so it gets printed to the console

features/mgmt_api/environment.py
```diff
@@ -0,0 +1,68 @@
...
+    config = {
+        "BEHAVE": True,
+        "CORE_APPS": CORE_APPS,
+        "MODULES": MODULES,
+        "INSTALLED_APPS": [],
+        "ELASTICSEARCH_FORCE_REFRESH": True,
+        "MGMT_API_ENABLED": True,
+        "AUTH_SERVER_SHARED_SECRET": "test-secret",
+        "CACHE_TYPE": "null",
+        "ASYNC_AUTH_CLASS": "newsroom.mgmt_api.auth.jwt:JWTTokenAuth",
```
MarkLark86: This shouldn't be needed here, as it should be defined in the Management API default settings

newsroom/mgmt_api/factory.py
```diff
@@ -0,0 +1,126 @@
...
+    def __init__(self, import_name=__package__, config=None, **kwargs):
+        if not hasattr(self, "settings"):
+            self.settings = Config(".")
+
+        if config and config.get("BEHAVE"):
+            # ``superdesk.tests.update_config`` adds ``planning`` to ``INSTALLED_APPS``
+            # So if we're running behave tests, reset this config here
+            config["INSTALLED_APPS"] = []
+
+        self.wsgi = self
```
MarkLark86, line 44: What does this line do?

newsroom/mgmt_api/utils.py
```diff
@@ -0,0 +1,36 @@
+import bson
```
MarkLark86: Looks like this file is not used yet, would be good to keep this initial PR minimal and add this code later in another task

newsroom/web/default_settings.py
```diff
@@ -144,6 +144,7 @@
     "newsroom.email_templates",
     "newsroom.search",
     "newsroom.notifications.commands",
+    "newsroom.mgmt_api",
```
MarkLark86: This shouldn't be needed. CORE_APPS is the old way, and this new module does not provide any `init_app`. So I believe this can be safely removed from this line

MarkLark86 (commented):
Looks good. Just a couple tiny things

features/mgmt_api/environment.py
MarkLark86: The `ASYNC_AUTH_CLASS` config I'm talking about here

newsroom/mgmt_api/factory.py
devketanpro: It ensures that `self.wsgi` references the app instance itself, allowing code (like [config_oauth](https://github.com/superdesk/newsroom-core/blob/async/newsroom/auth_server/oauth2.py#L65-L80)) to access `app.wsgi.config`. so this maintains compatibility where wsgi is expected.

MarkLark86 (approved):

PR #1237, opened 2025-02-13
Upgrade resources and services to async [CPCN-1022]

devketanpro:
### Purpose
Upgrade and override core modules (Companies, Navigations, Products, Topics, Users) for MGMT api

### What has changed
Implemented custom CP modules for MGMT:
Companies (added country field), Navigations, Products, Topics, Users.

Resolves: #[CPCN-1022]

### Currently investigating into the following issues:
**Users search (case-insensitive):** - `FIXED`
Searching for users by email does not return results.
**Authentication in resources:**  `FIXED`
Disabling `auth=False`  in `RestEndpointConfig` results in a 500 authentication error.

[CPCN-1022]: https://sofab.atlassian.net/browse/CPCN-1022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/mgmt_api/environment.py
+++ b/features/mgmt_api/environment.py
@@ -2,7 +2,7 @@ import asyncio
 from superdesk.tests.environment import setup_before_all, setup_before_scenario
 from newsroom.auth_server.oauth2 import generate_jwt_token, config_oauth
 from newsroom.mgmt_api.app import get_app as _get_app
-from newsroom.mgmt_api.default_settings import CORE_APPS, MODULES
+from newsroom.mgmt_api.default_settings import CORE_APPS, MODULES, ASYNC_AUTH_CLASS, URL_PREFIX
 from superdesk.tests import setup as setup_app
 import logging
 
@@ -54,7 +54,8 @@ async def before_scenario_async(context, scenario):
         "MGMT_API_ENABLED": True,
         "AUTH_SERVER_SHARED_SECRET": "test-secret",
         "CACHE_TYPE": "null",
-        "ASYNC_AUTH_CLASS": "newsroom.mgmt_api.auth:JWTTokenAuth",
+        "ASYNC_AUTH_CLASS": ASYNC_AUTH_CLASS,
+        "URL_PREFIX": URL_PREFIX,
     }
 
     context.app = get_app(config=config)
--- a/features/mgmt_api/environment.py
+++ b/features/mgmt_api/environment.py
@@ -54,6 +54,7 @@ async def before_scenario_async(context, scenario):
         "MGMT_API_ENABLED": True,
         "AUTH_SERVER_SHARED_SECRET": "test-secret",
         "CACHE_TYPE": "null",
+        "ASYNC_AUTH_CLASS": "newsroom.mgmt_api.auth:JWTTokenAuth",
     }
 
     context.app = get_app(config=config)
--- a/features/mgmt_api/mgmt_api_products.feature
+++ b/features/mgmt_api/mgmt_api_products.feature
@@ -80,5 +80,5 @@ Feature: Management API - Products
         {"_items": []}
         """
 
-        When we delete "/products/#products._id#"
+        When we delete this "/products/#products._id#"
         Then we get response code 204
--- a/features/mgmt_api/mgmt_api_products.feature
+++ b/features/mgmt_api/mgmt_api_products.feature
@@ -31,13 +31,54 @@ Feature: Management API - Products
         Then we get response code 200
 
         When we post to this "companies/#companies._id#/products"
-            """
-            [
-                {
-                    "product": "#products._id#",
-                    "seats": 5,
-                    "link": true
-                }
-            ]
-            """
-            Then we get response code 201
+        """
+        [
+            {
+                "product": "#products._id#",
+                "seats": 5,
+                "link": true
+            }
+        ]
+        """
+        Then we get response code 201
+
+        When we get "companies/#companies._id#/products"
+        Then we get existing resource
+        """
+        {"_items": [
+            {
+                "name": "A fishy Product",
+                "description": "new description",
+                "query": "fish",
+                "product_type": "agenda",
+                "seats": 5
+            }
+        ]}
+        """
+
+        When we get "companies/#companies._id#"
+        Then we get existing resource
+        """
+        {"products": [
+            {"_id": "#products._id#", "section": "agenda", "seats": 5}
+        ]}
+        """
+
+        When we post to this "companies/#companies._id#/products"
+        """
+        [
+            {
+                "product": "#products._id#",
+                "link": false
+            }
+        ]
+        """
+        Then we get response code 201
+        When we get "companies/#companies._id#/products"
+        Then we get existing resource
+        """
+        {"_items": []}
+        """
+
+        When we delete "/products/#products._id#"
+        Then we get response code 204
--- a/features/mgmt_api/mgmt_api_products.feature
+++ b/features/mgmt_api/mgmt_api_products.feature
@@ -1,0 +1,84 @@
+Feature: Management API - Products
+    Scenario: Get company products
+        When we post to this "/companies"
+        """
+        [{"name": "zzz company"}]
+        """
+        Then we get response code 201
+
+        When we post to this "/products"
+        """
+        [{
+            "name": "A fishy Product",
+            "description": "a product for those interested in fish",
+            "query": "fish",
+            "product_type": "agenda"
+        }]
+        """
+        Then we get response code 201
+
+        When we get "/products"
+        Then we get existing resource
+        """
+        {"_items": [
+            {"name": "A fishy Product"}
+        ]}
+        """
+        When we patch "/products/#products._id#"
+        """
+        {"description": "new description"}
+        """
+        Then we get response code 200
+
+       When we post to this "companies/#companies._id#/products"
+        """
+        [
+            {
+                "product": "#products._id#",
+                "seats": 5,
+                "link": true
+            }
+        ]
+        """
+        Then we get response code 201
+
+        When we get "companies/#companies._id#/products"
+        Then we get existing resource
+        """
+        {"_items": [
+            {
+                "name": "A fishy Product",
+                "description": "new description",
+                "query": "fish",
+                "product_type": "agenda",
+                "seats": 5
+            }
+        ]}
+        """
+
+        When we get "companies/#companies._id#"
+        Then we get existing resource
+        """
+        {"products": [
+            {"_id": "#products._id#", "section": "agenda", "seats": 5}
+        ]}
+        """
+
+        When we post to this "companies/#companies._id#/products"
+        """
+        [
+            {
+                "product": "#products._id#",
+                "link": false
+            }
+        ]
+        """
+        Then we get response code 201
+        When we get "companies/#companies._id#/products"
+        Then we get existing resource
+        """
+        {"_items": []}
+        """
+
+        When we delete "/products/#products._id#"
+        Then we get response code 204
--- a/features/mgmt_api/mgmt_api_products.feature
+++ b/features/mgmt_api/mgmt_api_products.feature
@@ -24,61 +24,20 @@ Feature: Management API - Products
             {"name": "A fishy Product"}
         ]}
         """
-        When we patch "/products/#products._id#"
+        When we patch to this "/products/#products._id#"
         """
         {"description": "new description"}
         """
         Then we get response code 200
 
-       When we post to this "companies/#companies._id#/products"
-        """
-        [
-            {
-                "product": "#products._id#",
-                "seats": 5,
-                "link": true
-            }
-        ]
-        """
-        Then we get response code 201
-
-        When we get "companies/#companies._id#/products"
-        Then we get existing resource
-        """
-        {"_items": [
-            {
-                "name": "A fishy Product",
-                "description": "new description",
-                "query": "fish",
-                "product_type": "agenda",
-                "seats": 5
-            }
-        ]}
-        """
-
-        When we get "companies/#companies._id#"
-        Then we get existing resource
-        """
-        {"products": [
-            {"_id": "#products._id#", "section": "agenda", "seats": 5}
-        ]}
-        """
-
         When we post to this "companies/#companies._id#/products"
-        """
-        [
-            {
-                "product": "#products._id#",
-                "link": false
-            }
-        ]
-        """
-        Then we get response code 201
-        When we get "companies/#companies._id#/products"
-        Then we get existing resource
-        """
-        {"_items": []}
-        """
-
-        When we delete "/products/#products._id#"
-        Then we get response code 204
+            """
+            [
+                {
+                    "product": "#products._id#",
+                    "seats": 5,
+                    "link": true
+                }
+            ]
+            """
+            Then we get response code 201
--- a/features/mgmt_api/mgmt_api_topics.feature
+++ b/features/mgmt_api/mgmt_api_topics.feature
@@ -2,13 +2,12 @@ Feature: Management API - Topics
     Background: Setup companies and users
         When we post to this "/companies"
         """
-        [{"_id":"619277ef8bbbbfac6034aab4","name": "zzz company"}]
+        [{"name": "zzz company"}]
         """
         Then we get response code 201
         When we post to this "/users"
         """
         {   
-            "_id":"619277ef8bbbbfac6034aab7",
             "first_name": "Foo",
             "last_name": "Bar",
             "email": "foo@bar.org",
@@ -96,26 +95,26 @@ Feature: Management API - Topics
         """
         [
             {   
-                "_id":"619277ef8bbbbfac6034aac1",
                 "name": "test",
                 "section": "wire"
             }
         ]
         """
-        Given newsroom "topics"
+        When we post to this "/topics"
         """
         [
             {
                 "label": "topic1",
-                "company": "619277ef8bbbbfac6034aab4",
+                "company": "#companies._id#",
                 "topic_type": "wire",
                 "query": "topic1",
                 "is_global": true,
-                "user": "619277ef8bbbbfac6034aab7",
-                "folder": "619277ef8bbbbfac6034aac1"
+                "user": "#users._id#",
+                "folder": "#topic_folders._id#"
             }
         ]
         """
+        Then we get response code 201
         When we get "/topics"
         Then we get list with 1 items
 
@@ -137,7 +136,6 @@ Feature: Management API - Topics
         When we post to this "/navigations"
         """
         {   
-            "_id": "67ad7453e5baa77730e925dd",
             "name": "navigation1",
             "description": "navigation1",
             "order": 1
@@ -154,7 +152,7 @@ Feature: Management API - Topics
             "query": "topic1",
             "is_global": true,
             "user": "#users._id#",
-            "navigation": ["67ad7453e5baa77730e925dd"]
+            "navigation": ["#navigations._id#"]
         }
         """
         Then we get response code 201
--- a/features/mgmt_api/mgmt_api_topics.feature
+++ b/features/mgmt_api/mgmt_api_topics.feature
@@ -1,0 +1,160 @@
+Feature: Management API - Topics
+    Background: Setup companies and users
+        When we post to this "/companies"
+        """
+        [{"_id":"619277ef8bbbbfac6034aab4","name": "zzz company"}]
+        """
+        Then we get response code 201
+        When we post to this "/users"
+        """
+        {   
+            "_id":"619277ef8bbbbfac6034aab7",
+            "first_name": "Foo",
+            "last_name": "Bar",
+            "email": "foo@bar.org",
+            "company": "#companies._id#",
+            "user_type": "public"
+        }
+        """
+        Then we get response code 201
+
+    Scenario: Create a topic
+        When we post to this "/topics"
+        """
+        {
+            "label": "topic1",
+            "company": "#companies._id#",
+            "topic_type": "wire",
+            "query": "topic1",
+            "is_global": true,
+            "user": "#users._id#",
+            "subscribers": [
+                {"user_id": "#users._id#"}
+            ]
+        }
+        """
+        Then we get response code 201
+        When we get "/topics"
+        Then we get existing resource
+        """
+        {
+        "_items" :
+            [
+                {
+                    "label": "topic1",
+                    "company": "#companies._id#",
+                    "topic_type": "wire",
+                    "query": "topic1",
+                    "is_global": true,
+                    "user": "#users._id#",
+                    "subscribers": [
+                        {"user_id": "#users._id#", "notification_type": "real-time"}
+                    ]
+                }
+            ]
+        }
+        """
+
+    Scenario: Delete a topic
+        When we post to this "/topics"
+        """
+        [{
+            "label": "topic1",
+            "company": "#companies._id#",
+            "topic_type": "wire",
+            "query": "topic1",
+            "is_global": true,
+            "user": "#users._id#"
+        }]
+        """
+        When we delete latest
+        Then we get response code 204
+
+    Scenario: Update a topic
+        When we post to this "/topics"
+        """
+        [{
+            "label": "topic1",
+            "company": "#companies._id#",
+            "topic_type": "wire",
+            "query": "topic1",
+            "is_global": true,
+            "user": "#users._id#"
+        }]
+        """
+        When we patch latest
+        """
+        {"label": "topic2"}
+        """
+        Then we get updated response
+        """
+        {"label": "topic2"}
+        """
+
+    Scenario: Topic with folder
+        Given newsroom "topic_folders"
+        """
+        [
+            {   
+                "_id":"619277ef8bbbbfac6034aac1",
+                "name": "test",
+                "section": "wire"
+            }
+        ]
+        """
+        Given newsroom "topics"
+        """
+        [
+            {
+                "label": "topic1",
+                "company": "619277ef8bbbbfac6034aab4",
+                "topic_type": "wire",
+                "query": "topic1",
+                "is_global": true,
+                "user": "619277ef8bbbbfac6034aab7",
+                "folder": "619277ef8bbbbfac6034aac1"
+            }
+        ]
+        """
+        When we get "/topics"
+        Then we get list with 1 items
+
+    Scenario: Validate if navigation exists
+        When we post to this "/topics"
+        """
+        {
+            "label": "topic1",
+            "company": "#companies._id#",
+            "topic_type": "wire",
+            "query": "topic1",
+            "is_global": true,
+            "user": "#users._id#",
+            "navigation": ["619277ef8bbbbfac6034aab7"]
+        }
+        """
+        Then we get response code 500
+
+        When we post to this "/navigations"
+        """
+        {   
+            "_id": "67ad7453e5baa77730e925dd",
+            "name": "navigation1",
+            "description": "navigation1",
+            "order": 1
+        }
+        """
+        Then we get response code 201
+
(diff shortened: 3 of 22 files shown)


features/mgmt_api/steps/steps.py
```diff
@@ -9,3 +19,47 @@ def given_empty_auth_token(context):
         context.headers = []
 
     context.headers = [header for header in context.headers if header[0] != "Authorization"]
+
+
+async def store_placeholder(context, url):
```
eos87: would be good to add some docstrings explaining the "why" of this method. Plus, types won't hurt either :)

newsroom/mgmt_api/companies.py
eos87, line 1: can you please add the types to code is this module (params and return types)? Also, it would be good to split it into smaller modules, perhaps one for CPCompanies and another  for CompanyProducts.

newsroom/mgmt_api/topics.py
eos87, line 1: same regarding the types

newsroom/mgmt_api/users.py
```diff
@@ -0,0 +1,63 @@
...
+from newsroom.types import UserResourceModel
+from superdesk.errors import SuperdeskApiError
+from typing_extensions import override
+
+from newsroom.mgmt_api.utils import validate_product_refs
+from superdesk.core.resources import ResourceConfig, MongoResourceConfig, RestEndpointConfig
+from content_api import MONGO_PREFIX
+from superdesk.core.module import Module
+from typing import Any
+from newsroom.core import get_current_wsgi_app
```
eos87: nit: can you please sort the imports?

newsroom/mgmt_api/users.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class CPUsersResource(UserResourceModel):
+    pass
+
+
+class CPUsersService(UsersService):
+    @override
+    async def on_create(self, docs):
```
eos87: types are missing here too

newsroom/mgmt_api/utils.py
eos87, line 1: same as above, types and please sort the imports so our code look neat.

eos87 (changes requested):

features/mgmt_api/mgmt_api_topics.feature
```diff
@@ -0,0 +1,160 @@
+Feature: Management API - Topics
+    Background: Setup companies and users
+        When we post to this "/companies"
+        """
+        [{"_id":"619277ef8bbbbfac6034aab4","name": "zzz company"}]
```
MarkLark86: Why have you included `_id` with the create company and user requests (I don't see that field supplied in behave tests from CP repo)

features/mgmt_api/mgmt_api_topics.feature
```diff
@@ -0,0 +1,160 @@
...
+        """
+        {"label": "topic2"}
+        """
+
+    Scenario: Topic with folder
+        Given newsroom "topic_folders"
+        """
+        [
+            {   
+                "_id":"619277ef8bbbbfac6034aac1",
```
MarkLark86: And again here with the `_id` field

features/mgmt_api/mgmt_api_topics.feature
```diff
@@ -0,0 +1,160 @@
...
+                "section": "wire"
+            }
+        ]
+        """
+        Given newsroom "topics"
+        """
+        [
+            {
+                "label": "topic1",
+                "company": "619277ef8bbbbfac6034aab4",
```
MarkLark86: Here you should be using `"#companies._id#"`, `"#users._id#"` and `"#topic_folders._id#"` for the IDs

features/mgmt_api/mgmt_api_topics.feature
```diff
@@ -0,0 +1,160 @@
...
+            "label": "topic1",
+            "company": "#companies._id#",
+            "topic_type": "wire",
+            "query": "topic1",
+            "is_global": true,
+            "user": "#users._id#",
+            "navigation": ["619277ef8bbbbfac6034aab7"]
+        }
+        """
+        Then we get response code 500
```
MarkLark86: The original API returned a response code of 400. Validation may not be working as intended here if it's raising an internal error

features/mgmt_api/mgmt_api_topics.feature
```diff
@@ -0,0 +1,160 @@
...
+
+        When we post to this "/navigations"
+        """
+        {   
+            "_id": "67ad7453e5baa77730e925dd",
```
MarkLark86: Again here with the `_id` field

features/mgmt_api/mgmt_api_topics.feature
```diff
@@ -0,0 +1,160 @@
...
+        When we post to this "/topics"
+        """
+        {
+            "label": "topic1",
+            "company": "#companies._id#",
+            "topic_type": "wire",
+            "query": "topic1",
+            "is_global": true,
+            "user": "#users._id#",
+            "navigation": ["67ad7453e5baa77730e925dd"]
```
MarkLark86: And here you should be able to use `"#navigations._id#"`

features/mgmt_api/mgmt_api_products.feature
```diff
@@ -0,0 +1,43 @@
...
+        ]}
+        """
+        When we patch to this "/products/#products._id#"
+        """
+        {"description": "new description"}
+        """
+        Then we get response code 200
+
+        When we post to this "companies/#companies._id#/products"
+            """
```
MarkLark86: The indents here are incorrect

features/mgmt_api/mgmt_api_products.feature
```diff
@@ -0,0 +1,43 @@
...
+            [
+                {
+                    "product": "#products._id#",
+                    "seats": 5,
+                    "link": true
+                }
+            ]
+            """
+            Then we get response code 201
```
MarkLark86: There are more tests in from CP repo that need to be here too:
https://github.com/superdesk/newsroom-app-cp/blob/cn-develop/server/features/mgmt_api_products.feature#L47C9-L86

MarkLark86:
@devketanpro For Users search (case-insensitive) - Searching for users by email does not return results.
You need to implement a custom REST Endpoint class to implement the search changes seen in the [original](https://github.com/superdesk/newsroom-app-cp/blob/cn-develop/server/cp/mgmt_api/users.py#L59) implementation - namely the cursor.collation(Collation(locale="en", strength=1)).

You do this by creating a custom ResourceRestEndpoints class - [example from superdesk-core](https://github.com/MarkLark86/superdesk-core/blob/SDESK-7466/superdesk/vocabularies_async/rest_api.py)
Then use that in the [resource config](https://github.com/MarkLark86/superdesk-core/blob/SDESK-7466/superdesk/vocabularies_async/module.py#L32).

In that custom class, you would override the [search_items](https://github.com/MarkLark86/superdesk-core/blob/SDESK-7466/superdesk/core/resources/resource_rest_endpoints.py#L560) method, and add there the MongoDB cursor changes (cursor.collation(Collation(locale="en", strength=1)))

-------
Actually, could you first try the MongoDB resource options [here](https://github.com/superdesk/newsroom-core/blob/e9f1bae22cdb06e1837c596e2bb6e2298659f9dd/newsroom/mgmt_api/users.py#L57).
Add:
```python
users_resource_config = ResourceConfig(
    name="users",
    data_class=CPUsersResource,
    service=CPUsersService,
    mongo=MongoResourceConfig(
        prefix=MONGO_PREFIX,
        collation={"locale": "en", "strength": 1}
    ),
    rest_endpoints=RestEndpointConfig(auth=False),
)
```

And see how that goes, otherwise we'll have to do it via the REST API

features/mgmt_api/mgmt_api_topics.feature
devketanpro: I'm using `_id` explicitly because when we use context IDs like `#companies._id#` or` #users._id#`, Pydantic validation fails with an error stating that `#users._id#` `is not a valid ObjectId—it must be a 12-byte input or a 24-character hex string.`

features/mgmt_api/mgmt_api_topics.feature
devketanpro: yep, but when validations fails, we get code `500` see
image

newsroom/mgmt_api/companies.py
```diff
@@ -0,0 +1,63 @@
...
+class CPCompaniesResource(CompanyResource):
+    """
+    CP Companies Schema
+    """
+
+    country: Optional[str] = "CAN"
+
+
+class CPCompaniesService(CompanyService):
+    async def on_create(self, docs: List[CPCompaniesResource]):
```
MarkLark86: `get_errors_company` and `validate_product_refs` are no longer required in `on_create` nor `on_update`, as they're covered in the CompanyResource model validation. (this includes `name`, `allowed_ip_list` as well as `products` IDs pointing to an existing product).

newsroom/mgmt_api/companies.py
```diff
@@ -0,0 +1,63 @@
...
+from newsroom.core import get_current_wsgi_app
+from .utils import validate_product_refs, get_errors_company
+
+
+class CPCompaniesResource(CompanyResource):
+    """
+    CP Companies Schema
+    """
+
+    country: Optional[str] = "CAN"
```
MarkLark86: Not sure it's worth implementing a custom CompaniesResource if all we're achieving is a default value for `country` (which exists in the original resource model).

Instead I would recommend to do this in the `on_create` and `on_update` methods on the custom service.

newsroom/mgmt_api/companies_products.py
```diff
@@ -0,0 +1,86 @@
...
+from newsroom import MONGO_PREFIX
+from newsroom.auth.utils import get_current_request
+from newsroom.mgmt_api.products import CPProductsService
+from newsroom.mgmt_api.companies import CPCompaniesService, company_resource_config
+from newsroom.core.resources import NewshubResourceModel, NewshubAsyncResourceService
+from newsroom.products.views import get_product_ref
+
+
+class CompanyProductsResource(NewshubResourceModel):
+    product: Annotated[Optional[str], validate_data_relation_async("products", convert_to_objectid=True)]
```
MarkLark86: Instead of using `Optional` (which requires importing it from the `typing` module), you can instead use pure Python types. Here it would be `str | None`.

newsroom/mgmt_api/companies_products.py
```diff
@@ -0,0 +1,86 @@
...
+                continue  # Skip if no product is provided
+
+            link = doc.link
+            product_data_cursor = await CPProductsService().find_by_id(ObjectId(product_id))
+            if not product_data_cursor:
+                continue
+
+            product = product_data_cursor
+
+            print(product, "\n\n\n\n\n\n\n\n\n")
```
MarkLark86: Remove debug code

newsroom/mgmt_api/companies_products.py
```diff
@@ -0,0 +1,86 @@
...
+    request = get_current_request()
+    company = await CPCompaniesService().find_by_id(ObjectId(request.get_view_args("companies")))
+    return company.to_dict()
+
+
+def get_company_products(company: Dict[str, Any]) -> List[dict]:
+    return company.get("products") or []
+
+
+class CompanyProductsService(NewshubAsyncResourceService[CompanyProductsResource]):
```
MarkLark86: This is different to a parent/child resource relationship, as the requests in this service updates the parent item (the company) and not the products resource.

Instead we should probably implement this as pure endpoints and not a resource/service.
Implementing endpoints for these two (as outlined in the swagger [docs](https://github.com/superdesk/newsroom-app-cp/blob/cn-develop/server/theme/swagger.yaml#L494)):
* GET companies/<regex("[a-f0-9]{24}"):company_id>/products
* POST companies/<regex("[a-f0-9]{24}"):company_id>/products

@petrjasek In the swagger docs I see these params for the GET, do we need to support those:

https://github.com/superdesk/newsroom-app-cp/blob/eab5e9f350dfd15b39d9e3871a731f28004829fc/server/theme/swagger.yaml#L501-L506

MarkLark86: I also don't see any behave tests for these endpoints, neither here nor in the CP repo.
It would be good to add behave tests for these 2 endpoints

MarkLark86:
```python
from pydantic import BaseModel

from superdesk.core.types import Request, Response
from superdesk.core.resources import fields
from superdesk.core.web import EndpointGroup

from newsroom.types import CompanyResource, CompanyProduct

company_products_endpoints = EndpointGroup("company_products", __name__)


class CompanyProductRouteArguments(BaseModel):
    company_id: fields.ObjectId

    async def get_company(self) -> CompanyResource | None:
        return CompanyResource.get_service().find_by_id(self.company_id)


@wire_endpoints.endpoint("/companies/<regex("[a-f0-9]{24}"):company_id>/products", methods=["GET"])
async def get_company_products(args: CompanyProductRouteArguments, params: None, request: Request) -> Response:
    company = await args.get_company()
    # TODO-ASYNC: implement this endpoint


@wire_endpoints.endpoint("/companies/<regex("[a-f0-9]{24}"):company_id>/products", methods=["POST"])
async def update_company_products(args: CompanyProductRouteArguments, params: None, request: Request) -> Response:
    company = await args.get_company()
    product_links_json = await request.get_json()
    product_links: list[] = [CompanyProduct.from_dict(link) for link in product_links_json]
    # TODO-ASYNC: implement this endpoint


module = Module(
    name="newsroom.mgmt_api.companies_products",
    endpoints=[company_products_endpoints],
)
```

MarkLark86: And then you would implement all the functionality from within https://github.com/superdesk/newsroom-app-cp/blob/cn-develop/server/cp/mgmt_api/companies.py

newsroom/mgmt_api/navigations.py
```diff
@@ -0,0 +1,28 @@
+from newsroom import MONGO_PREFIX
+from newsroom.navigations import NavigationsService
+from newsroom.types import NavigationModel
+
+from superdesk.core.module import Module
+from superdesk.core.resources import MongoResourceConfig, ResourceConfig, RestEndpointConfig
+
+
+class CPNavigationModel(NavigationModel):
```
MarkLark86: No need to use new classes here, you can directly use the ones from core

newsroom/mgmt_api/products.py
```diff
@@ -0,0 +1,29 @@
+from newsroom import MONGO_PREFIX
+from newsroom.products.service import ProductsService
+from newsroom.types import ProductResourceModel
+
+from superdesk.core.module import Module
+from superdesk.core.resources import MongoResourceConfig, ResourceConfig, RestEndpointConfig
+
+
+class CPProductResourceModel(ProductResourceModel):
```
MarkLark86: Same here, you can use the ones from core directly

newsroom/mgmt_api/topics.py
```diff
@@ -0,0 +1,69 @@
...
+        for doc in docs:
+            app.cache.set(str(doc.id), doc)
+
+    async def on_update(self, updates: Dict[str, Any], original: TopicResourceModel) -> None:
+        await super().on_update(updates, original)
+        app = get_current_wsgi_app()
+        app.cache.delete(str(original.id))
+
+
+class CPFolderResourceModel(TopicFolderResourceModel):
```
MarkLark86: And here, you should be able to use the ones from core types

newsroom/mgmt_api/users.py
```diff
@@ -0,0 +1,74 @@
...
+                message = "Company is required if user type is not administrator."
+                raise SuperdeskApiError.badRequestError(message=message, payload=message)
+            locale = doc.locale
+            if locale and locale not in app.config["LANGUAGES"]:
+                message = "Locale is not in configured list of locales."
+                raise SuperdeskApiError.badRequestError(message=message, payload=message)
+            if doc.company:
+                doc.company = ObjectId(doc.company)
+            if doc.products:
+                doc.products = await validate_product_refs(doc.products)
```
MarkLark86, line 35: Again here about validation, this is already handled in the `CompanyProduct` validation annotation

newsroom/mgmt_api/users.py
```diff
@@ -0,0 +1,74 @@
...
+class CPUsersService(UsersService):
+    @override
+    async def on_create(self, docs: List[UserResourceModel]) -> None:
+        await super().on_create(docs)
+        for doc in docs:
+            if doc.user_type != "administrator" and not doc.company:
+                message = "Company is required if user type is not administrator."
+                raise SuperdeskApiError.badRequestError(message=message, payload=message)
+            locale = doc.locale
+            if locale and locale not in app.config["LANGUAGES"]:
```
MarkLark86: This validation should ideally be placed on the core `UserResourceModel`, to be sure all Users no matter where they're created have this validation as well.

```python
# file: newsroom/types/users.py
from pydantic import field_validator

from superdesk.core import get_app_config


class UserResourceModel(NewshubResourceModel):
    @field_validator("locale", mode="before"):
    @classmethod
    def validate_locale(cls, value: str | None) -> str | None:
        if value is not None and value not in get_app_config("LANGUAGES", []):
            raise SuperdeskApiError.badRequestError("Locale is not in configured list of locales")
        return value
```

Or something along those lines.

MarkLark86: And probably same for `user_type` and `not company`, but using the `@model_validator(mode="before")` instead. See [Agenda](https://github.com/superdesk/newsroom-core/blob/async/newsroom/types/agenda.py#L263) for an example

newsroom/mgmt_api/users.py
```diff
@@ -0,0 +1,74 @@
...
+                raise SuperdeskApiError.badRequestError(message=message, payload=message)
+            if doc.company:
+                doc.company = ObjectId(doc.company)
+            if doc.products:
+                doc.products = await validate_product_refs(doc.products)
+
+    @override
+    async def on_update(self, updates: Dict[str, Any], original: UserResourceModel) -> None:
+        if updates.get("products"):
+            updates["products"] = await validate_product_refs(updates["products"])
```
MarkLark86, line 40: Not needed anymore

newsroom/mgmt_api/users.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @override
+    async def on_updated(self, updates: Dict[str, Any], original: UserResourceModel) -> None:
```
MarkLark86, line 45: Empty body function, is there a reason for this?

newsroom/mgmt_api/users.py
```diff
@@ -0,0 +1,74 @@
...
+        pass
+
+    @override
+    async def on_delete(self, doc: UserResourceModel) -> None:
+        app = get_current_wsgi_app()
```
MarkLark86: Please go over all the `on_` methods on services, and make sure to call `await super().on_`, as the base service does some things in them (such as executing signal listeners for them).

newsroom/mgmt_api/users.py
devketanpro: The parent `on_updated` method include validations such as checking the current user. However, in this API, those validations are not relevant, so this method is intentionally left empty

newsroom/mgmt_api/users.py
devketanpro: If we call `await super().on_delete()`, we might run into irrelevant validations that don't apply to this service. However, if signals are necessary, we could explicitly trigger them here instead. What do you think?

newsroom/mgmt_api/companies.py
devketanpro: I had removed `validate_product_refs` earlier, but without it, the product wasn’t getting updated correctly. This function also returns the updated product, which is necessary for the process. Removing it caused test cases in this scenario to fail: [testcases](https://github.com/superdesk/newsroom-core/pull/1237/files/9e9df9c559c5d7aeec91c0f8b6a310fa23114006#diff-98032a5c76317026e32a4eac91dabe3a1cb15a095c5193f47c89eaf06ff09421R80-R170)

features/mgmt_api/mgmt_api_topics.feature
MarkLark86: What is the value then for those context IDs, if the ID is invalid from the context then we should fix the context value

newsroom/mgmt_api/companies.py
MarkLark86: Fair enough. You should still be able to remove `get_errors_company` function though

newsroom/mgmt_api/users.py
MarkLark86: You can skip the parent class' implementation and use it's parent method instead.

i.e.
```python
await super(UsersService, self).on_updated(updates, original)
```

This will skip `UsersService.on_updated` and instead use  `NewshubAsyncResourceService.on_updated`.
This can be used for all service methods where required

features/mgmt_api/mgmt_api_topics.feature
devketanpro: `67b5a149197debe592d9e914`

newsroom/mgmt_api/users.py
MarkLark86: For example:
```python
>>> class A:
...     def test(self):
...         print("A")
... 
>>> class B(A):
...     def test(self):
...         super().test()
...         print("B")
... 
>>> class C(B):
...     def test(self):
...         super(B, self).test()
...         print("C")
... 
>>> A().test()
A
>>> B().test()
A
B
>>> C().test()
A
C
>>>
```

MarkLark86:
Depends on https://github.com/superdesk/newsroom-core/pull/1240 to fix breaking change from superdesk-core

features/mgmt_api/environment.py
```diff
@@ -54,6 +54,7 @@ async def before_scenario_async(context, scenario):
         "MGMT_API_ENABLED": True,
         "AUTH_SERVER_SHARED_SECRET": "test-secret",
         "CACHE_TYPE": "null",
+        "ASYNC_AUTH_CLASS": "newsroom.mgmt_api.auth:JWTTokenAuth",
```
MarkLark86: Is this required, as it's set in the [default_settings](https://github.com/devketanpro/newsroom-core/blob/CPCN-1022/newsroom/mgmt_api/default_settings.py#L69)

newsroom/mgmt_api/companies_products.py
```diff
@@ -0,0 +1,107 @@
...
+from superdesk.core.types import Request, Response
+from superdesk.core.resources import fields
+from superdesk.core.web import EndpointGroup
+
+
+def get_company_products(company: CompanyResource) -> List[CompanyProduct]:
+    return company.products or []
+
+
+company_products_endpoints = EndpointGroup("company_products", __name__, url_prefix="/api")
```
MarkLark86: The `url_prefix` should not be required here, in fact we have a [config](https://github.com/superdesk/newsroom-core/blob/c4cbf76a5ca9dc0d68bde8b689f082e25042a9ec/newsroom/mgmt_api/default_settings.py#L27) for the prefix

newsroom/mgmt_api/companies_products.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class CompanyProductRouteArguments(BaseModel):
+    company_id: fields.ObjectId
+
+    async def get_company(self) -> CompanyResource | None:
+        return await CompanyResource.get_service().find_by_id(self.company_id)
+
+
+@company_products_endpoints.endpoint("/api/companies/<regex('[a-f0-9]{24}'):company_id>/products", methods=["POST"])
```
MarkLark86: If you remove the `/api/` from the URL it will be relative to that `URL_PREFIX` config when registered with the system.
`companies/<regex('[a-f0-9]{24}'):company_id>/products`

Any URL that starts with `/` is absolute, other's are relative to that config

newsroom/mgmt_api/companies_products.py
```diff
@@ -0,0 +1,107 @@
...
+
+        product_data = await ProductsService().find_by_id(ObjectId(product_id))
+        if not product_data:
+            continue
+
+        updated_products = [p for p in updated_products if p._id != product_data.id]
+
+        if doc.link:
+            updated_products.append(get_product_ref(product_data, doc.seats).to_dict())
+            pass
```
MarkLark86: `pass` statement takes no effect / not needed

newsroom/mgmt_api/companies_products.py
```diff
@@ -0,0 +1,107 @@
...
+    products_data = []
+    for product in company_products:
+        product_data = await ProductsService().find_by_id(ObjectId(product._id))
+        if product_data:
+            item = {
+                "name": product_data.name,
+                "description": product_data.description,
+                "query": product_data.query,
+                "product_type": product_data.product_type,
+                "seats": product.seats or 0,
```
MarkLark86, line 95: The `seats` attribute should be coming from the company, like in the [original implementation](https://github.com/superdesk/newsroom-app-cp/blob/cn-develop/server/cp/mgmt_api/companies.py#L136).

newsroom/mgmt_api/users.py
```diff
@@ -0,0 +1,67 @@
...
+    @override
+    async def on_updated(self, updates: Dict[str, Any], original: UserResourceModel) -> None:
+        # Skipping UsersService.on_updated as its validations are not needed for this API.
+        await super(UsersService, self).on_updated(updates, original)
+
+    @override
+    async def on_delete(self, doc: UserResourceModel) -> None:
+        get_current_wsgi_app().cache.delete(str(doc.id))
+
+    async def find(self, req: Request):
```
MarkLark86: Nice fix. :+1: Don't forget the return type

MarkLark86 (commented):
Looks good, just a couple tiny things left now then it will be good to merge

newsroom/mgmt_api/companies_products.py
devketanpro: I tried this earlier, but it wasn't registered under `/api/`, resulting in a 404 error as the requested URL was not found.

newsroom/mgmt_api/companies_products.py
devketanpro: that `product` is coming from the `company_product`  see  [here](https://github.com/superdesk/newsroom-core/pull/1237/files/c4cbf76a5ca9dc0d68bde8b689f082e25042a9ec#diff-a21486d9ea0628c77acc3acd5c5a9b7e1130938f23d53ed3a8e222f7dd321dbbR85) and [here](https://github.com/superdesk/newsroom-core/pull/1237/files/c4cbf76a5ca9dc0d68bde8b689f082e25042a9ec#diff-a21486d9ea0628c77acc3acd5c5a9b7e1130938f23d53ed3a8e222f7dd321dbbR88)

features/mgmt_api/environment.py
devketanpro: yep in the env it is required

newsroom/mgmt_api/companies_products.py
MarkLark86: Oh, whoops I misread the value name

newsroom/mgmt_api/companies_products.py
MarkLark86: If the `EndpointGroup.url_prefix` is None and this url does not start with `/`, then that config is used.
See [here](https://github.com/superdesk/superdesk-core/blob/async/superdesk/factory/app.py#L314) for the registration code.

Basically we need this to work, because that URL prefix is also controlled via an env. variable and we need to support it here too

devketanpro: my bad, you are right its working

MarkLark86 (approved):

PR #1240, opened 2025-02-19
fix: Update services after core changes to create response type and validation

MarkLark86:
### Purpose
There were breaking changes in superdesk-core resource services and rest apis. This PR fixes those

### What has changed
* Update dev-requirements.txt
* Update service.create implementation and usage throughout code and tests to use returned instances instead of IDs
* Fix validation issue with TopicFolder REST API, validation in core is returning format used by Superdesk where as Newshub is different - Hack for now, will come up with a better implementation for this: see https://github.com/superdesk/superdesk-core/blob/8ec622888bf16617817d7e08d73a61b0525d9244/superdesk/core/resources/resource_rest_endpoints.py#L380-L404 for the change in core

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -8,9 +8,9 @@ aiofiles==24.1.0
     # via
     #   quart
     #   quart-uploads
-aiohappyeyeballs==2.4.4
+aiohappyeyeballs==2.4.6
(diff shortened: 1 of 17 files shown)


devketanpro (approved):
looks good

PR #1243, opened 2025-02-26
fix circular import error [CPCN-1020]

devketanpro:
### Purpose
Fixing circular imports error

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/newsroom/notifications/utils.py
+++ b/newsroom/notifications/utils.py
@@ -12,7 +12,6 @@ from newsroom.exceptions import AuthorizationError
 from newsroom.auth.utils import get_user_id_from_request
 from newsroom.topics.topics_async import TopicService
 from .services import NotificationsService
-from newsroom.types import WireItem
 
 
 def user_notifications_lookup(user_id: ObjectId) -> dict[str, Any]:
@@ -59,6 +58,7 @@ async def get_notifications_with_items() -> dict[str, Any] | None:
     """
 
     from newsroom.agenda import AgendaSearchServiceAsync
+    from newsroom.wire import WireSearchServiceAsync
 
     try:
         user_id = get_user_id_from_request(None)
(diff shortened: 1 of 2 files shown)


newsroom/notifications/utils.py
```diff
@@ -59,6 +58,7 @@ async def get_notifications_with_items() -> dict[str, Any] | None:
     """
 
     from newsroom.agenda import AgendaSearchServiceAsync
+    from newsroom.wire import WireSearchServiceAsync
```
MarkLark86: Instead of moving the import, can you please import the type at the top of this file, for both Agenda and Wire, and use the [find_by_ids](https://superdesk.readthedocs.io/en/async/core/resource_management.html#superdesk.core.resources.service.AsyncResourceService.find_by_ids) method

newsroom/topics/views.py
```diff
@@ -37,6 +36,8 @@ async def get_topics(args: RouteArguments, params: None, request: Request) -> Re
 )
 async def post_topic(request: Request) -> Response:
     """Creates a user topic"""
+    # Import here to prevent circular imports
+    from newsroom.notifications import push_user_notification, push_company_notification
```
MarkLark86: This should be fixable in [newsroom/notifications/utils.py](https://github.com/superdesk/newsroom-core/blob/async/newsroom/notifications/utils.py#L13), by replacing `WireSearchServiceAsync` with the type, and using `find_by_ids`

MarkLark86 (changes requested):

newsroom/topics/views.py
devketanpro: I tried but it is not working

MarkLark86: I see, then it is something else as well causing the circular imports, we can leave as is for now then.
Ideally it would be good to fix these circular imports, as importing modules inside a function is not good practice and I would prefer we avoid where possible

MarkLark86 (approved):

PR #1247, opened 2025-03-20
FIX:Blank/Empty option appearing in the Location filter [CPCN-1031]

devketanpro:
### Purpose
 handle location like this 
```.json
{
    "name": "",
    "country": "Canada",
    "type": "state"
}
```

and exclude locations of "`killed`" agenda items

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/assets/agenda/components/LocationFilter.tsx
+++ b/assets/agenda/components/LocationFilter.tsx
@@ -337,9 +337,6 @@ export class LocationFilter extends React.Component<any, any> {
                 </button>
             );
         } else if (item.type === LOCATION_TYPE.STATE) {
-            if (!item.name){
-                return null;
-            }
             return (
                 <button
                     key={`state.${item.name}[${index}]`}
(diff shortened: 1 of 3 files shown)


assets/agenda/components/LocationFilter.tsx
```diff
@@ -337,6 +337,9 @@ export class LocationFilter extends React.Component<any, any> {
                 </button>
             );
         } else if (item.type === LOCATION_TYPE.STATE) {
+            if (!item.name){
```
petrjasek: can this be also done on the server?

devketanpro: yep

newsroom/agenda/views.py
```diff
@@ -453,13 +461,14 @@ def gen_agg_terms(field: str):
...
             country_name = country_bucket["key"]
             for state_bucket in country_bucket["states"]["buckets"]:
-                regions.append(
-                    {
-                        "name": state_bucket["key"],
-                        "country": country_name,
-                        "type": "state",
-                    }
-                )
+                if state_bucket["key"]:
```
petrjasek: can you try to add `exlude: [""],` to the `gen_agg_terms` method so it won't be returned from elastic,
[should be possible to filter it out there](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#_filtering_values_with_exact_values_2) and make it work for all fields

petrjasek (approved):

PR #1261, opened 2025-04-10
FIX: MGMT API: Cannot post a new topic - error 400 [CPCN-1060]

devketanpro:
Resolves: #[CPCN-1060]

[CPCN-1060]: https://sofab.atlassian.net/browse/CPCN-1060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/types/topics.py
+++ b/newsroom/types/topics.py
@@ -1,5 +1,5 @@
 from enum import Enum, unique
-from pydantic import Field, field_validator
+from pydantic import Field, model_validator
 from typing import Any, Annotated
 
 from superdesk.utc import utcnow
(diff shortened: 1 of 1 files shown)


MarkLark86:
@devketanpro Can you change the destination branch to `release/3.0` please, as that's where this fix is going into. Later I can merge that branch into `async`

newsroom/types/topics.py
```diff
@@ -48,6 +48,11 @@ class TopicResourceModel(NewshubResourceModel):
     folder: Annotated[ObjectIdField | None, validate_data_relation_async("topic_folders")] = None
     advanced: AdvancedSearchParams | None = None
 
+    @field_validator("created", mode="before")
+    @classmethod
+    def set_created(cls, value):
+        return utcnow()
```
MarkLark86: This will always set `created` to `utcnow`, even when updating a Topic. What you could possibly do instead is:

```python
from pydantic import Field, model_validator
from superdesk.utc import utcnow

class TopicResourceModel(NewshubResourceModel):
    ...

    @model_validator(mode="before"):
    @classmethod
    def parse_dict(cls, values) -> dict:
        if not values.get("_created"):
            # `_created` needs to be set otherwise there is a clash given `TopicResourceModel` and
            # the base `NewshubResourceModel` both have the same member (`created`).
            # Without this `created_filter` does not get converted/saved
            values["_created"] = utcnow()
```

MarkLark86 (approved):

PR #1281, opened 2025-05-29
fix printing on factcheck, market place and media releases

marwoodandrew:
### Purpose

Fix the print endpoint in factcheck and media releases and market place.

### What has changed

The mechanism to verify that the user had the right to print the selected item was broken, now uses the function to verify the item is accessible to the user.

### Steps to test

Try to print an item from the sections above, if you can then it is fixed.

### Checklist

- [x ] This pull request is not adding new forms that use redux
- [x ] This pull request is adding missing TypeScript types to modified code segments where it's easy to do so with confidence
- [x ] This pull request is replacing `lodash.get` with optional chaining for modified code segments

Resolves: #[issue-number]

### DIFF
--- a/newsroom/factcheck/views.py
+++ b/newsroom/factcheck/views.py
@@ -57,7 +57,7 @@ async def search():
 @blueprint.route("/bookmarks_factcheck")
 @login_required
 async def bookmarks():
-    data = get_view_data()
+    data = await get_view_data()
     data["bookmarks"] = True
     return await render_template("factcheck_bookmarks.html", data=data)
 
--- a/newsroom/factcheck/views.py
+++ b/newsroom/factcheck/views.py
@@ -57,7 +57,7 @@ async def search():
 @blueprint.route("/bookmarks_factcheck")
 @login_required
 async def bookmarks():
-    data = await get_view_data()
+    data = get_view_data()
     data["bookmarks"] = True
     return await render_template("factcheck_bookmarks.html", data=data)
 
--- a/newsroom/factcheck/views.py
+++ b/newsroom/factcheck/views.py
@@ -9,7 +9,7 @@ from newsroom.auth.utils import get_user_from_request, get_company_from_request
 from newsroom.formatters import get_formatters_id_and_names
 from newsroom.factcheck import blueprint
 from newsroom.decorator import login_required, section
-from newsroom.wire import WireSearchServiceAsync
+from .search import FactCheckSearchServiceAsync
 from newsroom.wire.views import (
     update_action_list,
     get_previous_versions,
@@ -32,7 +32,7 @@ async def get_view_data():
         "company": str(company.id) if company else None,
         "navigations": [],
         "formats": get_formatters_id_and_names(SectionEnum.WIRE),
-        "saved_items": await WireSearchServiceAsync().get_current_user_bookmarks_count(SectionEnum.FACTCHECK),
+        "saved_items": await FactCheckSearchServiceAsync().get_current_user_bookmarks_count(),
         "context": "factcheck",
         "ui_config": await ui_config_service.get_section_config("factcheck"),
     }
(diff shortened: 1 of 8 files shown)


newsroom/market_place/views.py
```diff
@@ -48,7 +48,7 @@ async def get_view_data():
         "topics": [t for t in topics if t.get("topic_type") == SECTION_ID],
         "navigations": navigations,
         "formats": get_formatters_id_and_names(SectionEnum.WIRE),
-        "saved_items": await WireSearchServiceAsync().get_current_user_bookmarks_count(SectionEnum.MARKET_PLACE),
+        "saved_items": await MarketPlaceSearchServiceAsync().get_current_user_bookmarks_count(SectionEnum.MARKET_PLACE),
```
MarkLark86: The argument to `get_current_user_bookmarks_count` is probably not needed here anymore

newsroom/media_releases/views.py
```diff
@@ -32,7 +32,9 @@ async def get_view_data():
         "company": str(company.id) if company else None,
         "navigations": [],
         "formats": get_formatters_id_and_names(SectionEnum.WIRE),
-        "saved_items": await WireSearchServiceAsync().get_current_user_bookmarks_count(SectionEnum.MEDIA_RELEASES),
+        "saved_items": await MediaReleasesSearchServiceAsync().get_current_user_bookmarks_count(
```
MarkLark86: And same here

newsroom/media_releases/views.py
```diff
@@ -74,7 +76,8 @@ async def bookmark():
     assert data.get("items")
     await update_action_list(data.get("items"), "bookmarks", item_type="items")
     push_user_notification(
-        "saved_items", count=await WireSearchServiceAsync().get_current_user_bookmarks_count(SectionEnum.MEDIA_RELEASES)
+        "saved_items",
+        count=await MediaReleasesSearchServiceAsync().get_current_user_bookmarks_count(SectionEnum.MEDIA_RELEASES),
```
MarkLark86: And here

newsroom/wire/views.py
```diff
@@ -77,9 +77,13 @@
 HOME_EXTERNAL_ITEMS_CACHE_KEY = "home_external_items"
 
 
-async def set_permissions(wire_item: WireItem, ignore_latest=False):
+async def set_permissions(
+    wire_item: WireItem, ignore_latest: bool = False, service: Optional[WireSearchServiceAsync] = None
```
MarkLark86: With Python 3.10+ there is no need to import and use `typing.Optional`, you can instead use the in-built type `| None` which does the same without the need to import a type

MarkLark86 (commented):
Looks good, just a couple small improvement suggestions

newsroom/market_place/views.py
marwoodandrew: Done!

newsroom/media_releases/views.py
marwoodandrew: Done!

newsroom/media_releases/views.py
marwoodandrew: Done!

newsroom/wire/views.py
marwoodandrew: Done!

newsroom/factcheck/views.py
MarkLark86, line 1: https://github.com/superdesk/newsroom-core/blob/01f283d6afbe07e546ec54308eaa10e1943c43d4/newsroom/factcheck/views.py#L60 Requires `await` prefix as `get_view_data` is async

marwoodandrew: Yep, I know, I'd included it in this PR then removed it as I was going to raise another PR to fix bookmarks! However It's only a two file change so I'll add it to this one again! If that's OK!

MarkLark86 (approved):

PR #1287, opened 2025-06-04
fix sentry version conflict with superdesk

petrjasek:
### Checklist

Resolves: #[issue-number]

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -460,10 +460,8 @@ rsa==4.9
     #   oauth2client
 s3transfer==0.11.3
     # via boto3
-sentry-sdk[quart]==1.45.1
-    # via
-    #   -r requirements.txt
-    #   superdesk-core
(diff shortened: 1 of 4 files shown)


MarkLark86:
@petrjasek Would this be better off going into the `release/3.0` branch?

petrjasek:
seems like that only works with `superdesk-core#async` and not with the tagged version used in `release/3.0` branch

MarkLark86:
> seems like that only works with `superdesk-core#async` and not with the tagged version used in `release/3.0` branch

Fair enough, let's aim for a 3.1 release then. Would be good to change the base branch back to async, so release/3.0 will stay with the same Superdesk version it uses now

MarkLark86 (approved):

petrjasek:
depends on https://github.com/superdesk/superdesk-core/pull/2999

PR #1291, opened 2025-06-09
[STT-47] - Adds location.details searchability and fixes name required error

BrianMwangi21:
### Purpose
This PR adds `location.details` to AGENDA_SEARCH_FIELDS and also makes location.name optional so that an event can be posted with location details and no location name.

### What has changed
- Added `location.details` to AGENDA_SEARCH_FIELDS to make it searchable.
- To fix `location.0.name : Field is required` error, location.name field has been made optional. Hence, an event can be posted with location details but without location name.

Resolves: #[STT-47]

[STT-47]: https://sofab.atlassian.net/browse/STT-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/types/agenda.py
+++ b/newsroom/types/agenda.py
@@ -120,7 +120,7 @@ class AgendaCoverage(Dataclass):
 
 
 class EventLocation(Dataclass):
-    name: fields.TextWithKeyword
+    name: fields.TextWithKeyword | None = None
     address: Annotated[dict | None, fields.dynamic_mapping()] = None
     location: fields.Geopoint | None = None
     qcode: fields.Keyword | None = None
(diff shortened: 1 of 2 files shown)


newsroom/types/agenda.py
```diff
@@ -120,7 +120,7 @@ class AgendaCoverage(Dataclass):
 
 
 class EventLocation(Dataclass):
-    name: fields.TextWithKeyword
+    name: fields.TextWithKeyword | None = None
```
petrjasek, line 123: I'm not sure if we should actually allow publishing of locations without name from SD, but in theory this is more flexible

petrjasek (approved):

PR #1293, opened 2025-06-10
[CPCN-1091] fix: Use 500 page size for Cards, Products, Topics, Companies & Users

MarkLark86:
### Purpose
If there are 25 or more existing Topics for a single user, when creating more the front-end does not show these new Topics (as the default page size is 25).

### What has changed
~~Use [.search](https://github.com/superdesk/superdesk-core/blob/v3.0.0-alpha.3/superdesk/core/resources/service.py#L291) instead as the [.find](https://github.com/superdesk/superdesk-core/blob/v3.0.0-alpha.3/superdesk/core/resources/service.py#L696) method applies pagination. Also apply this change to other parts of the code that should use `search` instead~~

Use page size of 500 when getting resources like Cards, Products, Topics, Companies and Users (with filters applied).

**Note:** I reverted back to using `.find` as this applies the resource's configured default sort, where as `.search` does not. Applying a limit of 500 should be fine.

### Steps to test
Create 25 Topics, then create another one and check to see if they newer ones are available in the front-end.

image

Resolves: [CPCN-1091](https://sofab.atlassian.net/browse/CPCN-1091)

[CPCN-1091]: https://sofab.atlassian.net/browse/CPCN-1091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/agenda/filters.py
+++ b/newsroom/agenda/filters.py
@@ -257,18 +257,19 @@ def apply_product_planning_filters(request: NewshubSearchRequest[AgendaSearchReq
 
 
 def apply_agenda_query_string(request: NewshubSearchRequest[AgendaSearchRequestArgs]) -> None:
-    if not request.args.q:
+    search_text = request.args.q.strip() if request.args.q else None
+    if not search_text:
         return
 
     q_dict: dict[str, Any] | None = None
     try:
-        q_dict = json.loads(request.args.q)
+        q_dict = json.loads(search_text)
     except ValueError:
         pass
 
-    if q_dict is None:
+    if not isinstance(q_dict, dict):
         # Normal query string query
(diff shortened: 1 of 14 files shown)


eos87 (approved):
Looks good. Quick thought: since we use the same for `CardsResourceService` in multiple places... would it worth it to put it in a separate service method and reuse it?

MarkLark86:
@eos87 I have adjusted the fix as per your suggestion. I changed the base Newshub async service to default to 500 page size (which works across system resources, excluding content/planning resources). This should be a cleaner solution.

This also raised a typing/bug in the ManagementAPI for users email search, so I fixed that one up too in the PR

eos87 (approved):
Neat solution!

PR #1297, opened 2025-06-10
[CPCN-1028] Icon font update

fritzSF:
### Checklist

Resolves: #[issue-number]


dzonidoo:
@fritzSF fix the conflicts

package.json
```diff
@@ -57,7 +57,6 @@
     "redux-logger": "3.0.6",
     "redux-thunk": "2.4.2",
     "reselect": "3.0.1",
-    "sass": "~1.3.0",
```
petrjasek: we need it afaik, can you revert those changes to package.json and package-lock.json?

petrjasek (approved):

PR #1299, opened 2025-06-11
[CPCN-1089] fix: Agenda pagination using invalid logic

MarkLark86:
### Purpose
Agenda pagination was using `from` in the front-end for pagination. This was mistakenly interpreted as `page` on the back-end.

### What has changed
* As we don't have support for `from` param in new async search services, I opted to convert it to a `page` param instead, rather than update an older alpha version of Superdesk to add this support.
* Fixed an issue where the `_display_from` and `_display_to` attributes were set on agenda items without a date search being in place, which restricted the days the items would be shown up to (`_display_from` + 10 days).

Resolves: [CPCN-1094](https://sofab.atlassian.net/browse/CPCN-1094)

[CPCN-1094]: https://sofab.atlassian.net/browse/CPCN-1094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/agenda/agenda_search.py
+++ b/newsroom/agenda/agenda_search.py
@@ -106,7 +106,7 @@ class AgendaSearchServiceAsync(BaseWebSearchService[AgendaSearchRequestArgs, Age
             matching_event_ids: set[str] = (
                 set() if args.item_type is not None else await self._get_event_ids_matching_query(args)
             )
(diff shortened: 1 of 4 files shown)


eos87 (approved):
Looks good

PR #1303, opened 2025-06-17
[CPCN-1090] Set the right elastic search prefix

eos87:
### About this PR
The wire resource config relies on the Elasticsearch prefix CONTENTAPI_ELASTICSEARCH (see below). This PR updates the setting to use the correct prefix so that ConfigModel.load_from_dict can properly pick up its value.

https://github.com/superdesk/newsroom-core/blob/2fb44281900488ec7ffb6f010e3981bf22073054/newsroom/__init__.py#L19

### What has changed
- Fixed the setting prefix used in the wire resource configuration to ensure it is correctly loaded.

This PR requires https://github.com/superdesk/superdesk-core/pull/2971 to fully resolve the issue in [CPCN-1090].

Resolves: [CPCN-1090]

[CPCN-1090]: https://sofab.atlassian.net/browse/CPCN-1090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CPCN-1090]: https://sofab.atlassian.net/browse/CPCN-1090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarkLark86:
@eos87 Does this mean we should also change https://github.com/superdesk/newsroom-core/blob/4852814d6b6ad3a792ed4567b48646de18111d85/newsroom/web/default_settings.py#L487 to `CONTENT_API_ELASTICSEARCH_QUERY_STRING_DEFAULT_PARAMS`?

MarkLark86:
@eos87 Also there is a unit test failing ~~, any idea why?~~
```
FAILED tests/core/test_agenda.py::test_filter_events_only - assert 2 == 1
```

Oh, I'm guessing because it needs the `superdesk-core` fix so it doesn't interpret the `True` as a `1`?

eos87:
> @eos87 Also there is a unit test failing ~, any idea why?~
> 
> ```
> FAILED tests/core/test_agenda.py::test_filter_events_only - assert 2 == 1
> ```
> 
> Oh, I'm guessing because it needs the `superdesk-core` fix so it doesn't interpret the `True` as a `1`?

Indeed. I updated the branch in the requirements and it's all green now.

eos87:
> @eos87 Does this mean we should also change
> 
> https://github.com/superdesk/newsroom-core/blob/4852814d6b6ad3a792ed4567b48646de18111d85/newsroom/web/default_settings.py#L487
> 
> to `CONTENT_API_ELASTICSEARCH_QUERY_STRING_DEFAULT_PARAMS`?

@MarkLark86 yeah I thought so but, it is referenced exactly as it's defined in the settings so I decided not to change it. See line 15 below
https://github.com/superdesk/newsroom-core/blob/2fb44281900488ec7ffb6f010e3981bf22073054/newsroom/search/utils.py#L14-L16

MarkLark86:
Ah right. All good then

> > @eos87 Does this mean we should also change
> > 
> > https://github.com/superdesk/newsroom-core/blob/4852814d6b6ad3a792ed4567b48646de18111d85/newsroom/web/default_settings.py#L487
> > 
> > to `CONTENT_API_ELASTICSEARCH_QUERY_STRING_DEFAULT_PARAMS`?
> 
> @MarkLark86 yeah I thought so but, it is referenced exactly as it's defined in the settings so I decided not to change it. See line 15 below
> https://github.com/superdesk/newsroom-core/blob/2fb44281900488ec7ffb6f010e3981bf22073054/newsroom/search/utils.py#L14-L16

MarkLark86 (approved):

PR #1331, opened 2025-08-08
SDAN-733 The Subscriber Activity report is limited to 1000 rows this …

marwoodandrew:
…is to small

### Purpose

Addresses [https://sofab.atlassian.net/browse/SDAN-733](https://sofab.atlassian.net/browse/SDAN-733)
- Allows the Subscriber Activity report contain up-to 5000 rows

### What has changed
- [Bug fix](https://github.com/superdesk/newsroom-core/blob/7079882d5b65473bcb50ee153f01aa11832d80d2/newsroom/reports/reports.py#L225) in the application of the date range to the base query for the Subscriber Activity Report (both start and end date were set to date_from)
- Implements the ability to export the Subscriber Activity Report to CSV
- Port some additional queries to Async
- Additional rows in the report will allow the identification of which update of a story the action was made against.

### Steps to test

- Make a number of actions in Newsroom such as opening, previewing content over a number of days.
- Ensure the report correctly will filter the report based on the selected dates
- Export the report to CSV an ensure the data is consistent with the report displayed on the web page.

Note additional columns and the availability of an "Export to CSV" button
image

### Checklist

- [ x] This pull request is not adding new forms that use redux
- [ x] This pull request is adding missing TypeScript types to modified code segments where it's easy to do so with confidence
- [ x] This pull request is replacing `lodash.get` with optional chaining for modified code segments

Resolves: #[SDAN-733]

[SDAN-733]: https://sofab.atlassian.net/browse/SDAN-733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/reports/reports.py
+++ b/newsroom/reports/reports.py
@@ -1,27 +1,25 @@
 from typing import cast
-import io
-import csv
 from collections import defaultdict
 from copy import deepcopy
 
 from bson import ObjectId
 from quart_babel import gettext
-from werkzeug.utils import secure_filename
+from superdesk import get_app_config
 
 from superdesk.core import get_current_app
-from superdesk.core.resources.cursor import ElasticsearchResourceCursorAsync
-from superdesk.flask import abort, request, send_file
-import superdesk
-from superdesk.utc import utcnow
+from superdesk.core.resources.cursor import ElasticsearchResourceCursorAsync, ResourceCursorAsync
+from superdesk.flask import abort, request
+from superdesk.utc import utcnow, utc_to_local, get_date
 
-from newsroom.types import CompanyResource, NewsApiAuditResourceModel
+from newsroom.types import CompanyResource, NewsApiAuditResourceModel, AgendaItem, WireItem
+from newsroom.history_async import HistoryService
 from newsroom.auth.utils import get_company_from_request
 from newsroom.utils import (
     query_resource,
     get_entity_dict,
-    get_items_by_id,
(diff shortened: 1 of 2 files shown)


newsroom/reports/reports.py
```diff
@@ -260,37 +264,98 @@ async def get_subscriber_activity_report():
...
+    wire_ids = list(set(wire_ids))
+    agenda_ids = list(set(agenda_ids))
+
+    AGENDAITEM_CHUNK_SIZE: int = 100
+    agenda_items: dict[str, AgendaItem] = {}
+    for i in range(0, len(wire_ids), AGENDAITEM_CHUNK_SIZE):
+        agenda_cursor: ResourceCursorAsync[AgendaItem] = await AgendaItemService().search(
+            {"_id": {"$in": agenda_ids[i : i + AGENDAITEM_CHUNK_SIZE]}}, use_mongo=True
+        )
+        agenda_items: dict[str, AgendaItem] = {
```
MarkLark86: This is overwriting the `agenda_items` variable, should this rather be updating it instead?

MarkLark86 (approved):

newsroom/reports/reports.py
marwoodandrew: Yep, that's bad will fix!

PR #1344, opened 2025-08-20
Implementing the embed permissioning functionality

marwoodandrew:
### Purpose
Implementing the embed permissioning functionality

### What has changed
Based on Superdesk/Newsub products some media embeds can be hidden from some users.

### Checklist

Resolves: #[issue-number]


newsroom/formatters.py
```diff
@@ -29,6 +29,7 @@ class BaseFormatter:
     FILE_EXTENSION = None
     MEDIATYPE = "text"
     MULTI = False
+    MULTI_ZIP = False
```
marwoodandrew, line 31: I couldn't see a use for "MULTI" and was tempted to use it instead, but was afraid I'd missed something so added a new one!

MarkLark86: You're right, remnants of old code. I have removed it in favour of `MULTI_ZIP` (as it's more descriptive anyway)

marwoodandrew (commented):
LGTM can't seem to approve I guess 'cause I own it!

MarkLark86 (approved):

PR #1348, opened 2025-08-28
[SDAN-737] - Replace Playr with another library

BrianMwangi21:
### Purpose

This PR implements new video player enhancements to respect the `data-disable-download` attribute thus disabling
download functionality for product-filtered media using video.js

### What has changed
- Added Video.js
- Added logic to check for data-disable-download="true": if true, disable downloads via controlsList='nodownload', prevent right-click context menu, and remove native controls; else, enable downloads.
- Added cross-browser CSS for download prevention (webkit for Safari, context menu for Firefox).

### Steps to test
1. Set "EMBED_PRODUCT_FILTERING": true in CLIENT_CONFIG.
2. Follow the steps from this [PR](https://github.com/superdesk/newsroom-core/pull/1338)
3. View article in Newshub without the right permissions: confirm download disabled and no context menu
4. Assign product to company, refresh article: verify data-disable-download removed, downloads enabled.

Resolves: [SDAN-737](https://sofab.atlassian.net/browse/SDAN-737)

[SDAN-737]: https://sofab.atlassian.net/browse/SDAN-737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/styles/app.scss
```diff
@@ -28,3 +28,4 @@
 @import './time-picker';
 @import './content-divider';
 @import '../components/pagination/sd-pagination';
+@import 'video.js/dist/video-js.css';
```
MarkLark86: We have a [vendor.scss](https://github.com/superdesk/newsroom-core/blob/develop/assets/styles/vendor.scss) file where this would be better placed

assets/ui/utils.ts
```diff
@@ -24,4 +25,32 @@ export function bem(block: any, element: any, modifier: any) {
     }
 
     return classes.join(' ');
-}
\ No newline at end of file
+}
+
+export function setupVideoPlayers(root: HTMLElement) {
+    const players: any[] = [];
+    root.querySelectorAll('video').forEach((element) => {
```
MarkLark86: This should also be used for audio elements, would that be easy enough to support?

assets/ui/utils.ts
```diff
@@ -24,4 +25,32 @@ export function bem(block: any, element: any, modifier: any) {
...
+            element.addEventListener('contextmenu', (e) => e.preventDefault());
+            element.removeAttribute('controls');
+        } else {
+            element.setAttribute('controls', '');
+        }
+
+        element.setAttribute('data-vjs-initialized', 'true');
+        element.classList.add('video-js', 'vjs-big-play-centered');
+
+        const player = videojs(element, {
```
MarkLark86: We should only be using this videojs player if it is configured in the [CLIENT_CONFIG](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/web/default_settings.py#L341).

Maybe a boolean config on it like `use_videojs = False` (and default it to False, otherwise some customers may not like this change)

assets/ui/video-utils.ts
```diff
@@ -0,0 +1,29 @@
+import videojs from 'video.js';
+
+export function setupVideoPlayers(root: HTMLElement) {
```
MarkLark86: This looks like a duplicate to whats ui ui/utils. Might be good to choose 1 place for it

assets/ui/utils.ts
MarkLark86: OR
Maybe we apply this ONLY to video/audio that has the disable download attribute

assets/ui/utils.ts
BrianMwangi21: Yeah, it supports audio too.

assets/ui/utils.ts
BrianMwangi21: Got it. I had already added the `use_videojs` in the config before seeing your other comment. Let me revert that and just apply video js to those items with disable-download attribute

assets/ui/utils.ts
```diff
@@ -24,4 +25,45 @@ export function bem(block: any, element: any, modifier: any) {
     }
 
     return classes.join(' ');
-}
\ No newline at end of file
+}
+
+export function setupMediaPlayers(root: HTMLElement) {
+    const players: any[] = [];
```
MarkLark86: 1. Avoid `any` where possible, does videojs expose a ts type we can use instead?
2. For consistency, we tend to use the `Array<>` convention in our codebase

assets/ui/utils.ts
```diff
@@ -24,4 +25,45 @@ export function bem(block: any, element: any, modifier: any) {
...
+
+            const player = videojs(element, {
+                controls: true,
+                preload: 'auto',
+                fluid: true,
+            });
+            players.push(player);
+        } else {
+            element.setAttribute('controls', '');
+            players.push(null);
```
MarkLark86: Is this needed? If it isn't needed during cleanup cycle, then we can keep the `players` array only include those that need processing on cleaup

assets/styles/index.scss
```diff
@@ -4218,3 +4218,7 @@ article.list {
 .tooltip {
     pointer-events: none;
 }
+
+video::-webkit-media-controls-panel {
```
MarkLark86: Should this target all video elements, or just those with the `data-disable-download==true` attribute?
Does this work on non webkit based browsers (such as Firefox)

assets/ui/utils.ts
```diff
@@ -24,4 +25,45 @@ export function bem(block: any, element: any, modifier: any) {
...
+    const players: any[] = [];
+
+    root.querySelectorAll('video, audio').forEach((element) => {
+        if (element.getAttribute('data-vjs-initialized')) return;
+        const disable = element.getAttribute('data-disable-download') === 'true';
+
+        element.setAttribute('data-vjs-initialized', 'true');
+
+        if (disable) {
+            element.setAttribute('controlsList', 'nodownload');
```
MarkLark86, line 44: [controlsList](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList) is not supported on Firefox or Safari.

We need to make sure we can disable the download option for all browsers, desktop and mobile alike.

BrianMwangi21: Sure. That's why we have the other call above it to remove the controls which is supported by all browsers. 

This line is for an additional override. 

I've also added comments to explain.

MarkLark86 (approved):

PR #1349, opened 2025-08-29
[SDAN-738] Feature: Wire Embed Permissions

MarkLark86:
### Purpose
New feature to support Company level permissions for Wire item associations and body html embeds.

### What has changed
* New `embed_permissions`  dictionary on the Company resource, which stores ContentType against a list of allowed user actions (`display`, `download`)
* Update Company editor to allow managing these permissions from the web app
* Moved all previous embed functionality into a common `newsroom.wire.embeds` module
* On API response:
  * Remove all associations that the current Company does not have `display` permission for
  * Remove all embeds in body_html that the current Company does not have `display` permission for
  * Mark all embeds in body_html to indicate to front-end elements that should disallow downloading, if the current Company does not have the `download` permission for

### Screenshots
image

### Checklist

**Note:** Redux is used for the embed permission form, as the rest of the Company forms use it (that way we aren't mixing the react component state & redux state for the one form)

Resolves: [SDAN-738](https://sofab.atlassian.net/browse/SDAN-738)

[SDAN-738]: https://sofab.atlassian.net/browse/SDAN-738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


newsroom/companies/module.py
```diff
@@ -22,11 +24,20 @@ class CompanyConfigs(ConfigModel):
     company_types: list[CompanyType] = []
 
 
+def init_module(app: SuperdeskAsyncApp):
+    # Only set ``CLIENT_CONFIG.embed_permissions_enabled`` is this is the WebAPI
```
petrjasek: probably typo in the comment `s/is/if`

newsroom/web/default_settings.py
```diff
@@ -875,3 +872,15 @@
     "country": True,
     "place": True,
 }
+
+#: Enable/Disable Wire embedded permissions
+#:
+#: .. versionadded: 3.1
+#:
+WIRE_EMBED_PERMISSIONS = True
```
petrjasek: I wonder if this should be on by default

petrjasek (approved):

marwoodandrew (approved):

newsroom/web/default_settings.py
MarkLark86: I have disabled it by default now, because it comes with a performance hit (iterating all associations/embeds for all items)

PR #1356, opened 2025-09-10
[STT-1301] Refactor display field config types and field rendering

eos87:
### Purpose
This refactor makes the field rendering logic safer and easier to maintain. The old implementation relied heavily on `any` types and ad-hoc checks, which made it harder to understand and prone to subtle bugs. By introducing a proper `IDisplayFieldsConfig` union and linking field names directly to `IArticle`, the code is now easier to follow and less error-prone. 
**No UI changes are expected.**

> [!NOTE]  
> Added some documentation here https://sofab.atlassian.net/wiki/x/AwAd0QI

### What has changed
- Reworked `IDisplayFieldsConfig` into a single, clear union supporting simple fields (`keyof IArticle`), separators (`'/' | '//' | '-'`), arrays, styled wrappers, and component overrides.
- Added types: `Separator`, `ComponentName`, `StyledField`, `ComponentField`.
- Added `charcount?: number` to `IArticle`.
- Typed default meta arrays in `WireListItem` (`DEFAULT_META_FIELDS`, `DEFAULT_COMPACT_META_FIELDS`) as `IDisplayFieldsConfig`.
- Converted `VersionType` to TypeScript props (`{ value: string }`) and removed `prop-types`.
- Refactored `fields/index.tsx` for readability and type safety:
  - small, named type guards: `isSeparator`, `isStringField`, `isStyledField`, `isComponentField`
  - explicit `FieldResult` / `FieldComponentsProps`
  - cleaner separator handling 

### Checklist

Belongs to [STT-1301]

[STT-1301]: https://sofab.atlassian.net/browse/STT-1301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eos87:
A second PR will be added to allow a new component type like what we have for version type in order to resolve STT-1301

petrjasek (approved):

assets/interfaces/configs.ts
```diff
@@ -1,23 +1,31 @@
...
-    >
+import {IArticle} from './content';
+
+export type Separator = '/' | '//' | '-';
+
+// allowed component names
+export type ComponentName = 'version_type' | string;
+
+// component override: { field: "version", component: "version_type" }
+export interface ComponentField {
```
thecalcc, line 16: you can also do a generic argument here with default `never` that allows you to have a proper type for styles if you want to have it more specific. Also why not just put it as `CSSProperties`?

assets/wire/components/fields/index.tsx
```diff
@@ -9,11 +9,37 @@ import {Embargo} from './Embargo';
...
+}
+
+const ALLOWED_SEPARATORS: Separator[] = ['/', '//', '-'];
 const SEPARATOR_KEY = 'separator';
 
-const MAP_FIELD_TO_COMPONENT: any = {
+const isSeparator = (f: FieldConfig): f is Separator =>
+    typeof f === 'string' && (ALLOWED_SEPARATORS).includes(f as Separator);
+
+const isStringField = (f: FieldConfig): f is keyof IArticle => typeof f === 'string' && !isSeparator(f);
```
thecalcc, line 41: maybe we can have a comment here to explain why you check `!isSeperator(f)` since separator is also a string. or another function name like `isGenericString`

assets/wire/components/fields/index.tsx
```diff
@@ -61,46 +89,10 @@ export function FieldComponents({config, item, fieldProps = {}}: any) {
...
-                            <result.Component {...props} />
-                        </span>
-                    ),
-                }
-                : null;
-        }
-
-        return null;
-    } else if (Array.isArray(field) && field.length > 0) {
+function getComponentForField(item: IArticle, fieldConfig: FieldConfig): any {
```
thecalcc: could use a comment that it's recursive

assets/interfaces/configs.ts
```diff
@@ -1,23 +1,31 @@
...
+    field: keyof IArticle;
+    component: ComponentName;
+    styles?: never;
+}
+
+// styled wrapper: { field: <any valid config>, styles: {...} }
+export interface StyledField {
+    field: IDisplayFieldsConfig;
+    styles: { [key: string]: string | number };
+    component?: never;
```
thecalcc: why is `component` `never`?

assets/wire/components/fields/index.tsx
```diff
@@ -9,11 +9,37 @@ import {Embargo} from './Embargo';
...
+import {IArticle} from 'interfaces';
 
-const ALLOWED_SEPARATORS = ['/', '//', '-'];
+interface FieldResult {
+    key: string;
+    Component: React.ComponentType<any>;
+}
+
+interface FieldComponentsProps {
+    config: FieldConfig | FieldConfig[];
```
thecalcc: i think we don't use `type[]` rather `Array<type>`

assets/wire/components/fields/index.tsx
```diff
@@ -9,11 +9,37 @@ import {Embargo} from './Embargo';
 import {VersionCreated} from './VersionCreated';
 import {VersionType} from './VersionType';
 import {ExpiryDateLabel} from './ExpiryDateLabel';
+import {IDisplayFieldsConfig as FieldConfig, Separator, StyledField, ComponentField} from 'interfaces/configs';
+import {IArticle} from 'interfaces';
 
-const ALLOWED_SEPARATORS = ['/', '//', '-'];
+interface FieldResult {
+    key: string;
+    Component: React.ComponentType<any>;
```
thecalcc: could probably be `React.ReactNode` to avoid `any`

assets/wire/components/fields/index.tsx
```diff
@@ -9,11 +9,37 @@ import {Embargo} from './Embargo';
...
 const SEPARATOR_KEY = 'separator';
 
-const MAP_FIELD_TO_COMPONENT: any = {
+const isSeparator = (f: FieldConfig): f is Separator =>
+    typeof f === 'string' && (ALLOWED_SEPARATORS).includes(f as Separator);
+
+const isStringField = (f: FieldConfig): f is keyof IArticle => typeof f === 'string' && !isSeparator(f);
+
+const isObjectCfg = (f: FieldConfig): f is StyledField | ComponentField =>
+    typeof f === 'object' && f !== null && !Array.isArray(f);
```
thecalcc, line 44: could you make this one cleaner like `typeof f === 'object' && !Array.isArray(f ?? [])` and since you're doing a strict check for not null probably f should be `FieldConfig | null`

assets/interfaces/configs.ts
```diff
@@ -1,23 +1,31 @@
...
+}
+
+export type IDisplayFieldsConfig =
```
thecalcc, line 29: i guess it's not possible to enhance this type to add a unifying property like `type` or `kind` which would make type narrowing easier later, since this is coming from the backend?

thecalcc (commented):
looks good, but I guess my main question is readability - I never liked those functions that do type narrowing based on some properties, if we could add this `type`/`kind` property to `IDisplayFieldsConfig` it would improve the code a lot I think, but if it can't be added easily i guess readability of these functions would be the next best option

assets/wire/components/fields/index.tsx
eos87: no because `React.ReactNode` is a rendered value and what we need is the component itself before being rendered

assets/interfaces/configs.ts
eos87: leftover. fixed

assets/wire/components/fields/index.tsx
eos87: not needed - I think. The idea is that when it is a separator we treat it differently as we render it in a span tags and stop it there.

assets/interfaces/configs.ts
eos87: sorry I'm not following here. Can you please elaborate a little bit more how those properties would be used?

eos87:
> looks good, but I guess my main question is readability - I never liked those functions that do type narrowing based on some properties, if we could add this `type`/`kind` property to `IDisplayFieldsConfig` it would improve the code a lot I think, but if it can't be added easily i guess readability of these functions would be the next best option

Hi @thecalcc. I get your concern, but in this case the small, named type guards improve readability versus the previous chain of if/else checks. A few reasons: 
- The intent is explicit (like reading plain English) 
- Single responsibility
- They would make the code easier to test (in the near future or next PR ;))

Thoughts?

tomaskikutis:
[Discriminated unions](https://www.typescriptlang.org/play/?#code/PTAEBEEsGcGMCdIFtIDsCGAXApgE1AK6qQD2qoAZifKNABboAO2omAns9ALABQ7zoAMKR4sADYsAvKADeoANZpcALlAAiWCPHY1AblDx0uSAWirUBJACNsNAL67e-FgGUAjgXTwpshUtVq0B5eOvrQkLjY5pY29o58HK4MAtLCohKgAD6g7p7e8bwgoAAqiQC00OgULBREsJik5KZoAOasdCzGcIgoGKiYvLWo9Y2gLdiYAILe6AAU9ExROcnYAJTR1rayvKC0AO6QmLB0oPMrAHSKqLir2zy7u7Do0CwaWhJqyjsPj2TQJBJzmISC0zotzuFIrcigAFeAkZjwdigADkkOwKNAuBI2GgoFQJEwoGwAA8YESyKxEqi0toUedZgAmADMzIAnKtvj9vJgCPByABZLB0c4wgCSoAAVLQLoZjKYpdLGfEfk8XuognkdF97j9QLA-gDsECQWDmOc5SZoNCwHCEbZkSjLaZMdjcfjCcSydAKeRnKjciF6UzWRyuQ8eXzyAtzejFaBld87LxkzxCmAAKqVca8A2of6A4Gg8ZTGazORXFTqTTpHQAGgMRitqgArKA7KtVvpQEUAH4AdgAHOcWwAWXOGwumkvTbBzCv+DXBbxqBvo1Sj9ud7u7IoARgAbEA) are always a better choice when possible. When not possible or would be especially time consuming to use them, then casting to a type based on some properties is best that remains.

thecalcc (approved):

PR #1370, opened 2025-09-30
[SDAN-737-Fix] - Videojs fix + implementation on featured video/audio items

BrianMwangi21:
### Purpose
This PR adds a try/catch block for `setupMediaPlayers` function. It also sets the `data-vjs-initialized` as true only when successful to allow for retries. This PR also adds videojs to featured media audio/video to enable/disable downloads. It also hides the download button if it is not allowed.

### Checklist

Resolves: #[SDAN-737](https://sofab.atlassian.net/browse/SDAN-737)

[SDAN-737]: https://sofab.atlassian.net/browse/SDAN-737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/ui/components/ArticleMedia.tsx
+++ b/assets/ui/components/ArticleMedia.tsx
@@ -11,7 +11,8 @@ export default function ArticleMedia({isKilled, media, download}: any) {
     const disableDownload = media.disable_download;
 
     useEffect(() => {
-        const cleanupMediaPlayers = containerRef.current && setupMediaPlayers(containerRef.current);
+        if (!containerRef.current) return;
+        const cleanupMediaPlayers = setupMediaPlayers(containerRef.current);
         return () => cleanupMediaPlayers?.();
     }, [media]);
 
--- a/assets/ui/components/ArticleMedia.tsx
+++ b/assets/ui/components/ArticleMedia.tsx
@@ -1,29 +1,37 @@
-import React from 'react';
+import React, {useRef, useEffect} from 'react';
 import PropTypes from 'prop-types';
 import {gettext} from 'utils';
 import {getOriginalRendition} from 'wire/utils';
+import {setupMediaPlayers} from '../utils';
 
 export default function ArticleMedia({isKilled, media, download}: any) {
     const rendition = getOriginalRendition(media);
     const filename = media.slugline || rendition.media;
+    const containerRef = useRef<HTMLDivElement>(null);
+    const disableDownload = media.disable_download;
+
+    useEffect(() => {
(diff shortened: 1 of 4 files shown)


newsroom/wire/embeds.py
```diff
@@ -286,6 +286,10 @@ def _remove_or_disable_item_media(item: dict, company: CompanyResource, permitte
     disable_display, disable_download = _get_associations_to_remove_or_disable(item, company, permitted_products)
     disable_embed_codes = not company.is_permissioned_for_embed("embed_code", EmbedPermissionUserAction.DISPLAY)
 
+    # Tag featuredMedia association if download is disabled too
+    if "featuremedia" in disable_download and item.get("associations", {}).get("featuremedia"):
```
MarkLark86: This might be good to place it where the disable download is checked: https://github.com/superdesk/newsroom-core/blob/a0ce613d06851424cc0263cd82397b63f8e56b9f/newsroom/wire/embeds.py#L277-L280

This way we can add a `disable_download` attribute on _any_ association, for use by the front-end.

assets/ui/components/ArticleMedia.tsx
```diff
@@ -1,29 +1,37 @@
...
 
 export default function ArticleMedia({isKilled, media, download}: any) {
     const rendition = getOriginalRendition(media);
     const filename = media.slugline || rendition.media;
+    const containerRef = useRef<HTMLDivElement>(null);
+    const disableDownload = media.disable_download;
+
+    useEffect(() => {
+        const cleanupMediaPlayers = containerRef.current && setupMediaPlayers(containerRef.current);
+        return () => cleanupMediaPlayers?.();
```
MarkLark86, line 16: Would `containerRef.current &&` be best off placed inside the cleanup function, so the dom node is checked when the cleanup function is run?
i.e.

```ts
return () => {
    if (containerRef.current != null) {
        setupMediaPlayers(containerRef.current);
    }
}
```

assets/ui/utils.ts
```diff
@@ -28,14 +28,44 @@ export function bem(block: any, element: any, modifier: any) {
     return classes.join(' ');
 }
 
+function tryInitPlayer(el: HTMLElement, retries = 3, delay = 100): VjsPlayer | null {
```
MarkLark86: Under what conditions does initialising the player fail? It would be good to understand that, and try and fix the cause.

Also, when retrying it seems like this function doesn't return the player, as setTimeout is used and a null is returned.

MarkLark86 (approved):

PR #1372, opened 2025-10-02
set default MAX_CONTENT_LENGTH to 1GB

petrjasek:
it was 16MB by default

CPCN-1149

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/newsroom/web/default_settings.py
+++ b/newsroom/web/default_settings.py
@@ -874,3 +874,9 @@ CALENDAR_LOCATIONS_FILTER_OPTIONS = {
     "country": True,
     "place": True,
 }
+
+#: Maximum allowed content length for file uploads (in bytes)
+#:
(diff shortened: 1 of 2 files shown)


marwoodandrew (approved):

MarkLark86:
@petrjasek We have to be careful with regards to DDoS with changes like this.
Also if push_key is used, then it looks like the entire file is placed in memory to check key vs body, which CP are already having issues with OOM conditions.

https://github.com/superdesk/newsroom-core/blob/d81a6b8cce22e165c37e5e24706d8926f4a2fe1d/newsroom/push/utils.py#L132-L135

petrjasek:
yep issue is the old version had no limit, and there is also a limit set in nginx, so this is rather something just in case

MarkLark86 (approved):

PR #1385, opened 2025-10-30
fix: Loading of Event related Plannings failed to update store

MarkLark86:
### Purpose
There were 2 bugs with regards to loading an Event's related Planning items (mainly used when opening an Event in preview mode).
* The back-end was applying item_type filter when getting Agenda items by ID (which would exclude the desired Planning items)
* The front-end was failing to add the item to the Redux store (using incorrect Redux action/reducer)

### What has changed
* back-end: When getting Agenda by IDs, and no item_type filter applied, then don't apply default item_type filter
* front-end: Add new Redux action/reducer to store Agenda items (without updating search metadata, such as page, search result count etc)

Related to: [STT-1428](https://sofab.atlassian.net/browse/STT-1428) (discovered while testing Coverage schedule config)

[STT-1428]: https://sofab.atlassian.net/browse/STT-1428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/agenda/reducers.ts
+++ b/assets/agenda/reducers.ts
@@ -105,7 +105,7 @@ function receiveItemsFromSearch(state: IAgendaState, data: IRestApiResponse<IAge
 }
 
 function receiveItems(state: IAgendaState, items: Array<IAgendaItem>): IAgendaState {
-    const itemsById = Object.assign({}, state.itemsById);
+    const itemsById = {...state.itemsById};
     const itemIds = items.map((item) => {
         itemsById[item._id] = item;
         return item._id;
--- a/assets/agenda/reducers.ts
+++ b/assets/agenda/reducers.ts
@@ -8,6 +8,7 @@ import {
 } from 'interfaces';
 import {
     RECIEVE_ITEMS,
+    STORE_ITEMS,
     SET_LIST_GROUPS_AND_ITEMS,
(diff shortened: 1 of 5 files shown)


assets/agenda/reducers.ts
```diff
@@ -109,6 +104,20 @@ function recieveItems(state: IAgendaState, data: IRestApiResponse<IAgendaItem>):
     };
 }
 
+function receiveItems(state: IAgendaState, items: Array<IAgendaItem>): IAgendaState {
+    const itemsById = Object.assign({}, state.itemsById);
```
eos87: nit:
```suggestion
    const itemsById = {...state.itemsById};
```

assets/agenda/reducers.ts
```diff
@@ -161,7 +170,10 @@ export default function agendaReducer(state: IAgendaState = initialState, action
     switch (action.type) {
 
     case RECIEVE_ITEMS:
```
eos87, line 172: do you mind fixing the typo here taking advantage of this task? :)   
```suggestion
    case RECEiVE_ITEMS:
```

eos87 (approved):
Looks good, just left a couple of minor comments.

PR #1397, opened 2025-11-11
[STT-1431] - Changes required to support hiding Coverage schedule 

BrianMwangi21:
### Purpose
This PR adds to the previous [PR](https://github.com/superdesk/newsroom-core/pull/1396) on the same ticket to make sure where coverage is null it is handled gracefully

### What has changed
- Added null checks for `coverage.scheduled` to be handled gracefully
- Refactored `getCoverageTooltip` from nested if-else to be more cleaner

Resolves: #[STT-1431](https://sofab.atlassian.net/browse/STT-1431)

[STT-1431]: https://sofab.atlassian.net/browse/STT-1431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/agenda/utils.ts
+++ b/assets/agenda/utils.ts
@@ -993,7 +993,7 @@ export function formatCoverageDate(coverage: ICoverage) {
         parseDate(coverage.scheduled).format(COVERAGE_DATE_TIME_FORMAT);
 }
 
-export const getCoverageTooltip = (coverage: any, beingUpdated?: any, restrictCoverageInfo?: boolean) => {
+const prepareTooltipData = (coverage: any) => {
     const slugline = coverage.item_slugline || coverage.slugline;
     const coverageType = getCoverageDisplayName(coverage.coverage_type);
     const coverageScheduled = coverage.scheduled ? moment(coverage.scheduled) : null;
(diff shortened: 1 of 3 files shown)


assets/agenda/utils.ts
```diff
@@ -984,75 +984,98 @@ export function isItemTBC(item: IAgendaItem): boolean {
  * @return {String}
  */
 export function formatCoverageDate(coverage: ICoverage) {
+    if (!coverage.scheduled) {
+        return TO_BE_CONFIRMED_TEXT;
```
MarkLark86: If the Coverage does't have a schedule defined, we shouldn't show anything.
For STT, they don't want to advertise coverage due date at all

assets/agenda/utils.ts
```diff
@@ -984,75 +984,98 @@ export function isItemTBC(item: IAgendaItem): boolean {
...
+        }
+        return coverageScheduled ? 
+            gettext('Planned {{ type }} coverage {{ slugline }}, expected {{date}} at {{time}} {{assignedDetails}}', {
                 type: coverageType,
                 slugline: slugline,
                 date: formatDate(coverageScheduled),
                 time: formatTime(coverageScheduled),
                 assignedDetails,
+            }) :
+            gettext('Planned {{ type }} coverage {{ slugline }} {{assignedDetails}}', {
```
MarkLark86: There seems to be lots of duplication of text in these cases, between `restrictCoverageInfo` and `coverageScheduled`.

Maybe we can use a single variable before the `switch` statement or'ing between these two, and use that in each case block

assets/agenda/utils.ts
```diff
@@ -1011,6 +1007,14 @@ const handleStatus = (coverage: any, data: any, beingUpdated?: any, restrictCove
     const {slugline, coverageType, coverageScheduled, assignedDetails} = data;
     const {workflow_status} = coverage;
 
+    const showScheduledInfo = !restrictCoverageInfo && coverageScheduled;
+    const scheduledText = showScheduledInfo
```
MarkLark86: Not sure if this will work across all language types. We would still need this in the 1 `gettext` call along with the prefixed text

MarkLark86 (approved):

PR #1401, opened 2025-11-17
[STT-1455] - Fix newshub `initialize_data` command to handle `ui_config.json` loading

BrianMwangi21:
### Purpose
This PR fixes the `initialize_data --entity-name ui_config` command which was failing silently since the service was not being found. This is because it is an async entity but the command was trying to use the legacy superdesk service registry to find the service in the `import_file` function

### What has changed
- Created a new `async_entity_import` function in `newsroom/commands/utils.py` that handles async entities 
- Modified `initialize_data.py` to use the new async import function specifically for ui_config entities
- Improved error handling and logging for missing entities and files in the command itself

### Steps to test
1. Run the initialize data command for ui_config: `python3 manage.py initialize_data --entity-name ui_config`
2. Verify that the ui_config collection is created in the database and the config is loaded on the UI

Resolves: #[STT-1455](https://sofab.atlassian.net/browse/STT-1455)

[STT-1455]: https://sofab.atlassian.net/browse/STT-1455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/newsroom/commands/initialize_data.py
+++ b/newsroom/commands/initialize_data.py
@@ -94,7 +94,7 @@ class AppInitializeWithDataCommand:
 
             except Exception as ex:
                 logger.exception(ex)
-                logger.error("Unexpected error processing entity %s", name)
+                logger.error("Exception loading entity %s", name)
 
         logger.info("Data import finished")
         return 0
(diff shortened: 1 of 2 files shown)


MarkLark86 (approved):
We should ideally fix this in superdesk-core to support both eve/sync, eve/async and pydantic/async resources services.
But this will do for now I suppose

PR #1403, opened 2025-11-20
fix runtime error when locking

petrjasek:
there is

```
RuntimeError: generator didn't yield
```

when lock is already taken.

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/newsroom/push/tasks.py
+++ b/newsroom/push/tasks.py
@@ -1,6 +1,5 @@
 import logging
 
-from contextlib import contextmanager
 from superdesk.lock import lock, unlock
 
 from newsroom.celery_app import celery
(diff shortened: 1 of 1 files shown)


newsroom/push/tasks.py
```diff
@@ -13,33 +12,34 @@
...
         if item:
             await notifier.notify_new_item(item.to_dict(), check_topics=check_topics)
+    finally:
+        unlock(lock)
 
 
 @celery.task
 async def notify_new_agenda_item(_id, check_topics=True, is_new=False):
-    with locked(_id, "agenda"):
+    lock = get_lock(_id, "wire")
```
eos87: I think this should be `agenda`. Also variable name here

```suggestion
    agenda_lock = get_lock(_id, "agenda")
```

newsroom/push/tasks.py
```diff
@@ -13,33 +12,34 @@
...
-        unlock(lock_name)
-        logger.debug("Done with %s", lock_name)
+        return None
+    return lock_name
 
 
 @celery.task
 async def notify_new_wire_item(_id, check_topics=True):
-    with locked(_id, "wire"):
+    lock = get_lock(_id, "wire")
```
eos87: nit: I would suggest using a different variable name as we also import `lock` from superdesk
```suggestion
    wire_lock = get_lock(_id, "wire")
```

eos87 (changes requested):

eos87 (approved):

PR #1446, opened 2026-02-13
hide unposted planning items

petrjasek:
like we do with events

CPCN-1183

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/tests/core/test_push_events.py
+++ b/tests/core/test_push_events.py
@@ -1085,6 +1085,62 @@ async def test_filter_killed_events(client, app):
     assert "killed" == parsed["state"]
 
 
+async def test_push_killed_event_with_pubstatus(client, app):
+    """Test that events with state=killed and pubstatus=cancelled remain hidden"""
+    event = deepcopy(test_event)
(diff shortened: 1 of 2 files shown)


MarkLark86 (approved):

tests/core/test_push_events.py
```diff
@@ -1085,6 +1085,62 @@ async def test_filter_killed_events(client, app):
...
+    # Superdesk sends both state=killed and pubstatus=cancelled when an item is killed
+    event["state"] = "killed"
+    event["pubstatus"] = "cancelled"
+    await post_json(client, "/push", event)
+
+    # Item should be hidden from search (killed state takes precedence)
+    events = await get_json(client, "/agenda/search")
+    assert 0 == len(events["_items"])
+
+    # But should still be accessible directly
```
MarkLark86, line 1105: Is this right? If something is killed, should it only be accessible to admins?

petrjasek: I think the idea was that if you have a link you should be able to see it, eg. via email notification

PR #1466, opened 2026-03-24
SDAN-759 Scheduled updates not being displayed

marwoodandrew:
### Purpose
When a planning item is published to Newsroom with a scheduled update this is not reflected in Newsroom

### What has changed
The time of the scheduled update will be displayed.

### Steps to test

- Create a planning item with a text coverage, and also schedule an update.
- Fulfill the text coverage and publish it.
- Find the agenda item in newsroom agenda and select it, note that the string "Update expected @" an the time of the update is displayed in the coverage preview.

Should be something like
image

### Checklist

Resolves: SDAN-759

### DIFF
--- a/assets/agenda/utils.ts
+++ b/assets/agenda/utils.ts
@@ -494,7 +494,7 @@ export const getNextPendingScheduledUpdate = (coverage?: IFullCoverage) => {
     const scheduledUpdates = coverage.scheduled_updates || [];
     const deliveries = coverage.deliveries || [];
 
-    if (coverage.scheduled == null) {
+    if (coverage.scheduled === null) {
         // Not privileged to see coverage details
         return null;
     } else if (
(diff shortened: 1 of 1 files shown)


assets/agenda/utils.ts
```diff
@@ -494,7 +494,7 @@ export const getNextPendingScheduledUpdate = (coverage?: IFullCoverage) => {
     const scheduledUpdates = coverage.scheduled_updates || [];
     const deliveries = coverage.deliveries || [];
 
-    if (coverage.scheduled == null) {
+    if (coverage.scheduled === null) {
```
MarkLark86: What's the difference here between loose equality and strict equality. As far as I'm aware, the `scheduled` field would either have a datetime value or null/empty.

I believe it's empty when the user's company has the `Restrict Coverage Info` turned on, otherwise it should always have a value, no?

marwoodandrew: It should only return if the scheduled is defined and has a null value.

In my test the coverage does not have a scheduled field it is undefined!

I'll put a copy of the agenda record in the ticket!

MarkLark86 (approved):

PR #1479, opened 2026-04-21
[STT-1651] improve: Use cache in background thread

MarkLark86:
### Purpose
When modifying resources, sometimes the cache for that resource is cleared. If an exception is raised during clearing of this cache, the parent request will also fail.

### What has changed
* Use `CACHE_REDIS_` to set default Redis client timeout settings
* Added new `in_thread` methods to the Flask Caching extension we're using (for `app.cache`)
* Used the new cache in thread commands across the codebase
* Add tests

**Note:** There is a `TODO` comment in this PR that needs further discussion, and fix in a separate task.

Resolves: STT-1651

### DIFF
--- a/newsroom/factory/cache.py
+++ b/newsroom/factory/cache.py
@@ -28,7 +28,7 @@ class NewshubCache(Cache):
 
         super().init_app(app, config)  # type: ignore[arg-type]
 
-    def set_in_thread(self, key: str, value: Any, timeout: int | None = None):
+    def set_in_background(self, key: str, value: Any, timeout: int | None = None):
         """
         Add/update a key/value pair in the cache, in a background thread.
 
@@ -41,7 +41,7 @@ class NewshubCache(Cache):
 
         get_current_wsgi_app().add_background_task(self.set, key, value, timeout=timeout)
 
-    def set_many_in_thread(self, mapping: dict, timeout: int | None = None) -> None:
+    def set_many_in_background(self, mapping: dict, timeout: int | None = None) -> None:
         """
         Set multiple keys and values from a mapping, in a background thread.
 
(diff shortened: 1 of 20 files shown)


newsroom/factory/cache.py
```diff
@@ -0,0 +1,102 @@
...
+            # Timeout for initial connection
+            config["CACHE_OPTIONS"].setdefault(
+                "socket_connect_timeout", app.config.get("CACHE_REDIS_CONNECT_TIMEOUT", 2)
+            )
+            # Optionally retry once after a timeout
+            config["CACHE_OPTIONS"]["retry_on_timeout"] = True
+
+        super().init_app(app, config)  # type: ignore[arg-type]
+
+    def set_in_thread(self, key: str, value: Any, timeout: int | None = None):
```
petrjasek: would maybe call those methods like `set_in_background` , thread is an implementation detail

newsroom/factory/cache.py
```diff
@@ -0,0 +1,102 @@
...
+        :return: The value for the key, or None if the key is not found.
+        """
+
+        try:
+            loop = asyncio.get_running_loop()
+            return await asyncio.wait_for(loop.run_in_executor(None, self.get, key), timeout=async_timeout)
+        except asyncio.TimeoutError:
+            return None
+
+    async def get_dict_in_thread(self, keys: list[str], async_timeout: int = 2) -> dict[str, Any] | None:
```
petrjasek: seems like we only use those `get` methods for testing, should we have them public? or at all?

MarkLark86: I updated the PR to use more get/set background methods throughout the codebase (which should also help not hold up event loop as much).

The only method left that's only used in tests is that `get_dict_in_background` as I'm using that to test `set_many_in_background`. I left the `get_dict` variant in there as it can be useful to others outside of tests.

petrjasek (approved):

PR #1518, opened 2026-06-29
fix data updates to allow app based updates

marwoodandrew:
### Purpose
The parameter validation would not allow data_upates from the application repos's to be run individualy.

### What has changed
The validation that ensures any update script specified to the update command exists has been moved to the run function as by then the script folder for the application based updates has been added.

### Steps to test
Run an update script from the application

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/newsroom/data_updates.py
+++ b/newsroom/data_updates.py
@@ -2,6 +2,7 @@ import os
 import re
 import time
 import getpass
+import click
 import superdesk
 
 from string import Template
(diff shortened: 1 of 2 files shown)


newsroom/data_updates.py
```diff
@@ -101,6 +101,12 @@ def get_applied_updates(self):
 
     async def run(self, data_update_id=None, fake=False, dry=False):
         # TODO-ASYNC: revisit when `data_updates` module is migrated to async
+        if data_update_id and data_update_id not in get_data_updates_files(strip_file_extension=True):
+            print(
```
MarkLark86: In this scenario the command would have a return code of `0`. It might be good to use `raise click.UsageError` instead, so a non-zero return value is used

marwoodandrew: Updated to raise exception, example output below with exit code 2!

`Usage: manage.py data_upgrade [OPTIONS]
Try 'manage.py data_upgrade --help' for help.

Error: Invalid value for --id/-i: 'BOGUS_UPDATE' (choose from  ['00000_20250429-101000_users', '00001_20260511-124818_topics', '00002_20260512-120644_products', '00003_20260513-115834_users', '00004_20260514-134804_navigations', '00005_20260514-150759_products'])

Process finished with exit code 2
`

MarkLark86 (approved):

PR #1519, opened 2026-06-29
Add the ABC logo

marwoodandrew:
### Purpose
Adds a logo png for the Australian Broadcasting Corporation 

### What has changed
New Logo image

### Checklist

Resolves: #[issue-number]


MarkLark86 (approved):
This should be fine for now, as we already have other logos here too. We should probably look at moving customer specifics ones later

PR #1534, opened 2026-07-28
introduce email monitoring

petrjasek:
add celery beat task which sends an email and stores the status to redis.

NHUB-633

### Checklist

Resolves: #[issue-number]

### DIFF
--- a/newsroom/email_delivery_monitor.py
+++ b/newsroom/email_delivery_monitor.py
@@ -16,7 +16,6 @@ REDIS_KEY = "newsroom:email_delivery_monitor"
 
 class EmailDeliveryMonitor:
     def __init__(self):
-        self.started_at = utcnow()
         self.log_msg = "Email delivery monitor"
 
     def get_recipients(self) -> list[str]:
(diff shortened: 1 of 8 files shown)


newsroom/email_delivery_monitor.py
```diff
@@ -0,0 +1,87 @@
...
+            logger.info("%s status=started recipients=%s", self.log_msg, recipients)
+            self.record_status("started", timestamp, recipients)
+
+            subject = "Newsroom email delivery monitor"
+            text_body = (
+                "Email delivery monitor status=ok\n"
+                f"timestamp={timestamp.isoformat()}\n"
+                f"recipients={', '.join(recipients)}"
+            )
+            await send_email(to=recipients, subject=subject, text_body=text_body)
```
MarkLark86, line 73: We're not checking the return value from [send_mail](https://github.com/superdesk/newsroom-core/blob/develop/newsroom/email.py#L115). We should probably look at that too, in case email proxy server rejects the submission

petrjasek: looks like there is no return value from that send
https://github.com/pallets-eco/flask-mail/blob/main/src/flask_mail/__init__.py#L156-L158

MarkLark86: That explains why we never checked.
It's there on newer async email, so this will need updating afterwards

MarkLark86 (approved):

PR #1540, opened 2026-08-05
implement password reset via firebase admin sdk

petrjasek:
implement password reset for firebase users via
admin sdk and handle emails and tokens via newsroom.

CPCN-1218

### Checklist

Resolves: #[issue-number]


eos87 (approved):
Looks good.

